Issue #14, opened 2018-04-21
unhashable type: 'dict_node'


trondhindenes:
I'm getting this:
```
2018-04-22 00:46:35,869 - cfnlint - DEBUG - Get resources from template...
Traceback (most recent call last):
  File "/usr/local/bin/cfn-lint", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.6/dist-packages/cfnlint/__main__.py", line 157, in main
    matches.extend(runner.run())
  File "/usr/local/lib/python3.6/dist-packages/cfnlint/__init__.py", line 560, in run
    self.filename, self.cfn, self.ignore_checks))
  File "/usr/local/lib/python3.6/dist-packages/cfnlint/__init__.py", line 107, in run
    matches.extend(rule.matchall(filename, cfn))
  File "/usr/local/lib/python3.6/dist-packages/cfnlint/__init__.py", line 55, in matchall
    results = self.match(cfn)  # pylint: disable=E1102
  File "/usr/local/lib/python3.6/dist-packages/cfnlint/rules/resources/properties/Properties.py", line 195, in match
    resourcetype, resourcename, path, True))
  File "/usr/local/lib/python3.6/dist-packages/cfnlint/rules/resources/properties/Properties.py", line 131, in propertycheck
    parenttype, resourcename, arrproppath, False))
  File "/usr/local/lib/python3.6/dist-packages/cfnlint/rules/resources/properties/Properties.py", line 172, in propertycheck
    resourcename, proppath, False))
  File "/usr/local/lib/python3.6/dist-packages/cfnlint/rules/resources/properties/Properties.py", line 111, in propertycheck
    if prop not in resourcespec:
TypeError: unhashable type: 'dict_node'
```

The template deploys just fine.

kddejong:
Thanks @trondhindenes .  Looking into it.  I may need a few more details if I hit a wall.

kddejong:
My current theory is we are running into a list/array when we weren't expecting one.  Could be a list of lists.  The result is that when we do the for loop we are doing it on an array instead of a dict.  If I switch it to items to be specific we will still get an error.  Not sure how we are getting an array there thinking through some additional scenarios.

trondhindenes:
Let me know if you want me to post the template!

kddejong:
@trondhindenes that would be fantastic.

trondhindenes:
No problem, here's the unmodified one:
```
Parameters:
  VpcId:
[...]

PR #16, opened 2018-04-23
Fix prop and req checks when they get a list instead of dict

kddejong:
*Issue #, if available:* 
#14 

*Description of changes:*
- Fix properties and required checks when they get a list instead of an object or dict


cmmeyer:
Misspelling of "malformed" in error message:
`template is maflormed:`

Otherwise behavior is expected.

PR #24, opened 2018-04-30
Add override spec mechanism

fatbasstard:
Add a simple mechanism to override/alter/customize the AWS Resource Specification.

Example:

```
{
  "PropertyTypes": {
    "AWS::S3::Bucket.WebsiteConfiguration": {
      "Properties": {
        "ErrorDocument": {
          "Required": true
        }
      }
    }
  },
  "ResourceTypes": {
    "AWS::S3::Bucket": {
      "Properties": {
        "BucketName": {
          "Required": true
        }
      }
    }
  }
}

```

This files makes the Bucketname required, just like the Errordocument
(if a WebsiteConfiguration is specified)

This gives people the ability to "customize" the linting rules. Say a company wants to enforce certain values to be specified (like `BucketName`, `Encryption` or `Tags`) this allows the overriding of the AWS specifications.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@ may need to manually remove the cfn-lint binary.
 | --ignore-checks | ignore_checks | [IGNORE_CHECKS [IGNORE_CHECKS ...]] | Only check rules whose id do not match these values |
 | --log-level | log_level | {info, debug} | Log Level |
 | --update-specs | | | Update the CloudFormation Specs.  You may need sudo to run this.  You will need internet access when running this command |
+| --override-spec | | filename | Spec-style file containing custom definitions. Can be used to override CloudFormation definitions |
 | --version | | | Version of cfn-lint |
 
 ### Command Line
--- a/src/cfnlint/__main__.py
+++ b/src/cfnlint/__main__.py
@@ -156,7 +156,7 @@ def main():
             elif e.errno == 13:
                 LOGGER.error("Permission denied when accessing override spec file: %s", filename)
                 sys.exit(1)
-        except (json.decoder.JSONDecodeError, ScannerError) as err:
+        except (ValueError) as err:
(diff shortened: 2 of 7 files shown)


fatbasstard:
Need to write some unit tests for this, but this is the general idea. Simple yet powerful

fvanboven:
Would be nice to see the error output...

kddejong:
Yea not sure how I can do that yet... Let me think on that.

Using config file /cfnlint/pylintrc
************* Module cfnlint.helpers
C: 55, 0: Missing function docstring (missing-docstring)
W: 69, 8: Redefining name 'region' from outer scope (line 52) (redefined-outer-name)
W: 92, 8: Redefining name 'region' from outer scope (line 52) (redefined-outer-name)
************* Module cfnlint.__main__
W:160, 0: Bad indentation. Found 16 spaces, expected 12 (bad-indentation)
W:161, 0: Bad indentation. Found 16 spaces, expected 12 (bad-indentation)
E:159,16: Module 'json.decoder' has no 'JSONDecodeError' member (no-member)

kddejong:
Looks great to me.  Would love to see some tests.  Let me know if you need any help with writing tests.

fatbasstard:
Having some dependency/flow struggles with the loading of all the Rules, The CloudSpec loading. Let me think about this...

(specs are loaded immediately into memory, the specs are used in the __init__ of some rules)

kddejong:
@fatbasstard Yea not sure that makes sense anymore either.  Originally I had the individual rules loading the Specs so I put it into the __init__.  Then I moved the loading of the Specs into helpers so we didn't have those spec files being loaded across many different rules.  I'm not sure there is a good reason to keep those things in the init honestly.   Open to thoughts on that.

fatbasstard:
Added a test, which also displays the Spec loading "issue". I'll think about refactoring, but for now got something

Issue #19, opened 2018-04-26
Best Practices


kddejong:
A question was asked if we should be including template issues and best practices with the same repository of rules.

Example would be for an image ID property.
1. Refs a Parameter of Type Number would not work at all and should show a failure.
1. Refs a Parameter of type String shows a failure today because its not AWS::EC2::Image::Id.  In this case both would work but AWS::EC2::Image::Id is more right. 

Another example would be if you are specifying a password for a RDS database we make sure the REF to a parameter has the property NoEcho True set.  Again a best practice but not a requirement for the template to work.

This is a great question and deserves some conversation.  I see two options.
1. Separate out best practice rules into another repo that people can download and use as they want. 
1.  Keep Issues in this repo but allow people to turn off or enable best practices as needed.
  a. Tag Rules as Best Practice vs. Issue (wording to be worked on) and allow people to enable best practices as they need (disabled by default).  
  b. A variance on the last one would be making best practices Warnings and allowing people to disable Warnings via parameter.  Exit code would be determined on Errors and not on Warnings.

Right now I'm leaning towards a model in which we keep the rules in this repo but allow people to disable best practices.

cmmeyer:
I like the "Error" vs. "Warning" model similar to other linters (Flake8 for example).  This would allow separation of issues that will cause CloudFormation to fail from stylistic or other best practices.

kddejong:
So I think we are in agreement of this.  
- Errors will focus on things that will fail when running a deployment.
- Non-Errors will focus on best practice items.  

Pull request 28 includes a capability to disable all Warnings.

PatMyron:
Re-opening this issue to track progress on some broadly applicable best practices:

- [ ] [Sensitive properties](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/rules/resources/properties/Password.py)
- [ ] Encryption
- [ ] Backups/Versioning
- [ ] Autoscaling/On-demand
- [ ] Logging

kddejong:
Closing this.  We aren't tracking or working on this issue and can add individual rules on request.

PR #28, opened 2018-05-02
Fix/breakout warnings from errors

kddejong:
*Issue #, if available:* #19 

*Description of changes:*
- Remove best practice from errors so they can be classified differently.
- Better Error Code results (1 fatal, 2 lint error, 4 lint warning, 32 usage error)
- Allow for more robust disabling.   Example. ```--ignore-checks W``` will disable all Warnings.


cmmeyer:
You should expose the new `--ignore-checks` flag via the help text.
One error disappeared between checks -- was it intentionally removed? Looks like it could have turned into a Warning, but I'm not seeing it:
E2505 AllowedPattern for Parameter should be specified at Parameters/VpcCIDR. Example "^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])(\/([0-9]|[1-2][0-9]|3[0-2]))$"
CodePipelineOutputTest1.yaml:7:3

cmmeyer:
Also, looks like we have a regression on JSON w/ tabs:
2018-05-07 15:49:45,040 - cfnlint - ERROR - Template conditional-private-subnets.json is malformed: while scanning for the next token
found character '\t' that cannot start any token
  in "<unicode string>", line 315, column 47:
     ... on" : "CreatePrivateSubnet",

Issue #29, opened 2018-05-02
CodePipeline rules


seporaitis:
Hi,

As per contribution suggestions opening up issue first.

Today trying to update CloudFormation stack with CodePipeline I ran into two rollbacks:

1. Pipeline should start with a stage that only contains source actions.
2. Pipeline action declares 0 input artifacts.

Validate template didn't caught those and after quick search I found this project, which made it really easy to write additional rules to validate those two cases:

The first rule just goes through first stage in the pipeline and:

1. If there are no `Source` actions - pass.
2. If there are only `Source` actions - pass.
3. If there are `Source` and other actions - fail.

The second rule goes through stages, starting from the second one, and:

1. If the action category is `Build` and there are no `InputArtifacts` key or it's empty - fail.
2. Otherwise - pass.

I would gladly contribute these rules back to the project - wondering if you think they'd provide value? Are there any nuances that above rules miss? 

As a side question, and I am not sure if you can answer it, but why `cloudformation:ValidateTemplate` API call would not run this more comprehensive linter on the templates?

fatbasstard:
> As a side question, and I am not sure if you can answer it, but why cloudformation:ValidateTemplate API call would not run this more comprehensive linter on the templates?

This linter is brandnew (version 0.0.8 should give it away), I hope/guess/expect that it will be implemented in the `ValidateTemplate` in the future

cmmeyer:
Thanks for this! Let me dive in with the CodePipeline team to verify the logic, but this seems like a reasonable addition.  I'm noticing the PoC is failing tests in the build pipeline -- you may want to run tox locally against your branch to correct any test failures and submit a new commit.

> As a side question, and I am not sure if you can answer it, but why cloudformation:ValidateTemplate API call would not run this more comprehensive linter on the templates?

Improved validation is a high priority for the CloudFormation service team, but there are challenges around integrating this type of functionality at cloud scale within the service for 400k+ active users.  Even once API based validation exists, there are still places were a local linter just makes sense (IDE integration for example).  This linter was written by the awesome @kddejong out of AWS Professional Services and struck me as an excellent stop gap, so when he mentioned he was requesting open source approval, I jumped on it.  

Think of it as a middle ground -- an open source linter you can imbed into your own applications/pipelines, with the opportunity for direct interaction with the service team to validate implementation.

PR #30, opened 2018-05-02
PoC of simple CodePipeline rules

seporaitis:
*Issue #:* #29 

*Description of changes:*

A proof of concept of two rules described in #29. If deemed worthy of adding to the project - I will tidy up the code and add test cases.


kddejong:
If this helps.  You should be able to run tox locally to do some testing.

Looks like these come from Python 2.7
```
Using config file /cfnlint/pylintrc
************* Module cfnlint.rules.resources.codepipeline.ActionDeclaresZeroInputArtifacts
E: 29, 0: invalid syntax (<string>, line 29) (syntax-error)
************* Module cfnlint.rules.resources.codepipeline.PipelineShouldStartWithSourceStage
E: 32, 0: invalid syntax (<string>, line 32) (syntax-error)
```

seporaitis:
Thanks @kddejong I have looked up CodePipeline [documentation(https://docs.aws.amazon.com/codepipeline/latest/userguide/reference-pipeline-structure.html) and refactored the `Stages` validation. Instead of just once check I added, I believe, all that apply to stages from documentation.

This is still work in progress - the build will fail. I still plan to refactor `src/cfnlint/rules/resources/codepipeline/ActionDeclaresZeroInputArtifacts.py` into `Actions.py` and do validation of actions as per documentation. This will probably come in over the weekend.

Meanwhile, your feedback is welcome. Mainly, I am not sure if I broke down the functions in a right way. My inspiration were `Vpc` and `Elb` rule classes. Also, I attempted to use `cfn.check_resource_property`, but I believe it does not support properties which are lists.

From: `src/cfnlint/__init__.py:612:get_values`:

```
        if isinstance(value, (dict, list)):
            if len(value) == 1:
                for obj_key, obj_value in value.items():
```

the `value.items()` fails if `value` is a `list. I wasn't sure if this was intended behavior or not so decided not to touch it yet.

kddejong:
@seporaitis I'll try to check this out tonight.  We ran into that with something else so I want to do some double checking and validating.

kddejong:
that looks an issue to me.  Especially since it breaks into a .items().  If it was actually a list we should have done enumerate instead.    Switching it and running tox shows success.  Trying to think if there is something I'm missing.

seporaitis:
@kddejong I have put up a working version for review that passes the build and has some more validation. Looking forward to hear your feedback!

kddejong:
This is looking great.  Going to dive into this tonight and see if I can see anything else.

src/cfnlint/rules/resources/codepipeline/CodepipelineStageActions.py
```diff
@@ -0,0 +1,228 @@
...
+                    b=max_
+                )
+                matches.append(RuleMatch(
+                    path + [artifact_type],
+                    message
+                ))
+        else:
+            if artifact_count != constraints[constraint_key]:
+                message = (
+                    'Action "{action}" declares {number} {artifact_type} which is less than '
```
kddejong: Error is "less than" but the comparison is not equal.

src/cfnlint/rules/resources/codepipeline/CodepipelineStageActions.py
```diff
@@ -0,0 +1,228 @@
...
+                message
+            ))
+
+        return matches
+
+    def check_version(self, action, path):
+        """Check that action type version is valid."""
+        matches = []
+
+        if action.get('ActionTypeId', {}).get('Version') != 1:
```
kddejong: We may want to check that ActionTypeId is a dict and not a list (or some other object) before doing the get on the result.

Going to my other comment if we had tests to validate ActionTypeId is an object we would just ignore a failure here.  Then here we are just making sure its an object before we do a .get (if its a list it will crash).   I always debate generic try/catches here for common errors like this.  At least when your checks aren't validating that specifically you can ignore them.  

But I'm thinking a generic CodePipeline syntax checker would be nice.  Just covering the basic structure and supported types for the elements.

kddejong: Never mind that is in the CloudFormation spec from what I can tell.  So we should be putting in a more generic try/catch on AttributeError for 'AttributeError: 'list_node' object has no attribute 'get'' that you could skip but that would result in cfn-lint not crashing.

Or we would want to test the items are dicts before doing a get.

src/cfnlint/rules/resources/codepipeline/CodepipelineStageActions.py
```diff
@@ -0,0 +1,228 @@
...
+        for resource in resources:
+            path = resource['Path']
+            properties = resource['Value']
+
+            for sidx, stage in enumerate(properties['Stages']):
+                stage_path = path + ['Stages', sidx]
+                action_names = set()
+                for aidx, action in enumerate(stage['Actions']):
+                    action_path = stage_path + ['Actions', aidx]
+
```
kddejong: could do 
```
try:
  matches.extend(self.check_names_unique(action, action_path, action_names))
  matches.extend(self.check_version(action, action_path))
  matches.extend(self.check_owner(action, action_path))
  matches.extend(self.check_artifact_counts(action, 'InputArtifacts', action_path))
  matches.extend(self.check_artifact_counts(action, 'OutputArtifacts', action_path))
except AttributeError as err:
  self.logger.debug('Got Attribute Error. Should have been caught by generic linting. '
                                 'Ignoring the error here: %s', err)
```

This would result in template.
```
Stages:
        - Name: Source
          Actions:
            - Name: Github
              ActionTypeId:
              - Category: Source
                Owner: ThirdParty
                Provider: GitHub
                Version: 1
              OutputArtifacts:
                - Name: MyApp
              Configuration:
                Owner: awslabs
                Repo: cfn-python-lint
                PollForSourceChanges: true
                Branch: master
                OAuthToken: 'secret-token'
```
Getting Error (instead of crashing)
```
E3002 Expecting an object at Resources/TestPipeline/Properties/Stages/0/Actions/0/ActionTypeId
test/templates/bad/resources_codepipeline_stages_one_stage.yaml:17:15
```

seporaitis:
Thanks for the feedback! I will have time to look deeper into it tomorrow night. Also, FYI, I will be away from computer for two weeks starting next week, and thus less responsive, but definitely keen to work on it and see it merged when I get back.

seporaitis:
I have updated with the `AttributeError` try/except block and couple additional checks that I noticed could fail the linter.

src/cfnlint/rules/resources/codepipeline/CodepipelineStageActions.py
```diff
@@ -0,0 +1,254 @@
...
+            ))
+
+        return matches
+
+    def check_owner(self, action, path):
+        """Check that action type owner is valid."""
+        matches = []
+
+        owner = action.get('ActionTypeId').get('Owner')
+        if owner not in self.VALID_OWNER_STRINGS:
```
kddejong: I'm debating if this should be.
`if owner not in self.VALID_OWNER_STRINGS and owner is not None:` since the Owner check will be covered by the required properties check (E3003) if someone misses Owner with this version they will get double errors.

```
E3003 Property Owner missing from resource DeployPipeline
test/templates/quickstart/codepipeline.yaml:441:11

E2541 For all currently supported action types, the only valid owner strings are AWS, Custom, ThirdParty
test/templates/quickstart/codepipeline.yaml:444:15
```

where as with the fix it would just be 
```
E3003 Property Owner missing from resource DeployPipeline
test/templates/quickstart/codepipeline.yaml:441:11
```

src/cfnlint/rules/resources/codepipeline/CodepipelineStageActions.py
```diff
@@ -0,0 +1,254 @@
...
+                    message
+                ))
+
+        return matches
+
+    def check_action_keys(self, action, path):
+        """Check that action has required keys inside."""
+        matches = []
+
+        if 'Name' not in action:
```
kddejong: this is a double check with E3003.
```
E2541 Expecting Name key in Action.
test/templates/quickstart/codepipeline.yaml:441:11

E3003 Property Name missing from resource DeployPipeline
test/templates/quickstart/codepipeline.yaml:441:11
```

src/cfnlint/rules/resources/codepipeline/CodepipelineStageActions.py
```diff
@@ -0,0 +1,254 @@
...
+            message = 'Expecting Name key in Action.'
+            matches.append(RuleMatch(
+                path,
+                message
+            ))
+        if 'ActionTypeId' not in action:
```
kddejong: Ummm this should be the same and covered by E3003 but its not working which means I think you found another bug.

```
"ActionTypeId": {
          "Documentation": "http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-codepipeline-pipeline-stages-actions.html#cfn-codepipeline-pipeline-stages-actions-actiontypeid",
          "Required": true,
          "Type": "ActionTypeId",
          "UpdateType": "Mutable"
        },
```

kddejong: I apologize.  This was already fixed in the main branch.

seporaitis:
Apologies for missing out the updates on recent `master` - I forgot to rebase. Following your feedback I removed redundant checks, also found and remedied one more which would've caused these two to appear simultaneously when `Stages: []`:

``
E2540 The first stage of a pipeline must contain at least one source action.
temp.yml:13:7

E2540 The first stage of a pipeline must contain only source actions.
temp.yml:13:7
``

cmmeyer:
Generally working great! The only thing I encountered is that the "Version" attribute of  ActionTypeId isn't accepting strings.  

`Version: '1'` and `Version: "1"` both throw the error:
"For all currently supported action types, the only valid version is 1."

Here's the definition from the API (https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-codepipeline-pipeline-stages-actions-actiontypeid.html#cfn-codepipeline-pipeline-stages-actions-actiontypeid-version)

**version**
A string that describes the action version.
Type: String
Length Constraints: Minimum length of 1. Maximum length of 9.
Pattern: [0-9A-Za-z_-]+
Required: Yes

cmmeyer:
Also, where is "For all currently supported action types, the only valid version is 1."  documented?  I'm having trouble finding it in any of the specs.

If that's a best practice, it might be best to enforce as a warning instead of an error.

kddejong:
@cmmeyer. https://docs.aws.amazon.com/codepipeline/latest/userguide/reference-pipeline-structure.html

“For all currently supported action types, the only valid version string is "1".”

cmmeyer:
Disregard that second comment -- @kddejong showed me the requirement. Looks like the only piece that was missed was Version needs to be a *string* of "1".

kddejong:
Worth noting that with the discussion on #42 we may require this to be a string via another check as that is what the Spec states.

cmmeyer:
@kddejong does PR #83 make my concern here obsolete? If so I can merge this then review your PR.

cmmeyer:
Going to go ahead and merge as is. Really love this functionality and want to get it in the linter!

cmmeyer:
I just realized this didn't get called out in the release notes, but this is live in 0.2.0.

I wanted to let you know how much we appreciate the work that went into this and I for one am already using this functionality on a daily basis. Thanks!

seporaitis:
Thank you for merging in the changes!

PR #31, opened 2018-05-02
Fix required property checks

fatbasstard:
Found out the required property linter missed/broke on certain scenario's:

- Required Properties that are not `PrimitiveType` did not get linted
- Conditionals were not supported completed correctly

In this PR:
- Added a `bad` test template with these scenario's (`properties_required.yaml`)
- Rewritten part of the required check to solve the issues.
- Corrected existing `good` template that was actually not good
- Updated generic bad template since there's another issue now in there (`Code` missing is the Lambda function)


kddejong:
Looks good to me.

PR #32, opened 2018-05-04
Add whitelist/blacklist Resource feature to the override-spec

fatbasstard:
*Issue #, if available:*

Add a new feature to the `--override-spec` functionality to block the usage of resources using a whitelist/blacklist setup. In this way it's easy to prevent the unwanted use of specific resources.

Along the way:
- Restructured test path setup
- Added documentation about the functionality in the readme


kddejong:
I think I got this logic.  If no includes it will allow all resources.  If there is an includes it will only do the defined resources in `IncludeResourceTypes` (not doing all resources).  Then it will exclude anything defined in `ExcludeResourceTypes`.  So Exclude always wins which I think is the way to go.

fatbasstard:
Yes, completely correct. We use this feature is our current linter @ Coolblue. We encourage developers to write their own CloudFormation as much as possible and use a linter to guide them in this using the AWS Rules combined with our Coolblue rules. (We do enforce tagging by overwriting the `Required` of the `Tags` for instance) 

We use the whitelist/blacklist feature to prevent them from using services that are not yet "Enabled" for usage. Although I must say this implementation is actually better then our current one :P

fatbasstard:
Definitely understand and slightly agree. Although then I'll have to create another solution (we can solve this using a custom Rule for instance) to work around this. In the current version we're using (0.0.7) we actually use monkey patching to get the same result. 

I agree that it's a bit special, but this linter is already so powerful that I'd decided to offer it as a "feature". See what happens ;)

kddejong:
@cmmeyer Sorry I missed your response when I submitted my review.  I get where you are coming from but I don't mind this approach.  We give people the ability to extend the running rules with their own which would give them the same result.   I think the ease of doing it this way makes sense for users over writing their own rules.  

That being said I think we should sync up and decide where that line should be drawn.

fatbasstard:
I'm the same in this. I get the concerns but personally think it's within the boundaries... Anyway good to get a/the discussion started on the boundaries since there probably will be more opinions/ideas will come in in the future once this gets adopted more.

cmmeyer:
Working for me as well. Going ahead and merging this PR.

cmmeyer:
So I went ahead and merged this, but I'm really feeling icky about it now that I'm playing with it.  
For instance, I just added an Exclude file with all of EC2 ({ "ExcludeResourceTypes": [ "AWS::EC2::*" ] }).

Now I get errors like:
`E3001 Invalid Type AWS::EC2::SubnetRouteTableAssociation for resource PublicSubnet2RouteTableAssociation in us-east-1
simple-vpc.template.yml:197:9`

Which could be infuriating if I was a dev looking at the docs and saying "What are you talking about? It's *right there*!"

There should be some sort of indicator that this is by policy instead of based on the spec.  This was my initial concern with layering business logic over service functionality...

fatbasstard:
The same error is thrown if a resource type is used that is not available in a specific region (e.g. `AWS::Glue::Connection` in `sa-east-1`, which might be just as confusing.

Issue #35, opened 2018-05-04
Catch duplicate resources


fatbasstard:
Dictionaries in Python don't support duplicate keys. If a YAML/JSON contains duplicate keys, the last occurrence is "used". 

If somebody is copy/pasting a bit too enthusiastic in a big template, weird result will happen....  Change a resource, deploy it, nothing changes. Spand hours on debugging only to find out the Resource occorred 2 times in the template... Yes, experienced this.... Pretty annoying.

Example:

```
AWSTemplateFormatVersion: "2010-09-09"
Resources:
  mySnsTopic:
    Type: AWS::SNS::Topic
    Parameters:
      TopicName: "unused-topic-name"
  myS3Bucket:
    Type: AWS::S3::Bucket
  mySnsTopic:
    Type: AWS::SNS::Topic
    Parameters:
      TopicName: "used-topic-name"
```

There will be 1 SNS topic deployed, the one with `used-topic-name`. In this case pretty clear, but in bigger templates this is not desirable.

fatbasstard:
See also: https://github.com/aws/aws-cli/issues/2404

kddejong:
Totally ran into this when working on the JSON parser to work with tabbed JSON files.  Was wondering how to fix this but this is fantastic.  Thanks i'll review.

fatbasstard:
Just added the JSON version (somewhere I removed the code for JSON while making assumption)... Actually found a duplicate in the tests

PR #36, opened 2018-05-04
Check for duplicates

fatbasstard:
*Issue #35*

Now breaks on templates with duplicate resources. Since duplicates don't exists in a dict, it's "imposible" to check this is a normal rule.

Now hooks in the loading and reports the duplicate key name (and the line number) and throws an error at loading. Which sort of makes sense since the template itself is actually "invalid"

### DIFF
--- a/src/cfnlint/__main__.py
+++ b/src/cfnlint/__main__.py
@@ -19,6 +19,7 @@ import argparse
 import logging
 import json
 from yaml.parser import ParserError, ScannerError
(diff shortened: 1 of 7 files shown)


kddejong:
Will this work on JSON templates as well?

fatbasstard:
I thought I tested/build that... Let me double check and add tests

fatbasstard:
Messed JSON when I cleaned up my code for the PR... Json fix incoming

PR #37, opened 2018-05-05
Fix circular dependencies that are more than 1 level deep

kddejong:
*Issue #, if available:*

*Description of changes:*
- Fix detection of circular dependencies that are more than two resources deep.


cmmeyer:
Looks good. Merging.

PR #43, opened 2018-05-07
Returns conditions objects and lists when using high level conditions

### LINKED ISSUE #41: Complex Condition breaks Properties check (E3002)
Properties check breaks on complex Conditions that return multiple sub-properties. Example:

```
Description: A sample template
Parameters:
  EnableGeoBlocking:
    Type: String
Conditions:
  EnableGeoBlocking: !Equals [ !Ref EnableGeoBlocking, "true" ]
Resources:
  CloudFrontDistribution:
    Type: "AWS::CloudFront::Distribution"
    Properties:
      DistributionConfig:
        Enabled: true
        Restrictions:
          GeoRestriction:
            !If
              - EnableGeoBlocking
              -
                RestrictionType: whitelist
                Locations:
                  - BE
                  - LU
                  - NL
              - RestrictionType: none
```

This returns an error on the `!If`:

```
E3002 Expecting an object at Resources/CloudFrontDistribution/Properties/DistributionConfig/Restrictions/GeoRestriction/Fn::If/2
template.yaml:15:11
```

kddejong:
*Issue #, if available:*
Fixes #41 

*Description of changes:*
- Returns objects and lists that are part of condition results instead of dropping them or converting them to values (instead of keys)

### DIFF
--- a/src/cfnlint/__init__.py
+++ b/src/cfnlint/__init__.py
@@ -576,9 +576,11 @@ class Template(object):
             result = {}
             result['Path'] = path[:] + [index + 1]
             if not isinstance(item, (dict, list)):
+                # Just straight values and pass them through
                 result['Value'] = item
                 matches.append(result)
             elif len(item) == 1:
+                # Checking for conditions inside of conditions
                 for sub_key, sub_value in item.items():
                     if sub_key in cfnlint.helpers.CONDITION_FUNCTIONS:
                         results = self.get_condition_values(sub_value, result['Path'] + [sub_key])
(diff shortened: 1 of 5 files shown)


cmmeyer:
The base code seems fine, but does this need additional tests?

I notice this also adds a conditional check for AWS::CloudFront::Distributions -- I assume this was what surfaced the error?

kddejong:
@cmmeyer  at least in relation to the use of a condition as shown in the issue.  I added it as a test.  Are there more condition tests you can think of?
1. Conditions for values of a key (https://github.com/kddejong/cfn-python-lint/blob/47c18ff2f3705d50de37324e67fa21cf7e148a67/test/templates/good/conditions.yaml#L27)
1. Conditions to include a value (or object) in an array (https://github.com/kddejong/cfn-python-lint/blob/47c18ff2f3705d50de37324e67fa21cf7e148a67/test/templates/good/conditions.yaml#L32)
1. Conditions to include certain properties  (https://github.com/kddejong/cfn-python-lint/blob/47c18ff2f3705d50de37324e67fa21cf7e148a67/test/templates/good/conditions.yaml#L48) 

That being said I included some more direct tests so we can make sure we cover the scenarios and the results.

PR #55, opened 2018-05-11
Update Lambda Memory Size to 3008.  

### LINKED ISSUE #54: E2530: Support Lambda Memory values up to 3008
E2530 warns that Lambda memory capacity must not be greater than 1536. This should now be 3008 since Lambda doubled the maximum memory capacity. 

https://aws.amazon.com/about-aws/whats-new/2017/11/aws-lambda-doubles-maximum-memory-capacity-for-lambda-functions/

kddejong:
*Issue #, if available:*
Fixes #54 

*Description of changes:*
- Move lambda to lmbd as Lambda is a Python keyword
- Update Lambda Memory Size to 3008.


src/cfnlint/rules/resources/lmbd/FunctionMemorySize.py
```diff
@@ -31,17 +31,17 @@ def check_value(self, value, path):
...
                   'and it must be a multiple of 64. You cannot specify a size ' \
-                  'larger than 1536. The default value is 128 MB at {0}'
+                  'larger than 3008. Error at {0}'
 
         try:
             value = int(value)
 
-            if value < 128 or value > 1536:
-                matches.append(RuleMatch(path, message.format(value, ('/'.join(path)))))
+            if value < 128 or value > 3008:
```
fatbasstard: Isn't it an idea to make them constants? Or even configurable? Instead of changing it in 4 places

kddejong: fixed in most recent pull request

PR #56, opened 2018-05-11
Fix Circular Dependency Outputs bug

fatbasstard:
A reference in an Output using `!Sub` to a Resource with the same name as the Output is seen as a circular dependency:

```
E3004 Circular Dependencies for resource ElasticIP.  Circular dependency with [ElasticIP]
test/templates/good/generic.yaml:113:3
```
Added a check on a `Resource` type, just like in the `Ref` and `Fn::GetAtt` part.

Added this scenario to the `good/generic.yaml` template for the tests


kddejong:
Yikes.  Good catch

PR #58, opened 2018-05-14
Fix FindInMap not supporting FindInMap for the 0 element in the array

### LINKED ISSUE #57: Nesting FindInMap not supported
Don't ask why, but we have the following CloudFormation:

```
CidrBlock: !FindInMap [!FindInMap ["Environment", !Ref "Environment", "Subnets"], "eu-west-1a", "Management"]
```

Which causes an error:
```
E1011 Map Name should be a string for Resources/XXX/Properties/CidrBlock/Fn::FindInMap
../xxx.yaml:
```
The FindInMap rule does not support recursive FindInMaps.

kddejong:
*Issue #, if available:*
Fixes #57 

*Description of changes:*
Fixes FindInMap nesting for the 0 element of the array and adds tests.


cmmeyer:
I changed the name of the map in test/templates/good/functions_findinmap.yaml to "MyMap1" and received an exception:

2018-05-14 14:59:01,612 - cfnlint - ERROR - Tried to process rules on file test/templates/good/functions_findinmap.yaml but got an error: list indices must be integers or slices, not list

Same error if you remove the Mapping section completely.

PR #62, opened 2018-05-15
Add rule to warn for unused mappings

cmmeyer:
*Issue 59*

*Description of changes:*
Added a rule to see if defined Mappings are used anywhere in the template, new warning is (W7001) to match the equivalent Conditional rule on which this is based.

No new tests were required since the existing mappings.yaml has sufficient bad examples.


src/cfnlint/rules/mappings/Used.py
```diff
@@ -0,0 +1,63 @@
...
+
+            for maptree in maptrees:
+
+                if isinstance(maptree[-1], list):
+                    ref_mappings.append(maptree[-1][0])
+                else:
+                    ref_mappings.append(maptree[-1])
+
+#            # Get resource's Mappings (I don't think this applies?)
+#            for _, resource_values in cfn.get_resources().items():
```
fatbasstard: This can be removed indeed. The only thing that can reference a Maping is a FindInMap function

test/rules/mappings/test_used.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+class TestUsedMappings(BaseRuleTestCase):
+    """Test template mapping configurations"""
+    def setUp(self):
+        """Setup"""
+        super(TestUsedMappings, self).setUp()
+        self.collection.register(Used())
+
+    success_templates = [
```
fatbasstard: What happens if you just run this on the base's success_templates?

This override was added to the Conditions version because too much of the "good" quickstart templates became invalid (see comments in the PR of the Conditions Used rule)

src/cfnlint/rules/mappings/Used.py
cmmeyer: Cool! I'll pull this code out.  I thought I had :)

test/rules/mappings/test_used.py
cmmeyer: There are a couple of QuickStarts with unused Mappings, so I had to use the same trick.  @kddejong and I discussed moving this to an "exclude" model to make it easier to maintain going forward (especially if we add more sample templates).

I'll create a new issue recommending that approach going forward.

PR #63, opened 2018-05-15
Fix hardcoded AZ check

fatbasstard:
The (new) check on hardcoded AZ breaks if AZ's are specified at a load balancer. This is an array of values which returned a `IndexError: list index out of range` due to the slice (`child_path[-2]`)

Additional changes:
- Fixe some naming in the AZ parameters check
- Renamed the test of the AZ parameter test to `parameters_az.yaml`
- Created new test file for the AZ properties (`properties_az.yaml`)
 
### DIFF
--- a/src/cfnlint/__init__.py
+++ b/src/cfnlint/__init__.py
@@ -783,11 +783,10 @@ class Template(object):
                         check_split(
                             value=value, path=new_path[:] + child_path, **kwargs))
             elif isinstance(child_path[-1], int):
-                if child_path[-2] == 'Fn::If':
-                    if check_value:
-                        matches.extend(
-                            check_value(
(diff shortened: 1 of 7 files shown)


src/cfnlint/__init__.py
```diff
@@ -783,11 +783,10 @@ def check_value(self, obj, key, path,
                         check_split(
                             value=value, path=new_path[:] + child_path, **kwargs))
             elif isinstance(child_path[-1], int):
-                if child_path[-2] == 'Fn::If':
```
kddejong: yea I'm not sure I understand why I did this...

kddejong:
Nevermind on the test... its in other templates my bad.

PR #65, opened 2018-05-15
Support multiple functions without RestApiId

fatbasstard:
If there are multiple events without `RestApiId` specified, the Transform generates an error:

```
E0001 While Performing a Transform got error: A resource with ID "ServerlessRestApi" already exists within this template
test/templates/good/transform_serverless_function.yaml:4:1
``` 
From the documentation (https://github.com/awslabs/serverless-application-model/blob/develop/versions/2016-10-31.md#api):

```
If not defined, a default AWS::Serverless::Api resource is created using a generated Swagger document contains a union of all paths and methods defined by Api events defined in this template that do not specify a RestApiId.
```

Added a check to create only 1 RestApi resource

### DIFF
--- a/src/cfnlint/transforms/Serverless/Functions.py
+++ b/src/cfnlint/transforms/Serverless/Functions.py
@@ -114,11 +114,12 @@ class Functions(CloudFormationTransform):
                     )
 
                 events = resource_properties.get('Events', {})
+                generated_api = False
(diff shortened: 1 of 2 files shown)


kddejong:
Merging this for now.  Later we can work on a merge when it becomes a need.

PR #67, opened 2018-05-15
Fix GetAtts in SubStrings that go to custom and CloudFormation stacks

### LINKED ISSUE #66: Best way to handle parameters from nested stacks?
This project is awesome and super useful!

I can imagine the complexity of linting nested stacks is super high and wouldn't expect that, but how are folks handling situations where Outputs from nested stacks are being used?

For example, given this code:
```
            /etc/aws-signing-proxy.yml:
              content: !Sub |
                listen-address: 127.0.0.1
                port: 9200
                target: https://${ElasticsearchStack.Outputs.DomainEndpoint}
                region: ${AWS::Region}
```

always throws:
```
E1019 String parameter ElasticsearchStack.Outputs.DomainEndpoint not found in string for Resources/ServerLaunchConfig/Metadata/AWS::CloudFormation::Init/es_proxy_setup/files//etc/aws-signing-proxy.yml/content/Fn::Sub
```

I don't want to disable all E1019 checks because they're useful.  Would it maybe be right, for now, to exclude all parameters that have `/.Outputs./` in their name?

kddejong:
*Issue #, if available:*
Fixes #66 

*Description of changes:*
- Fixes sub string validating getatt requests to custom and cloudformation stack resources.  Technically getatts of those resource types come back as resourcename.* and this wasn't handled when doing the 'in' based search.  Had to some more string tests to handle these resource types.  

### DIFF
--- a/src/cfnlint/__init__.py
+++ b/src/cfnlint/__init__.py
@@ -482,10 +482,15 @@ class Template(object):
         resourcetypes = cfnlint.helpers.RESOURCE_SPECS['us-east-1'].get('ResourceTypes')
         results = {}
         resources = self.template.get('Resources', {})
+
+        astrik_types = (
+            'Custom::', 'AWS::CloudFormation::Stack',
+            'AWS::Serverless::', 'AWS::CloudFormation::CustomResource'
+        )
         for name, value in resources.items():
             if 'Type' in value:
                 valtype = value['Type']
-                if valtype.startswith(('Custom::', 'AWS::CloudFormation::Stack', 'AWS::Serverless::')):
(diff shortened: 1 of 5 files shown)


irvingpop:
hey @kddejong thanks for the fast turnaround on this!   One weird side-effect I'm seeing after pulling your branch is that a number of load-balancer related GetAtts now throw errors that didn't previously:

```

BetaKevin:
I’ll test tonight and try to get this fixed. Thanks for catching this... again.

irvingpop:
Working perfectly now, thank you!

PR #68, opened 2018-05-15
Fix getting config like ignore_checks from the template

irvingpop:
Signed-off-by: Irving Popovetsky <irving@chef.io>

*Description of changes:*
I noticed that setting configuration in my templates (such as `ignore_checks`) wasn't working.  digging around, I noticed this if statement was never evaluating to `True`: https://github.com/awslabs/cfn-python-lint/blob/master/src/cfnlint/__main__.py#L77

because:
```python
>>> template is dict
False
>>> template.__class__
<class 'cfnlint.parser.dict_node'>
```

this is a more reliable check:
```python
>>> isinstance(template, dict)
True
```


kddejong:
Makes sense to me. We test using isinstance a lot not sure how we missed this one. Thanks.

PR #79, opened 2018-05-19
Don't add role if there's one existing already

janza:
This PR fixes an issue with a function (`FunctionName`) that has explicitly defined role in the format of: `FunctionNameRole`. Trying to lint that will fail with message: 

> A resource with ID "FunctionNameRole" already exists within this template

The issue is fixed by not creating a new Role resource if serverless function has an existing reference to a role.


kddejong:
Seems good and simple to me. Let me dig a little deeper when I have a little more time.

kddejong:
@janza as a side question do you know if this applies to other items or just to roles?

janza:
It could be that other items have the same issue but I'm not sure.

To properly fix the issue it might make sense to try including the samtranslator as a dependency and transform the template using latest code from https://github.com/awslabs/serverless-application-model

Issue #47, opened 2018-05-07
Move logic away friom __main__.py to improve usability


fatbasstard:
Currently there is a lot of logic implemented in `__main__.py` which is nice for CLI usage, but is not as handy when used in code.

It would improve the usability of this linter and makes it easier to integrate with other tooling if logic from this class is moved away. Focus should be on parsing the arguments, run the Rules and parse the output.

ATM the logic has to be copied into other tooling that implements this

cmmeyer:
Considering main() is now 5 lines of actual code, I think this is as skinny as we're going to get. Closing issue.

PR #80, opened 2018-05-20
Break out main and create a core sub module for running

### LINKED ISSUE #69: Test command line argument processing
Just had an issue reported that would have hopefully been caught with testing of the parameters and how they are processed. 

Add testing so we can make sure we are honoring the parameters correctly.

kddejong:
*Issue #, if available:* 
Fixes #69
#47 

*Description of changes:*
- Break out __main__ logic into a new sub module core
- Write tests around the new core functions
- Rename Parser to cfn_yaml


fatbasstard:
Also helps for https://github.com/awslabs/cfn-python-lint/issues/47

src/cfnlint/core.py
```diff
@@ -0,0 +1,365 @@
...
+            if config_name == 'log_level':
+                if isinstance(config_value, (six.string_types, six.text_type)):
+                    defaults['log_level'] = config_value
+
+    return defaults
+
+
+def run_checks(filename, template, rules, transforms, ignore_checks, regions):
+    """Run Checks against the template"""
+    if regions:
```
cmmeyer: Does it make sense to break out supported regions into a constant?  That way we can re-use elsewhere and easily update if more regions are added.

kddejong: Most recent change should cover this.  We were also setting a Regions variable in helpers and so I removed this one and used that one.

cmmeyer: Ha! No wonder I had the idea to break them out in my head -- we already had! :)  Looks good.

👍

PR #83, opened 2018-05-23
Fix/cross types

### LINKED ISSUE #42: Remove try... convert of integers (E3002)
In the property Type check, when the property value should be an integer (but it's not), there is an attempt to convert it:

https://github.com/awslabs/cfn-python-lint/blob/master/src/cfnlint/rules/resources/properties/Properties.py#L56

Since both YAML and JSON just handle these types out the box, the `try..except` can in theory be removed (since the type's don't match, a string is not an integer).

Personally I like the check to be more "strict', who do other think?  

Example (JSON): 
https://github.com/awslabs/cfn-python-lint/blob/master/test/templates/quickstart/vpc-management.json#L816

Example (YAML):
https://github.com/awslabs/cfn-python-lint/blob/master/test/templates/good/resource_properties.yaml#L464

These values can easily be written as actual integers
```
      - CidrIp:
          Ref: SSHLocation
        FromPort: 22 
        IpProtocol: tcp
        ToPort: 22
```

kddejong:
*Issue #, if available:*
Fixes #42 

*Description of changes:*
- Check primitive types to be more strict.  Check if strings are strings, integers are integers, booleans are booleans, etc.  
Example: cfn-lint will now fail if the value is supposed to be a boolean and its a string like "true",  "True", "false", "False"


cmmeyer:
Merging now, but let's be very explicit about this change and how to disable it in the release notes -- it really does throw errors on a lot of perfectly usable templates.

Issue #86, opened 2018-05-24
AWS::NoValue without quotes causes error: "Template is malformed: found unexpected ':'"


alytle:
I'm getting an error with this template, but CFN builds it fine, and the js cfn-linter claims it has no errors or warnings.

**Debug Logs:**
```
$ cfn-lint --template templates/rds.yaml --log-level debug
2018-05-24 10:54:46,166 - cfnlint - ERROR - Template templates/rds.yaml is malformed: found unexpected ':'
2018-05-24 10:54:46,166 - cfnlint - ERROR - Tried to parse templates/rds.yaml as JSON but got error: Expecting value: line 1 column 1 (char 0)
```

If I comment out the `MonitoringInterval` and `MonitoringRoleArn` in the `rDBServerInstance` resource, the error seems to go away. So it seems related to my !If statements.

These lines appear to be the issue:
```
MonitoringInterval: !If [ EnhancedMonitoring, !Ref pProdMonitoringInterval, !Ref pDevMonitoringInterval ]
MonitoringRoleArn: !If [ EnhancedMonitoring, !GetAtt [ rDBMonitoringRole, Arn ], !Ref AWS::NoValue ]
```

In fact, if I comment out just `MonitoringRoleArn` I get an entirely different failure.

**Debug Logs:**
```
cfn-lint --template templates/rds.yaml --log-level debug
2018-05-24 11:02:20,595 - cfnlint - DEBUG - Transform templates if needed
2018-05-24 11:02:20,596 - cfnlint - DEBUG - Run scan of template...
...
2018-05-24 11:02:20,606 - cfnlint - DEBUG - Starting match function for rule E3002 at 2018-05-24 11:02:20.606796
2018-05-24 11:02:20,606 - cfnlint - DEBUG - Get names of all parameters from template...
2018-05-24 11:02:20,607 - cfnlint - DEBUG - Get resources from template...
2018-05-24 11:02:20,607 - cfnlint - DEBUG - Get condition values...
2018-05-24 11:02:20,607 - cfnlint - DEBUG - Get condition values...
2018-05-24 11:02:20,607 - cfnlint - DEBUG - Complete match function for rule E3002 at 2018-05-24 11:02:20.607507.  Ran in 0:00:00.000713
[...]

PR #87, opened 2018-05-24
update to more blindly pass conditions back to the primitive type check

kddejong:
*Issue #, if available:*
#86

*Description of changes:*
- Sends objects forward from conditions back to the primitive type check.


cmmeyer:
Heya, I'm having a hard time following this one -- not fully understanding the code change and seeing the same results as issue #86 for that template.

kddejong:
@cmmeyer there are two issues in this template.  One is related to AWS::NoValue not having quotes and how that is handled with the parser.  Second is if you put quotes around AWS::NoValue it still throws an error.  There was technically a few things here but they got masked into one fix.   Condition Values was returning a response that wasn't appropriate for the logic.  Since this check is focused on basic structure checks we are just going to check each condition path against the same primitive check.

kddejong:
I hate to do this but I shouldn't have done this under master (accident) so I'm going to redo this pull request.

PR #101, opened 2018-06-01
Fix type checking of nested functions

fatbasstard:
Issue 42

Hotfix on the type checking. Got an error the following:

```
  ApplicationLoadBalancer:
    Type: "AWS::ElasticLoadBalancingV2::LoadBalancer"
    Properties:
      SecurityGroups:
        - !If [IsAcceptance, !ImportValue "import-something", !Ref "AWS::NoValue"]
```
The "!ImportValue" is not a string, which is true.

Not sure if it's fixable in the `check_values()` method, for now hotfixed it in the type checking. 

The reasons that there's 1 less failure in the test is here: https://github.com/awslabs/cfn-python-lint/blob/master/test/templates/quickstart/vpc.json#L489 (Which is valid but broke on the function)

(Also Validated against our codebase)

### DIFF
--- a/src/cfnlint/rules/resources/properties/ValuePrimitiveType.py
+++ b/src/cfnlint/rules/resources/properties/ValuePrimitiveType.py
@@ -43,7 +43,8 @@ class ValuePrimitiveType(CloudFormationLintRule):
         """Check Value"""
         matches = list()
         primitive_type = kwargs.get('primitive_type', {})
-        if isinstance(value, dict) and primitive_type == 'Json':
+
+        if isinstance(value, dict):
             return matches
         if primitive_type in ['String']:
             if not isinstance(value, (str, six.text_type, six.string_types)):


kddejong:
I think there is an underlying problem with how I wrote some of these core functions to make life easier.  I found it last week with an issue, solved it, but it was only a mask to the real issue.

Looking at these functions more I am getting annoyed with myself.  The results are inconsistent.  ImportValue in a condition should have never called check_value (in my mind) but it does because get_condition_values acts differently when there is a condition.

For instance this test is a great example.  
https://github.com/awslabs/cfn-python-lint/blob/master/test/module/test_template.py#L129

If its a Ref you get 
```
{'Path': [1, 'Ref'], 'Value': 'Sample'}, {'Path': [2], 'Value': 'String'}
```
and if its not a Ref you get
```
{'Path': [1], 'Value': {'ImportValue': 'Sample'}, {'Path': [2], 'Value': 'String'}
```

I think that inconsistency is hard to deal with down the line.  get_values has the same issue.  
I think being consistent with the results in these underlying functions will help a lot in fixing a lot of these issues that keep popping up.

The result is I have a much bigger change that gets the same result while drinking way too much coffee so I could wrap my head around this.

I like these core functions as then people don't have to account for all this crazy logic in their rules but they need a lot more testing and understanding of what the results should be in simple and complex situations.  If anything hangs up my PR it will be this.

fatbasstard:
Exactly what I encountered. Love all the helper function, they make life a lot easier... But they're hard to  read/debug/fix if anything goes "wrong".

Couldn't fit it in my head today, so I kept those function out of scope and came up with a "hotfix"... Which kept the test working, and applies nicely to all our stacks we have. 

If this can be fixed structurally, +1 of course, but if this (as a hotfix) is doable for now would be great. (Else we'd have to disable this rule.... And I like it :P )

kddejong:
all that and the changes around the test are nearly the same.
https://github.com/awslabs/cfn-python-lint/pull/102/files#diff-77812654eb16a303e2ac67e292cc70d0R53

fatbasstard:
Looks like your PR indeed might fix this... Let's wait for untill that one is merged :+1:

cmmeyer:
Just merged #102 -- let me know if that supersedes this PR,  @fatbasstard

fatbasstard:
This fixed this error indeed 👍  Now gives another error, but I'll create a new Issue/PR for this

PR #102, opened 2018-06-02
Update conditions and get values to return full objects

kddejong:
*Issue #, if available:*
Should replace #101 

*Description of changes:*
- Get full objects returned from conditions and get value functions.

### DIFF
--- a/src/cfnlint/__init__.py
+++ b/src/cfnlint/__init__.py
@@ -17,6 +17,7 @@
 import logging
 import sys
 import os
+import re
 from datetime import datetime
 from yaml.parser import ParserError
 import cfnlint.helpers
@@ -621,8 +622,7 @@ class Template(object):
                                 matches.extend(results)
                         elif sub_key == 'Ref':
                             if sub_value != 'AWS::NoValue':
-                                result['Value'] = sub_value
-                                result['Path'] += ['Ref']
+                                result['Value'] = item
                                 matches.append(result)
                         else:
                             # Return entire Item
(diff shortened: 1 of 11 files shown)


src/cfnlint/__init__.py
```diff
@@ -655,28 +655,33 @@ def get_values(self, obj, key, path=[]):
...
                         results = self.get_condition_values(obj_value, path[:] + [obj_key])
                         if isinstance(results, list):
                             matches.extend(results)
-                    else:
-                        result = {}
-                        result['Path'] = path[:] + [obj_key]
-                        result['Value'] = obj_value
-                        matches.append(result)
+
+                if not is_condition:
```
cmmeyer: You set is_condition statically above and I'm not seeing a path where is_condition can go "True" in this function .  Is this just to match the general pattern in the files below?

src/cfnlint/__init__.py
```diff
@@ -758,39 +765,42 @@ def check_value(self, obj, key, path,
...
+            if not isinstance(value, dict):
                 if check_value:
                     matches.extend(
                         check_value(
                             value=value, path=new_path[:] + child_path, **kwargs))
+            else:
+                if len(value) == 1:
+                    for dict_name, _ in value.items():
+                        if dict_name in cfnlint.helpers.FUNCTIONS:
+                            function_name = 'check_%s' % camel_to_snake(dict_name.replace('Fn::', ''))
```
cmmeyer: Ugh.  We already discussed this.  Still, ugh. :)
Could you maybe throw a comment in here for when I inevitably forget what we discussed?

Issue #95, opened 2018-05-30
Unused Mapping Warning does not work with nested FindInMap


fatbasstard:
Related to issue https://github.com/awslabs/cfn-python-lint/issues/57

When a nested FindInMap is used, it raises a warning on the mapping not used.

Example:

Resource part:
```
CidrBlock: !FindInMap [!FindInMap ["Environment", !Ref "Environment", "Subnets"], "eu-west-1a", "Management"]
```
Mapping:
```
Mappings:
  Environment:
    acceptance:
      Subnets: "AcceptanceSubnets"
    testing:
      Subnets: "TestingSubnets"
    production:
      Subnets: "ProductionSubnets"
  AcceptanceSubnets:
    eu-west-1a:
      Management: "XXX.XXX.XXX.XXX/XX"
...
```

This raises the warning: `W7001 - Mapping AcceptanceSubnets not used (line: 90)`

kddejong:
Ha, thought this was going to be easy but yea this is a tricky one.  Since `!FindInMap ["Environment", !Ref "Environment", "Subnets"]` actually ends up resolving the map name so we can validate if its used.  I would assume the same issue would occur for
```CidrBlock: !FindInMap [!Ref MappingName, "eu-west-1a", "Management"]```

Options.
1. Back-off the warning based on the knowledge of a complicated situation in which it becomes hard to track.
1.  Try to find a way to resolve the possible "values" that may come back from FindInMap function.  This could get tricky quick depending on where people put Refs and the values in the map.  

So it seems like the quick fix is to do 1 and more inline with what our typical approach has been.  Typical approach being to not raise warnings or errors for situations in which we can't definitely resolve what the value may be.

salexpdx:
I hit the W7001 warning in a slightly different situation but since you're looking at that error here, I'll add my note and can create a separate issue if you prefer.  Many of my maps are based on account number as the map name and region as the top level key. That causes me to see the mapping not used error.
An example: 
!FindInMap [!Ref "AWS::AccountId", !Ref "AWS::Region", VPCIPRange ]

kddejong:
Released with the newest version.  Worth noting this check will get disabled when Refs and FindInMaps are used for the mapping name in a FindInMap.

PR #103, opened 2018-06-02
Fix/find in map

kddejong:
*Issue #, if available:*
#95

*Description of changes:*
- Disables mappings used check when findinmap have complex situations around the mapping name


src/cfnlint/rules/mappings/Used.py
```diff
@@ -29,25 +29,32 @@ def match(self, cfn):
...
             maptrees = cfn.search_deep_keys('Fn::FindInMap')
-
             for maptree in maptrees:
 
                 if isinstance(maptree[-1], list):
-                    ref_mappings.append(maptree[-1][0])
+                    map_name = maptree[-1][0]
+                    if isinstance(map_name, dict):
+                        self.logger.debug(
+                            'Mapping Name has a function can have too many variations. '
```
cmmeyer: 'Mapping Name has a function THAT can have too many variations. '

PR #106, opened 2018-06-04
Handle empty yaml files

fatbasstard:
Already handled correctly for json files

Issue https://github.com/awslabs/cfn-python-lint/issues/104

yaml loader loads empty string as `None`, convert these to an empty dict so it's handled correctly.


kddejong:
seems like an easy fix this way and a good safe guard for empty files.

PR #115, opened 2018-06-12
Make Parseable output really parseable

fatbasstard:
The `ParseableFormatter` is not really parseable at the moment since the Error ID is in the Error message.

Since it's not really used ATM (As far as I could find out) I changed the current one. This makes the output actually easy to parse...


kddejong:
I can be ok with this.  Technically the most parseable output is the JSON output 😄

fatbasstard:
Yup,  normally it is. But for SublimeLinter a regex is/can be used: https://github.com/fatbasstard/aws-cfn-lint-sublime/blob/master/linter.py#L22

So splitting the output in a simpler format makes it simple

PR #116, opened 2018-06-12
Added the amount of resources

fatbasstard:
Issue https://github.com/awslabs/cfn-python-lint/issues/75

Add the amount of resources so it's clear how many resources there are

### DIFF
--- a/src/cfnlint/rules/resources/LimitNumber.py
+++ b/src/cfnlint/rules/resources/LimitNumber.py
@@ -35,7 +35,7 @@ class LimitNumber(CloudFormationLintRule):
         # Check number of resources against the defined limit
         resources = cfn.get_resources()
         if len(resources) > LIMITS['resources']['number']:
-            message = 'Maximum number of resources ({0}) exceeded'
-            matches.append(RuleMatch(['Resources'], message.format(LIMITS['resources']['number'])))
+            message = 'The number of resources ({0}) exceeds the limit ({1})'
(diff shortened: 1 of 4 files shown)


src/cfnlint/rules/resources/LimitNumber.py
```diff
@@ -35,7 +35,7 @@ def match(self, cfn):
         # Check number of resources against the defined limit
         resources = cfn.get_resources()
         if len(resources) > LIMITS['resources']['number']:
-            message = 'Maximum number of resources ({0}) exceeded'
-            matches.append(RuleMatch(['Resources'], message.format(LIMITS['resources']['number'])))
+            message = 'Maximum number of resources ({0}) exceeded ({1})'
```
kddejong: Does the language seem backwards with the parameters?

Maximum number of resources (# of Resources in the template) exceeded (the Resource LIMIT)?
or how about
The number of resources (len(resources)) exceeded the limit of (resource limit)?

fatbasstard:
Good one, I'll clarify

cmmeyer:
We should probably do this for all limit rules (parameters, mappings, resources, and outputs).  

@fatbasstard  -- do you want to add them to this commit, or shall I merge this one then update the others?

fatbasstard:
I'll add then, on it

cmmeyer:
#78 should give you all the files that need changing.

PR #117, opened 2018-06-12
Implement Serverless Application Model to perform the transformations

fatbasstard:
*Issue https://github.com/awslabs/cfn-python-lint/issues/81*

At the moment the itself linter performs the transformation for Serverless resources.There are issues with the transformations, it's complex and it should not be in the linter. The linter should handle the linting (It's not the responsibility of the linter to do this).

The PR removes the code and uses the Serverless Application Model itself (just like the AWS-CLI does) to perform the transformation (https://github.com/awslabs/serverless-application-model)

- The linter only has to lint and not worry about transformations
- Transformation functionality is always "up to date"
- Less code and complexity

Along the way:
- FIxed the Serverless "good" test templates (that apparently are not good... :P)
- Removed specific Servicess code tests
- Create a `TransformError` rule (just like the `ParseError`)

Only issue with this ATM is that we're missing information on the failing resource, so it's always "line: 1" now.  

Oh... And SAM currently only works on Python 2.7 :(


src/cfnlint/__init__.py
```diff
@@ -822,21 +758,25 @@ class Runner(object):
...
         self.rules = rules
         self.filename = filename
         self.verbosity = verbosity
-        self.transforms = transforms
         self.cfn = Template(template, regions)
 
     def transform(self):
         """Transform logic"""
-        LOGGER.debug('Transform templates if needed')
+        LOGGER.debug('Transform templates if needed using SAM')
```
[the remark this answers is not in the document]
fatbasstard: Yeah, true. I'll strip that a bit

src/cfnlint/transform.py
```diff
@@ -0,0 +1,161 @@
...
+        """
+        Transform the Template using the Serverless Application Model.
+        """
+        matches = []
+
+        # Boto's mechanism is used to fetch the region. Since we know which we use
+        # to lint, make sure Boto loads that one.
+        # https://github.com/awslabs/serverless-application-model/blob/master/samtranslator/translator/arn_generator.py#L39
+        # https://github.com/awslabs/serverless-application-model/blob/master/tests/translator/test_translator.py#L120
+        old_region = os.environ.get('AWS_DEFAULT_REGION', '')
```
kddejong: Oh this kind of sucks... we have to rewrite an environment variable?  Is that what the one failed build pointed too?

src/cfnlint/transform.py
```diff
@@ -0,0 +1,161 @@
...
+
+                def warning(self, message):
+                    """ Ignore warnings from SAM """
+                    pass
+
+            with WarningSuppressLogger(parser.logging):
+                sam_translator.translate(sam_template=self._template, parameter_values={})
+        except InvalidDocumentException as e:
+            for cause in e.causes:
+                matches.append(cfnlint.Match(
```
kddejong: There may not be much we can do here but it would be nice to get into line and column numbers.  Maybe we can provide some help back to them on this part :)

src/cfnlint/transform.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+    def load_managed_policies(self):
+        """
+        Load the ManagedPolicies locally, based on the AWS-CLI:
+        https://github.com/awslabs/aws-sam-cli/blob/develop/samcli/lib/samlib/default_managed_policies.json
+        """
+        return cfnlint.helpers.load_resources('data/Serverless/ManagedPolicies.json')
+
+    def _replace_local_codeuri(self):
```
kddejong: huh thats too bad too.  Be nice if they provided a light weight transformer.  Again maybe we can do to improve their code and lighten ours.

fatbasstard: This is based on the "validate" step of the CLI, did not really feel the urge on improving their code ATM

src/cfnlint/transform.py
fatbasstard: This is literally the error on the Build, or when you don't have a region specified, check the links. ATM this is an issue in SAM itself :(

Alternative could fix this in Tox

src/cfnlint/transform.py
jfuss: @kddejong Like a [plugin](https://github.com/awslabs/serverless-application-model/issues/367)!? :) 

I agree through. Using the SAM Translator library is a little cumbersome to setup and use. I ran into this as well when integrating it into SAM CLI but haven't gotten the time to submit an issue on revamping how to use the code yet.

There is defiantly overlap here for us (SAM CLI) and this project to work on some of these things together and get these common actions into one place (cross pollination :) ). Would love to explore some of that as it would be a win for both of us (and others who may need or want the functionality).

For context on why we have this in the SAM CLI:
So the real purpose of this is for SAM CLI to be able to validate the template correctly. The SAM Translator expects a S3 Object (dictionary with bucket, key, version keys) or an S3 URI ([docs](https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#properties)) for the CodeUri. This is also true for the `AWS::Serverless::Api`'s `DefinitionUri` Property. In SAM CLI however, we want customers to use this CodeUri or DefinitionUri` to point to their code on their local machine. If we don't replace, then the validation will fail. The `package` command already does the replacing of these keys before a customer deploys their code and template, so this is not to break the mental model for the tool.

src/cfnlint/transform.py
fatbasstard: Removed this and covered it in TOX now.

src/cfnlint/transform.py
jfuss: At the moment, SAM requires a region. This is to allow for some of the wacky stuff we needed to do to hind some things away. SAM should probably just default to something sane (whatever that is). It makes sense to have the region required when in the service context when but not really when it is a library. Improvement for sure.

src/cfnlint/transform.py
jfuss: Line numbers would be cool. I can see this being a huge improvement for customers using SAM. Would you mind creating a [GH Issue](https://github.com/awslabs/serverless-application-model/issues) for it?

fatbasstard: I'll take care of that 👍

jfuss:
Hey, hope you guys didn't mind me jumping and commenting here. I am one of the engineers that works on SAM/SAM CLI.

Note: This project is awesome and soon will be fully SAM Compatible 🎉👍!

Issue #114, opened 2018-06-11
Exceptions handling 


kisst:
Based on the documentation in the Readme, and the https://github.com/awslabs/cfn-python-lint/blob/master/test/templates/good/core/config_parameters.yaml file.

I setup this list 
```
~/git/cfn-test$ yq r my-test-stack.yaml Metadata.cfn-lint
config:
  ignore_checks:
  - E2504
  - E2507
  - E2520
  - E2523
```
where yq is mikefarah's yq, similar to jq, but still I got all the errors marked bellow.

```
~/git/cfn-test$ cfn-lint --template my-test-stack.yaml | grep ^E
E2507 IAM Policy statement key Fn::If isn't valid
E2507 IAM Policy statement missing Effect
E2507 IAM Policy statement missing Action or NotAction
E2507 IAM Policy statement missing Resource or NotResource
E2523 Only one of [VirtualName, Ebs] should be specified for Resources/FileServer/Properties/BlockDeviceMappings/1
E2504 Iops shouldn't be defined for type gp2 for Resources/FileServer/Properties/BlockDeviceMappings/0/Ebs/Iops
E2520 Parameter MasterUsername should NOT exist with SourceDBInstanceIdentifier for Resources/DB/Properties
E2520 Parameter MasterUserPassword should NOT exist with SourceDBInstanceIdentifier for Resources/DB/Properties
E2520 Parameter MasterUsername should NOT exist with DBSnapshotIdentifier for Resources/DB/Properties
E2520 Parameter MasterUserPassword should NOT exist with DBSnapshotIdentifier for Resources/DB/Properties
```

and 
```
~/git/cfn-test$ cfn-lint --version
cfn-lint 0.2.1
```

This whole list is needed only just to go around this for now https://github.com/awslabs/cfn-python-lint/issues/112

kddejong:
Made some good progress on error #112 this weekend but still trying to flush it out and get it well tested.  

Previous:
```
yq r test/templates/bad/properties_required.yaml Metadata.cfn-lint
null
```
```
$ cfn-lint --template test/templates/bad/properties_required.yaml | grep E3012 | wc -l
      10
```

Update the template.
```
$ yq r test/templates/bad/properties_required.yaml Metadata.cfn-lint
config:
  ignore_checks:
  - E3012
  - E2541
```
```
cfn-lint --template test/templates/bad/properties_required.yaml | grep E3012 | wc -l
       0
```

[...]

PR #119, opened 2018-06-13
Simplify ignore rules logic instead of passing it all over

kddejong:
*Issue #, if available:*
#114 - As a result of investigating this issue I wanted to make the ignore rules/checks functionality simpler.  
*Description of changes:*
- Instead of passing ignore checks through to the Rules runner filter the rules collection on init so all future rules are automatically filtered.

### DIFF
--- a/src/cfnlint/__init__.py
+++ b/src/cfnlint/__init__.py
@@ -182,12 +182,19 @@ class TransformsCollection(object):
 class RulesCollection(object):
     """Collection of rules"""
 
-    def __init__(self):
+    def __init__(self, ignore_rules=None):
         self.rules = []
+
+        # Make Ignore Rules not required
+        if ignore_rules:
+            self.ignore_rules = ignore_rules
+        else:
+            self.ignore_rules = []
 
     def register(self, obj):
         """Register rules"""
-        self.rules.append(obj)
+        if self.is_rule_enabled(obj):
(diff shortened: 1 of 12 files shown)


src/cfnlint/__init__.py
```diff
@@ -197,30 +203,33 @@ def __len__(self):
...
+        for rule in more:
+            if self.is_rule_valid(rule):
+                self.rules.append(rule)
 
     def __repr__(self):
         return '\n'.join([rule.verbose()
                           for rule in sorted(self.rules, key=lambda x: x.id)])
 
-    def resource_property(self, filename, cfn, ignore_checks, path, properties, resource_type, property_type):
+    def is_rule_valid(self, rule):
```
fatbasstard: `Valid` sound a bit weird. Can the naming be improved a bit? `is_rule_enabled` or something?

test/rules/__init__.py
```diff
@@ -50,7 +50,7 @@ class BaseRuleTestCase(BaseTestCase):
 
     def setUp(self):
         """Setup"""
-        self.collection = RulesCollection()
+        self.collection = RulesCollection([])
```
fatbasstard: It's defaulted, why not keep it `RulesCollection()`? Keep the "no excluded rule" logic clean (in the constructor).

Also for all other places with this change of course

Issue #112, opened 2018-06-08
False errors given with property relationships when conditions are used


kddejong:
Right now we check for attributes that should or shouldn't be defined together.  These scenarios can get tricky with conditions and AWS::NoValue. 

- [x] Short term hotfix to not give errors when complex situations are found
- [x] Ability to analyze if a related resource is available when conditions are used
- [x] Pull out values of properties in an object when values include conditions
- [x] Related objects and their values when each object has a condition applied

Related issues:
- https://github.com/awslabs/aws-cfn-lint-atom/issues/8

kddejong:
Working to invalidate pull request #121.  When the condition numbers are high the scenarios get large and can cause linting to be slow.
Working to replace with #352 which is more light weight and less costly.  

As it exists #121 will build a list of scenarios for the conditions.   That logic may have to be moved to a new rule that checks if Ref/GetAtt/DependsOn resources exist when using resource level conditions. 

Conditions when used in place of a resource property have to be singular.
```
Properties:
   Fn::If: [myCondition, { "ImageId": "ami-123456"}, {"LaunchConfiguration": "myLaunch"}]
```
is valid but the following is invalid because you cannot put an Fn::If with other properties.   (Still double checking this)
```
Properties:
  LaunchConfiguration: myLaunch
  Fn::If: [myCondition, { "ImageId": "ami-123456"}, {"Ref": "AWS::NoValue"}]
```

This means our Inclusive and Exclusive checks will be easier to write. If someone puts a condition in the properties section there isn't a mixing of properties from inside and outside of the condition that will cause confusion.

kddejong:
starting with pull request #523 

Next will be to get property values when conditions are used for values of those properties.

ap-hyperbole:
Facing a similar issue with E2540 linting error.

Error :

>  E2540 At least one stage in pipeline must contain an action that is not a source action.

Template 


`  Pipeline:
    Type: AWS::CodePipeline::Pipeline
    Properties:
      RoleArn: !GetAtt CodePipelineServiceRole.Arn
      ArtifactStore:
        Location: !Ref PipelineBucket
        Type: S3
      Stages:
        - Name: Source
          Actions:
          - Name: Source
            ActionTypeId:
              Category: Source
              Owner: AWS
              Version: "1"
              Provider: S3
            Configuration:
              S3Bucket: !Ref ThisEnvBucket
              S3ObjectKey: !Sub ${ProjectName}.zip
[...]

PR #121, opened 2018-06-13
Process templates playing out condition scenarios

kddejong:
*Issue #, if available:*
#112 

*Description of changes:*
- Play out the condition scenarios to make testing conditions easier.

### DIFF
--- a/test/module/cfn_json/test_cfn_json.py
+++ b/test/module/cfn_json/test_cfn_json.py
@@ -15,7 +15,7 @@
   SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 """
 import json
-from cfnlint import Template, RulesCollection  # pylint: disable=E0401
+from cfnlint import Template, RulesCollection, Runner  # pylint: disable=E0401
 from cfnlint.core import DEFAULT_RULESDIR  # pylint: disable=E0401
 import cfnlint.decode.cfn_json  # pylint: disable=E0401
 from testlib.testcase import BaseTestCase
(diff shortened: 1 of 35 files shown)


src/cfnlint/data/AdditionalSpecs/Exclusive.json
```diff
@@ -53,12 +53,6 @@
       ]
     },
     "AWS::RDS::DBInstance": {
-      "DBSnapshotIdentifier": [
```
kddejong: This needs to be converted to another check as DBSnapshotIdentifier can be an empty string which means CloudFormation will use the Username and Password fields.

test/module/cfn_json/test_cfn_json.py
```diff
@@ -64,10 +66,11 @@ def test_success_parse(self):
             filename = values.get('filename')
             failures = values.get('failures')
             template = json.load(open(filename), cls=cfnlint.cfn_json.CfnJSONDecoder)
-            cfn = Template(filename, template, ['us-east-1'])
-
+            runner = Runner(
```
kddejong: Switch to Runner which removes duplicates.  Important part of running the processor now.  We don't need duplicate alerts for something that failed across all the scenarios.

src/cfnlint/transform.py
```diff
@@ -103,7 +108,7 @@ def warning(self, message):
                     pass
 
             with WarningSuppressLogger(parser.logging):
-                sam_translator.translate(sam_template=self._template, parameter_values={})
+                self._template = sam_translator.translate(sam_template=self._template, parameter_values={})
```
kddejong: When doing a full deepcopy you need to be able to save the output of this function so you can consume it later.

kddejong:
Closing this as we have taken a different approach.  I still will maintain this code until we can finalize our picture on handling conditions.

PR #126, opened 2018-06-14
Add overview of all rules to the documentation

fatbasstard:
Added/generated al list of all the rules present in the toolkit. Cleaned up the Readme a bit by moving some of the extended documentation to a separate `docs` folder

Also added an `--update-documentation` to generate the rules documentation (Kept the argument a bit more "general" in case there is other documentation we can generate in the future. 
Hiding it from the help (`-h`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ cfn-lint applies the configuration from the CloudFormation Metadata first and th
 ## Rules
 This linter checks the CloudFormation by processing a collection of Rules, where every rules handles a specific function check or validation of the template.
 
-This collection of rules can be extended with custom rules using the `--apopend-rules` argument.
+This collection of rules can be extended with custom rules using the `--append-rules` argument.
 
 More information describing how rules are set up and an overview of all the Rules that are applied by this linter are documented [here](docs/rules.md)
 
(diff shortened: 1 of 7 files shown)


kddejong:
It would be awesome to have a way to update these rule docs automatically... may need to think about that going forward.

fatbasstard:
I've got a generation script in the codebase on my pc... It has some identical behaviour like the updating of the specification. It's not really related to linter, but more "Maintenance" on it  (Run a command and make a PR of the outcome :P)

I'll let that sink for a moment 🤔

README.md
```diff
@@ -43,8 +43,7 @@ There are IDE plugins available to get direct linter feedback from you favorite
 | --append-rules | append_rules | [RULESDIR [RULESDIR ...]] | Specify one or more rules directories using one or more --append-rules arguments. |
 | --ignore-checks | ignore_checks | [IGNORE_CHECKS [IGNORE_CHECKS ...]] | Only check rules whose ID do not match or prefix these values.  Examples: <br />- A value of `W` will disable all warnings<br />- `W2` disables all Warnings for Parameter rules.<br />- `W2001` will disable rule `W2001` |
 | --log-level | log_level | {info, debug} | Log Level |
-| --update-specs | | | Update the CloudFormation Specs.  You may need sudo to run this.  You will need internet access when running this command |
```
cmmeyer: I disagree with hiding/suppressing the update options.  Being able to manually update the resource spec is hugely useful for folks who want to quickly adopt a new service. Some folks want access the minute they are available.

src/cfnlint/core.py
```diff
@@ -176,7 +177,11 @@ def append_parser(parser, defaults):
         version='%(prog)s {version}'.format(version=__version__)
     )
     parser.add_argument(
-        '--update-specs', help='Update the CloudFormation Specs',
+        '--update-specs', help=argparse.SUPPRESS,
```
cmmeyer: Same here. Why suppress? Give the folks the option to update if they add local rules or want the latest spec.

fatbasstard: That still works, just "hidden it", put it back to what it was?

PR #131, opened 2018-06-17
Add some IAM limits checks

seporaitis:
*Issue #, if available:* N/A

*Description of changes:* Add checks for _some_ of IAM limits, as defined in [documentation](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-limits.html). Mainly those, which can be determined by looking at an individual templates:

- managed policies attached to IAM user, group or role - 10
- groups IAM user can be a member of - 10
- roles in an instance profile - 1
- `AssumeRolePolicyDocument` size - <2048 characters

The ID `2508` was picked semi-arbitrarily. It is close to `2507` - another IAM related check.

Associated tests are also added and pass. Locally I get linter errors:

```
Using config file /home/julius/code/cfn-python-lint/pylintrc
************* Module cfnlint.transform
E: 19, 0: Unable to import 'samtranslator.parser' (import-error)
E: 20, 0: Unable to import 'samtranslator.translator.translator' (import-error)
E: 21, 0: Unable to import 'samtranslator.public.exceptions' (import-error)

-----------------------------------
Your code has been rated at 9.96/10
```

but since they are not related to files I changed I presume it is part of `master` branch or my local setup.

The motivation for this change is that it comes from learning the managed policies limit the hard-way - a CloudFormation stack was stuck in `UPDATE_IN_PROGRESS` for about 40-odd minutes, before it timed out with message below and rolled back:

```
Cannot exceed quota for PoliciesPerGroup: 10 (Service: AmazonIdentityManagement; Status Code: 409; Error Code: LimitExceeded
```

Having this check would have saved some time for me and hopefully can save for someone else.

I hope you will consider this as improvement and I look forward to feedback and I will update this as you think appropriate to be merged into `master`.

### DIFF
--- a/src/cfnlint/rules/resources/iam/Limits.py
+++ b/src/cfnlint/rules/resources/iam/Limits.py
@@ -23,7 +23,7 @@ from cfnlint import RuleMatch
 
 class Limits(CloudFormationLintRule):
     """Check if IAM limits are not breached"""
-    id = '2508'
+    id = 'E2508'
     shortdesc = 'Check IAM resource limits'
     description = 'See if IAM resources do not breach limits'
     tags = ['base', 'resources', 'iam']
--- a/src/cfnlint/rules/resources/iam/Limits.py
+++ b/src/cfnlint/rules/resources/iam/Limits.py
@@ -1,0 +1,132 @@
+"""
+  Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+  Permission is hereby granted, free of charge, to any person obtaining a copy of this
+  software and associated documentation files (the "Software"), to deal in the Software
+  without restriction, including without limitation the rights to use, copy, modify,
+  merge, publish, distribute, sublicense, and/or sell copies of the Software, and to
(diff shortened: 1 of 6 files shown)


fatbasstard:
Local linting error a thing of Tox, it doesn't automatcally recreate Virtual Environments: http://tox.readthedocs.io/en/latest/example/basic.html#forcing-re-creation-of-virtual-environments

src/cfnlint/rules/resources/iam/Limits.py
```diff
@@ -0,0 +1,132 @@
...
+import datetime
+import json
+
+from cfnlint import CloudFormationLintRule
+from cfnlint import RuleMatch
+
+
+class Limits(CloudFormationLintRule):
+    """Check if IAM limits are not breached"""
+    id = '2508'
```
fatbasstard: Rules should start with an `E` for errors or a `W` for warnings. I think you're missing the E here ;)

cmmeyer:
Great new rule!  Thanks!

PR #145, opened 2018-06-20
Add filename to the YAML Parse error output

fatbasstard:
*Issue https://github.com/awslabs/cfn-python-lint/issues/142*

Pass along the filename when parsing the yaml so it's outputted in the error (instead of '')


kddejong:
Is this just a YAML issue? Will JSON output filename?

ashemedai:
@kddejong Frank showed me the JSON output, the formatting is of the linter output is quite different there, so this doesn't necessarily apply one-on-one as I thought for the JSON case.

PR #160, opened 2018-06-25
Resource names should only contain alphanumeric

jvanbrunschot-coolblue:
*Description of changes:*
CloudFormation Resource names can only contain alphanumeric. As stated here: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resources-section-structure.html

Same goes for Outputs:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/outputs-section-structure.html

and Parameters:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html

and Mappings:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/mappings-section-structure.html

and according the following error when uploading a CloudFormation template:
`Template validation error: Template format error: Resource name SNS_TOPIC is non alphanumeric.`


cmmeyer:
Good catch!

PR #166, opened 2018-06-26
Support multiple templates per invocation

### LINKED ISSUE #149: Support checking multiple files at once
Please support an arbitrary number of files passed in on the commandline. For example:

```
cfn-lint template1.yaml template2.yaml
```

alexjurkiewicz:
(fixes #149)

cc @asottile


src/cfnlint/core.py
```diff
@@ -93,9 +93,12 @@ def create_parser():
 
     # Alllow the template to be passes as an optional or a positional argument
     parser.add_argument(
-        'template', nargs='?', help='The CloudFormation template to be linted')
+        'templates', nargs='*', default=[],
+        help='The CloudFormation templates to be linted')
     parser.add_argument(
-        '-t', '--template', dest='template_alt', help='The CloudFormation template to be linted')
```
kddejong: I have to check all the editor plugins.  I know for atom, vs code this will break them.  I would like to keep `--template` available until we can get them converted to `-t-` or `--templates` and give a little bit of time for people to update.

src/cfnlint/core.py
```diff
@@ -267,7 +266,14 @@ def get_template_args_rules(cli_args):
         parser.print_help()
         exit(1)
 
-    return(args, filename, template, rules, fmt, formatter)
+    return (
```
kddejong: Line 264 above needs to be changed.
`if not filename` should be `if not filenames`?

src/cfnlint/__main__.py
```diff
@@ -24,13 +24,13 @@
...
+    (args, files, rules, fmt, formatter, exit_code) = cfnlint.core.get_template_args_rules(sys.argv[1:])
 
-    return(
-        cfnlint.core.run_cli(
-            filename, template, rules, fmt,
-            args.regions,
-            args.override_spec, formatter))
+    for filename, template in files.items():
+        exit_code |= cfnlint.core.lint_file(filename, template, rules, fmt,
+        args.regions, args.override_spec, formatter)
```
kddejong: pylint style here.  Need the indentation aligned to the open parans on line 30.

alexjurkiewicz:
I have a usability issue. There's poor interaction between `-a`/`-i` and positional template specification.

The issue is a commandline like `cfn-lint -i E001 test.yml`

Argparse parses this as two ignore rules, because the positional arguments are optional. You can fix parsing the usual way (`cfn-lint -i E001 -- test.yml`) but this would be very annoying.

We could fix this by making the positional argument mandatory. This would fix the above case but is a backwards compatibility break as discussed in #168. There would still be issues when specifying multiple files (`cfn-lint -i E001 test.yml test2.yml`) as the `-a`/`-i` would try to consume all but the last one, but I don't see a way around that.

`cfn-lint` has a few modes which don't require templates (update specs, update docs, list rules) which would be affected by this change. I propose those can be dealt with as special cases.

What do you think?

asottile:
Other tools handle this by allowing an argument to be specified more than once:

```
cfn-lint -i E001 -i E002 -i E003
```

or to use commas and parse after-the-fact

```
cfn-lint -i E001,E002,E003
```

alexjurkiewicz:
Following discussion in [Slack](awsdevelopers.slack.com):

1. Need the CLI to transition to supporting multiple templates slowly, so code editor integrations can catch up (so allow existing template specification methods to keep working)
2. A separate PR can be made to change how -r/-i/-a handling is done as per examples from asottile in previous comment

I'll open a new PR soon, because this one got a little side tracked. SIde note I'm going to be AFK for a few weeks soon so apologies in advance

PR #167, opened 2018-06-26
Add pre-commit

alexjurkiewicz:
No point merging this until `cfn-lint` supports multiple templates per invocation, see #166 for that

cc @asottile

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,5 @@
+-   id: cfn-python-lint
+    name: AWSLabs CloudFormation Linter
+    entry: cfn-lint
+    language: python
+    files: \.(json|yaml|yml)$
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ The linter provides the possibility to implement these customized specifications
 
 More information about how this feature works is documented [here](docs/customize_specifications.md)
 
-## pre-Commit
+## pre-commit
 If you'd like cfn-lint to be run automatically when making changes to files in your Git repository, you can install [pre-commit](https://pre-commit.com/) and add the following text to your repositories' `.pre-commit-config.yaml`:
 
 ```yaml
--- a/README.md
+++ b/README.md
@@ -109,5 +109,18 @@ The linter provides the possibility to implement these customized specifications
 
 More information about how this feature works is documented [here](docs/customize_specifications.md)
 
+## Pre-Commit
+If you'd like cfn-lint to be run automatically when making changes to files in your Git repository, you can install [pre-commit](https://pre-commit.com/) and add the following text to your repositories' `.pre-commit-config.yaml`:
+
+```yaml
+repos:
+-   repo: https://github.com/awslabs/cfn-python-lint
+    rev: v0.7.4 # The version of cfn-lint to use
+    hooks:
+    -   id: cfn-python-lint
+```
+
+You can see the available cfn-lint versions on the [releases page](https://github.com/awslabs/cfn-python-lint/releases).
+
 ## Credit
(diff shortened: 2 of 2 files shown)


.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,6 @@
+-   id: cfn-python-lint
+    name: AWSLabs CloudFormation Linter
+    entry: cfn-lint
+    language: python
+    files: \.(json|yaml|yml)$
+    language_version: python2
```
asottile: does it also work on python3? (if so can just drop this line I would think)

cmmeyer:
Is this now dependent on #178?  I assume we still want to hold off until the multi-file situation is resolved?

asottile:
Still needs #149 to be solved, yes

kddejong:
Now that parsing multiple files is there I wonder if this is valid again.  @alexjurkiewicz ?

alexjurkiewicz:
I think it is! This PR should probably also include instructions in README.md on how to use pre-commit though.

kddejong:
and just to make sure I have this correctly... that would include adding a file to a user's repo called `.pre-commit-config.yaml` with the following content?  Do I need to include rev?
```
repos:
-   repo: https://github.com/awslabs/cfn-python-lint
    rev: v0.7.5
    hooks:
    -   id: cfn-python-lint
```

asottile:
yep `rev` is necessary for `pre-commit` to know where to clone and [what revision to check out](https://pre-commit.com/#using-the-latest-version-for-a-repository) (fortunately in most cases you can just run `pre-commit autoupdate` and not really care about the value).

(unless there's a conventional place for cfn files, in which case this PR should probably set `files:` to that) it probably also would need something like:

```yaml
    -   id: cfn-python-lint
        files: ^my-cfn-dir/.*\.(yml|yaml|json)$
```

kddejong:
@alexjurkiewicz you have time to update the README or can I help with this?

alexjurkiewicz:
Next week I do, but if you're eager I don't mind you replacing this commit
😁

alexjurkiewicz:
Added documentation, squashed and rebased

README.md
```diff
@@ -109,5 +109,18 @@ The linter provides the possibility to implement these customized specifications
 
 More information about how this feature works is documented [here](docs/customize_specifications.md)
 
+## Pre-Commit
```
asottile: very very very very minor, `pre-commit` tends to be stylized lowercase

README.md
```diff
@@ -109,5 +109,18 @@ The linter provides the possibility to implement these customized specifications
...
+If you'd like cfn-lint to be run automatically when making changes to files in your Git repository, you can install [pre-commit](https://pre-commit.com/) and add the following text to your repositories' `.pre-commit-config.yaml`:
+
+```yaml
+repos:
+-   repo: https://github.com/awslabs/cfn-python-lint
+    rev: v0.7.4 # The version of cfn-lint to use
+    hooks:
+    -   id: cfn-python-lint
+```
```
asottile: I think you'll also want to mention:

```yaml
    files: path/to/cfn/dir/.*\.(json|yml|yaml)$
```

the default configuration will lint all yaml / json files in the repository (which probably (well definitely, since at least one is .pre-commit-config.yaml!) won't work).

README.md
asottile: /o\ both characters lower case

(that's me hiding, I shouldn't have said anything -- sorry for the inconvenience)

alexjurkiewicz:
d'oh not sure how I missed that 😹

cmmeyer:
Great use case! Let me get back on my laptop tonight and kick the tires as a sanity check, but it all seems reasonable via code review

PR #171, opened 2018-06-26
Use yaml C implementation when available

asottile:
I don't have any benchmarks except the testsuite, but that's a pretty good benchmark right?

### before
```
Ran 213 tests in 102.601s
```

### after
```
Ran 213 tests in 24.784s
```

### DIFF
--- a/src/cfnlint/cfn_yaml.py
+++ b/src/cfnlint/cfn_yaml.py
@@ -21,13 +21,19 @@ from yaml.composer import Composer
 from yaml.reader import Reader
 from yaml.scanner import Scanner
 from yaml.resolver import Resolver
-from yaml.parser import Parser
 from yaml import ScalarNode
 from yaml import SequenceNode
 from yaml import MappingNode
 from yaml.constructor import SafeConstructor
 from yaml.constructor import ConstructorError
 import cfnlint
+
+try:
+    from yaml.cyaml import CParser as Parser
+    cyaml = True
+except ImportError:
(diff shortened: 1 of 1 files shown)


kddejong:
Seems a lot better. Thanks!

asottile:
I can't see the build failure -- mind using a public service (such as travis-ci, etc.) such that the failure is visible?

kddejong:
yea I can take a look into that.  You should be able to catch this with tox locally too.
```
************* Module cfnlint.cfn_yaml
C: 32, 4: Imports from package yaml are not grouped (ungrouped-imports)

------------------------------------
Your code has been rated at 10.00/10
```

asottile:
oh I see `pylint36`

well this is a bit of a picky lint error, what does it want that _doesn't_ suck :'(

cmmeyer:
Yeah -- we're exposing a few of the "expected errors" in our tests.  We should figure out a way to hide those, but I don't think it's been a priority.

Looks like it's that ungrouped thing -- trying to keep our 10/10 in pylint :)

asottile:
I think this is a bug in pylint, going to open an issue there and locally disable that

fatbasstard:
Nice performance gain! :+1:

cmmeyer:
Approved! Preparing to merge.

A good follow up would be an output (via -d or some other mechanism) to show which parser is in use -- may help for to troubleshooting down the line.

PR #175, opened 2018-06-27
Add run to check for unneeded DependsOn usage

fatbasstard:
*Issue https://github.com/awslabs/cfn-python-lint/issues/151*

According to an AWS Blog post:
```
You can also introduce references between elements by using either the { "Ref": "MyResource" } or the { "Fn::GetAtt" : [ "MyResource" , "MyAttribute" ] } functions. 
When you use one of these functions, CloudFormation behaves as if you’ve added a DependsOn attribute to the resource. 
 ```
Source: https://aws.amazon.com/blogs/devops/optimize-aws-cloudformation-templates/

This rules add this check as a Warning `W3005` (Other `DependsOn` check is an error with the same number).

### DIFF
--- a/docs/rules.md
+++ b/docs/rules.md
@@ -27,7 +27,7 @@ Rule `E3012` is used to check the types for value of a resource property.  A num
 
 
 ## Rules
-The following **71** rules are applied by this linter:
+The following **72** rules are applied by this linter:
 
 | Rule ID  | Title | Description | Tags |
 | -------- | ----- | ----------- |----- |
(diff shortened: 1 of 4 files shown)


fatbasstard:
@cmmeyer Can you doublecheck if the behaviour described in that blog is really true?

cmmeyer:
Confirmed with the team.  The direct quote:
"I personally refer to Ref/GetAtt as implicit dependencies and DependsOn as explicit dependency.
If you have Ref/GetAtt then a DependsOn statement is superfluous."

PR #183, opened 2018-06-29
Extend rules with source URL

fatbasstard:
*WIP: Opened for comments and in-between reviews*

Extend the rules with a `Source`, which basically contains information about the reasoning behind the rules. This is an array of string-information: `Resource Specification` for Spec file based information, or a URL to documentation/blogs/whatever explaining the thing the rule tackles

Split up in separate commits now to make the individual steps visible


src/cfnlint/rules/parameters/Used.py
```diff
@@ -26,7 +26,8 @@ class Used(CloudFormationLintRule):
     id = 'W2001'
     shortdesc = 'Check if Parameters are Used'
     description = 'Making sure the parameters defined are used'
-    tags = ['base', 'parameters']
+    sources = ['Best Practice']
```
kddejong: I was hoping sources was going to stay as URLs.  We could add these in Atom as links for error messages.

fatbasstard: Agree, see if I can fit it in

fatbasstard:
Yeah, maybe the match level can combine the Rule sources and the Specification URLS (which are in there for all resources) and on that level build a list of documentation urls.

I do think the rule level is great when keeping customization in mind. Say `cfn-nag` will create a ruleset, or a company has it's own. That will make it really simple to explain WHY the rule is there

fatbasstard:
`I was thinking sources was going to be used on the Match level and not on the Rules level.`

Well, the rules are included in the Match, so the information is available

fatbasstard:
Some are hard to link to a url. Like unused parameters/outputs/mappings whatsover, or alll the rules driven by the Spec file. If there is no replacement for that, I'd like to stick to also supporting strings

kddejong:
When I was thinking about at the Match level we could change the source for each  match so we could include link to the properties documentation.  Any rule using the CloudFormation spec falls into this.  The source at the rule level is generic and still useful but being able to do it at the Match level we could a lot more specific links.  That being said I can see the value of a rule level `sources` value so this may be an additional task.

I still like the idea of keeping source as a URL.  Is it worth while to add a type to the sources?
`sources = ['https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/conditions-section-structure.html']`
to
```
sources = [
  {'type': 'url', 'value': 'https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/conditions-section-structure.html'},
  {'type': 'string', 'value': 'CloudFormation Spec' }
]
```

fatbasstard:
We should be able to push it to Url's... `CloudFormation Spec` could point to documentation in the linter repo itself. Same goes for the "Best Practises" for instance 🤔

PR #187, opened 2018-07-03
Start to optimize testing

kddejong:
*Issue #, if available:*

*Description of changes:*
- Start to optimize testing.  Only test quickstarts templates once against all rules validating the output.  Generic good templates also only get run once.  Going forward will use rule testing for specific rule items.  How does this rule handle scenario X, Y, and Z. Then using quickstarts or other public templates for full integration testing.

Reduction of time to execute tests by over 30 seconds per python environment.


fatbasstard:
`Files changed 226`

...

Let's take a look 😄

cmmeyer:
Nice speed boost.  Should make PRs more delightful. :)

PR #201, opened 2018-07-10
Fix select and cidr function logic

### LINKED ISSUE #199: E1024 should allow Fn::ImportValue
cfn-lint version: 0.3.3

The following template:

```
---
Parameters:
  SubnetIndex:
    Description: 'Index of the subnet'
    Type: Number
    MinValue: 0
    MaxValue: 5
  SubnetCount:
    Description: 'To slice the IP address ranges you need to specify how many subnets you want to create in the VPC'
    Type: Number
    MinValue: 1
    MaxValue: 6
Resources:
  Subnet:
    Type: 'AWS::EC2::Subnet'
    Properties:
      CidrBlock: !Select [!Ref SubnetIndex, !Cidr [{'Fn::ImportValue': 'vpc-CidrBlock'}, !Ref SubnetCount, 12]]
      VpcId: 'vpc-123456'
```

Produces the following exceptions and findings:

```
# not sure about this
E0002 Unknown exception while processing rule E1017: too many values to unpack
bug.yml:1:1

# looks like Cidr also supports Fn::ImportValue (at leats the template works)
E1024 Cidr ipBlock should be Cidr Range, Ref, or Select for Resources/Subnet/Properties/CidrBlock/Fn::Select/1/Fn::Cidr/0
bug.yml:18:7

# not quiet sure about that, 12 should be a int?
E1024 Cidr sizeMask should be a int for Resources/Subnet/Properties/CidrBlock/Fn::Select/1/Fn::Cidr/2
bug.yml:18:7
```

kddejong:
*Issue #, if available:*
Fixes #199 

*Description of changes:*
- Fix Select function check to support unpacking a dict in the first element
- Fix Cidr to support ImportValue and appropriately checking number for the last element


src/cfnlint/rules/functions/Select.py
```diff
@@ -73,10 +73,10 @@ def match(self, cfn):
                                     matches.append(RuleMatch(
                                         tree, message.format(key, '/'.join(map(str, tree)))))
                         else:
-                            message = 'Select should be a list of 2 elements for {0}'
+                            message = 'Select should be an object of 1 supported function at {0}'
```
cmmeyer: Does it make sense to list the supported functions in the message?

PR #211, opened 2018-07-13
Do an actual deepcopy of objects

kddejong:
*Issue #, if available:*

*Description of changes:*
- Fix deepcopy to not just be a copy
- Remove dict and list custom objects.  They aren't needed for location searching and add extra layers.

### DIFF
--- a/src/cfnlint/__init__.py
+++ b/src/cfnlint/__init__.py
@@ -18,6 +18,7 @@ import logging
 import sys
 import os
 import re
+from copy import deepcopy
 from datetime import datetime
 from yaml.parser import ParserError
 import cfnlint.helpers
@@ -343,6 +344,14 @@ class Template(object):
             'Outputs',
             'Rules'
         ]
+
+    def __deepcopy__(self, memo):
+        cls = self.__class__
+        result = cls.__new__(cls)
+        memo[id(self)] = result
+        for k, v in self.__dict__.items():
+            setattr(result, k, deepcopy(v, memo))
+        return result
 
     def get_resources(self, resource_type=[]):
         """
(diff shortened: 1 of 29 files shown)


src/cfnlint/__init__.py
```diff
@@ -344,6 +345,14 @@ def __init__(self, filename, template, regions=['us-east-1']):
             'Rules'
         ]
 
+    def __deepcopy__(self, memo):
```
cmmeyer: This code repeats four times -- is there a way to consolidate this into a base class or library?

src/cfnlint/cfn_json.py
```diff
@@ -119,10 +119,19 @@ def __new__(self, x, start_mark, end_mark):
             return cls.__new__(self, x)
 
         def __deepcopy__(self, memo):
-            return self
```
cmmeyer: Here

src/cfnlint/cfn_yaml.py
```diff
@@ -84,17 +84,25 @@ def __new__(self, x, start_mark, end_mark):
             return cls.__new__(self, x)
 
         def __deepcopy__(self, memo):
-            return self
```
cmmeyer: And here.

src/cfnlint/__init__.py
kddejong: In some ways yes and others no.  We need to support deepcopy on custom objects so this one will have to remain.

src/cfnlint/cfn_json.py
kddejong: The better option for this would be to make this class singular for the json and yaml parsers.  There isn't a reason to have them copied.

PR #220, opened 2018-07-17
Document Resource Specification based rules

fatbasstard:
Started documentation of the rules that are based on the Resource Specification.  With this documentation we can get rid of the rules without `source_url` (currently pointing to the root of this repo) and give more in-depth explanation rules so prevent discussions in the future (e.g. with the strict type checking).

Just did one as an example for now. Feedback on the approach is welcome!


fatbasstard:
@kddejong Check, I'll continue this approach then and fill in 1 rule at a time untill they all are covered.

I'll leave it up to you and @cmmeyer to merge it when needed. (This first can already be merged, I'll continue on the other rules along the way)

src/cfnlint/rules/resources/properties/Properties.py
```diff
@@ -25,7 +25,7 @@ class Properties(CloudFormationLintRule):
     shortdesc = 'Resource properties are valid'
     description = 'Making sure that resources properties ' + \
                   'are properly configured'
-    source_url = 'https://github.com/awslabs/cfn-python-lint'
+    source_url = 'https://github.com/awslabs/cfn-python-lint/blob/master/docs/cfn-resource-specification.md#properties'
```
cmmeyer: Oh! Now I understand the strategy here.  Cool.

PR #232, opened 2018-07-20
Add Route53 AAAA records check

fatbasstard:
*Issue https://github.com/awslabs/cfn-python-lint/issues/94*

Added checking of `AAAA` records (IPv6) and corrected and cleaned some of the code along the way.

Finding a "good" regex was a challenge since the IPv6 format is a bit more complex. Wanted to stick to a/the regex solution. Maybe some "bad" IP addresses may pass, but, most importantly, "good" addresses should never fail...

Ended up using the RegEx from here: http://home.deds.nl/~aeron/regex/ (Also used the test data there in the fixtures)

Used https://www.helpsystems.com/intermapper/ipv6-test-address-validation to do some extra validations


src/cfnlint/rules/resources/route53/RecordSet.py
```diff
@@ -64,6 +64,25 @@ def check_a_record(self, path, recordset):
...
+        matches = list()
+
+        if not recordset.get('AliasTarget'):
+            resource_records = recordset.get('ResourceRecords')
+            for index, record in enumerate(resource_records):
+
+                if isinstance(record, six.string_types):
+                    tree = path[:] + ['ResourceRecords', index]
+                    full_path = ('/'.join(str(x) for x in tree))
+                    print(record)
```
kddejong: Ooops. Missed this one.  Need to get it deleted.

PR #234, opened 2018-07-21
Append to spec to include relationship info for Ref and GetAtt

### LINKED ISSUE #34: Ref and GetAtt Specificiations
I've been writing some individual checks that would check the Ref or GeAtt of resource properties.

For example we would check the that a Ref for VpcId goes to a resource of AWS::EC2::VPC.  If it was to a parameter we would check the parameter is of appropriate type (String, AWS::EC2::VPC::Id)
```
Type: "AWS::EC2::Subnet"
Properties:
  CidrBlock: String
  VpcId: String
```

After writing a few of these I'm starting to see a lot of repeat and I'm wondering if this would be better defined in a JSON/YAML document.   We can then write a singular rule to check these items limiting repeated code.
An example of that document is as follows.
```
{
  "RelationshipTypes": {
    "String": {
      "Ref": {
        "Parameters": [
          "String"
        ]
      }
    },
    "VpcId": {
      "Ref": {
        "Resources": [
          "AWS::EC2::VPC"
        ],
        "Parameters": [
          "String",
          "AWS::EC2::VPC::Id"
        ]
      },
      "GetAtt": {
        "AWS::EC2::SecurityGroup": "VpcId"
      }
    }
  },
  "ResourceTypes": {
    "AWS::EC2::Subnet": {
      "Properties": {
        "VpcId": {
          "RelationshipType": "VpcId"
        },
        "MapPublicIpOnLaunch": {
          "RelationshipType": "String"
        }
      }
    }
  }
}
```

kddejong:
*Issue #, if available:*
Fixes #34

*Description of changes:*
- Appends to the CloudFormation Spec to include Ref and GetAtt knowledge


kddejong:
Updated to:
- handle many different levels of conditions (Fn::If)
- GetAtts to Custom Resources are now skipped
- Better testing for when looking for List or values inside a List

Still to do:
Handle GetAtt to nested CloudFormation stacks.  Unknown Outputs from those stacks cannot be validated.

kddejong:
Finished GetAtts to custom resource exceptions.

kddejong:
This is probably ready for a more in depth review.

cmmeyer:
So looking at this and #259 I wonder if we need two separate files? Seems like the only difference is one is for string values and the other is for resource values, unless I missed something?

Would it make sense to consolidate to a single extended spec file, or even patch the existing specs to include value?

kddejong:
I completely agree with this.  It would be nice to combine the spec files into one.  So then the discussion would become what should the format be, what makes sense for both use cases, etc.

kddejong:
Oh, hmm.  Patching could work possibly too.  Would we want to patch that much?

kddejong:
This will get closed with #473

kddejong:
Replaced by #525

PR #248, opened 2018-07-27
load all instances of all CloudFormationLintRule classes in a module

### LINKED ISSUE #247: Adding new CloudFormationLintRule's requires the filename to be identical to the Class name
0.0.9

*Description of issue.*

When creating your own rules, you have to add  CloudFormationLintRules, but the loader requires
the python filename to be identical to the Classname.

It would be nice if:
- we can create files with names matching the python convention.
- we could add multiple lint rules in a single module.

mvanholsteijn:
*Issue #, if available:*
closes #247 

*Description of changes:*
When loading the modules, we inspect all objects and add instances of subclasses of CloudFormationLintRule.

### DIFF
--- a/src/cfnlint/helpers.py
+++ b/src/cfnlint/helpers.py
@@ -22,6 +22,7 @@ import imp
 import logging
 import re
 import pkg_resources
+import inspect
 
 
 LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


fatbasstard:
Looking pretty awesome! 🎉 

Is there a way to add a Test for this, so we can "proof" it works (and keeps working in the future)?

PR #259, opened 2018-08-07
Add Spec file and test for properties Allowed Values

fatbasstard:
*Issue https://github.com/awslabs/cfn-python-lint/issues/50*

Introduce new rule `E3030` to check allowed values as mentioned in the documentation. Properties sometimes have a list of "allowed values" (e.g. https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-vpc.html#cfn-aws-ec2-vpc-instancetenancy).

These values are not in the Spec files, and probably will never be. This PR will add a new Additional Spec file (`AllowedValues.json`) which contains these allowed values in the same format as the Spec files itself, and the other Additional Spec files. In this way all Allowed Values are maintained in a single place which makes more transparent, easier to maintain and remove this logic from current (existing) rules.

This is a first version (MVP) that introduces the new Rule and the definition file. The definition file contains a first set of known allowed values. Also "moved" some of the existing rule logic already (including altering/moving the test logic).

It's an MVP rule at the moment that is open for extension/improvement when required. By adding it already we can move move allowed values in the future from rules and more values can be added to the definition.

### DIFF
--- a/src/cfnlint/data/AdditionalSpecs/AllowedValues.json
+++ b/src/cfnlint/data/AdditionalSpecs/AllowedValues.json
@@ -46,6 +46,19 @@
         }
       }
     },
+    "AWS::EC2::Volume": {
+      "Properties": {
+        "VolumeType": {
+          "AllowedValues": [
+            "gp2",
+            "io1",
+            "sc1",
+            "st1",
+            "standard"
+          ]
+        }
+      }
+    },
     "AWS::EC2::VPC": {
       "Properties": {
         "InstanceTenancy": {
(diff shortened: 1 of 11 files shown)


src/cfnlint/rules/resources/properties/AllowedValues.py
```diff
@@ -0,0 +1,149 @@
...
+                                ', '.join(map(str, allowed_values))
+                            )))
+
+            # For all specified properties, check all nested properties
+            for prop in value:
+                proptree = path[:]
+                proptree.append(prop)
+                if prop in resourcespec:
+                    if 'Type' in resourcespec[prop]:
+                        if resourcespec[prop]['Type'] == 'List':
```
kddejong: Would this ever actually get hit?  I'm not sure a Type List is in this spec.

fatbasstard: Need to add valid support like Ref's etc also in there anyway. I'll continue on changing this POC

fatbasstard: More like: If it's a ref to a parameter, check the default values from the parameter

kddejong: Got it.  Checking the default value would be great.

The other one but probably a different rule is if it had AllowedValues that those individual allowed values are within the spec.

fatbasstard: But that might be bit "much" for now (mvp)

kddejong: Yep agreed. Different rule diffferent day

src/cfnlint/data/AdditionalSpecs/AllowedValues.json
```diff
@@ -0,0 +1,134 @@
...
+          "AllowedValues": [
+            "provisioned",
+            "serverless"
+          ]
+        }
+      }
+    }
+  },
+  "PropertyTypes": {
+    "AWS::DMS::Endpoint": {
```
kddejong: Shouldn't this have a property with a `.` between?

fatbasstard: Yeah, wrong place... Will fix

cmmeyer:
So looking at this and #234 I wonder if we need two separate files? Seems like the only difference is one is for string values and the other is for resource values, unless I missed something?

Would it make sense to consolidate to a single extended spec file, or even patch the existing specs to include value?

src/cfnlint/rules/resources/properties/AllowedValues.py
```diff
@@ -0,0 +1,162 @@
...
+            for prop in resourcespec:
+
+                if prop in allowed_resourcespec:
+                    allowed_values = allowed_resourcespec[prop].get('AllowedValues')
+
+                    # Check if there's a value at all. Required is handled by
+                    # the required properties rule (E3003)
+                    if prop in value:
+
+                        if not isinstance(value[prop], dict):
```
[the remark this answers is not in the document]
fatbasstard: Funny detail BTW. This example breaks the existing Runtime check (`E0002`):

```
E0002 Unknown exception while processing rule E2531: sequence item 5: expected str instance, int found
test/fixtures/templates/bad/properties_allowedvalues.yaml:1:
```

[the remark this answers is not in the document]
kddejong: Yea that is covered by pull request #457

fatbasstard:
Replaced by a new PR that uses the new Spec Patch mechanism

Issue #264, opened 2018-08-07
method to load yaml from a string


jaymccon:
*cfn-lint version: cfn-lint 0.4.2*

*Description of issue.*

When consuming cfn-lint as a module there is currently only an interface to load yaml from a file(`cfnlint.core.decode.cfn_yaml.load`). In our project([taskcat](https://github.com/aws-quickstart/taskcat/)) we have a need to load templates from strings as well. 

While we could just do something like below, think it would be useful to others to have a simpler interface for yaml strings:

```python
from cfnlint.decode.cfn_yaml import MarkedLoader, multi_constructor
loader.add_multi_constructor('!', multi_constructor)
loader = MarkedLoader(my_yaml_string, None)
template = loader.get_single_data()
```

avattathil:
Agree, something like cfnlint.core.decode.cfn_yaml.loads would be a nice add

PR #265, opened 2018-08-07
add method to load yaml from string, fixes ##264

jaymccon:
*Issue #264*

*Description of changes:*

Adds a `loads` function that parses yaml from a string. Refactored the `load` function to consume this new function so that there's no code duplication.


cmmeyer:
Cool new feature! Should open up some interesting new integration routes. Thanks!

PR #273, opened 2018-08-08
don't trigger when value of protocol is an intrinsic function. Fixes #272

### LINKED ISSUE #272: [E2503] fails incorrectly when intrinsic function used in Protocol value
*cfn-lint version: 0.4.2*

*Description of issue.*

This is valid, and conforms to the spec, but rule throws an error:

```yaml
Parameters:
  TestParam:
    Type: String
    Default: TCP
Conditions:
  TestCond: !Equals ['a', 'a']
Resources:
  OpenShiftMasterELB:
    Type: AWS::ElasticLoadBalancing::LoadBalancer
    Properties:
      Subnets:
        - subnet-1234abcd
      SecurityGroups:
        - sg-1234abcd
      Listeners:
        # Fails on Protocol
        - InstancePort: '1'
          InstanceProtocol: !Ref TestParam
          LoadBalancerPort: '1'
          Protocol: !Ref TestParam
        # Also fails on Protocol
        - InstancePort: '2'
          InstanceProtocol: !If [TestCond, TCP, SSL]
          LoadBalancerPort: '2'
          Protocol: !If [TestCond, TCP, SSL]
        # Works
        - InstancePort: '3'
          InstanceProtocol: !If [TestCond, TCP, SSL]
          LoadBalancerPort: '3'
          Protocol: TCP
```

jaymccon:
*Issue #, if available: #272*

*Description of changes:*
protocol was being treated as if it's always a string with a list of valid values, when Ref or Fn::If are used to populate the value this rule was generating a false positive.

### DIFF
--- a/src/cfnlint/rules/resources/elb/Elb.py
+++ b/src/cfnlint/rules/resources/elb/Elb.py
@@ -30,13 +30,21 @@ HTTPS has certificate HTTP has no certificate'
     def match(self, cfn):
         """Check ELB Resource Parameters"""
 
+        def is_intrinsic(input_obj):
+            """Checks if a given input looks like an intrinsic function"""
+
+            if isinstance(input_obj, dict) and len(input_obj) == 1:
+                if list(input_obj.keys())[0] == 'Ref' or list(input_obj.keys())[0].startswith('Fn::'):
+                    return True
+            return False
+
         matches = list()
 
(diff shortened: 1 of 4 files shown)


jaymccon:
err, rebased and ended up somehow including an extraneous commit (4414483), will sort it out in the morning.

PR #279, opened 2018-08-10
Set AWS region to control the region in SAM

fatbasstard:
*Issue https://github.com/awslabs/cfn-python-lint/issues/253*

There is a "hidden" depency on AWS being configured locally when SAM Transofmations. SAM extracts the region when generating AWS Service Role ARNS.

Apart from this dependancy, this is also means SAM works on a different region than the linter, which could result in unexpected results.

This PR sets the `AWS_DEFAULT_REGION` Environment Variable (also as [suggested by SAM](https://github.com/awslabs/serverless-application-model/blob/master/samtranslator/translator/arn_generator.py#L39) as the way to control this, since this is the first value being checked by Boto).

So this PR makes sure the region of SAM transformations and the linter are the same, and removed the dependency. 

### DIFF
--- a/src/cfnlint/transform.py
+++ b/src/cfnlint/transform.py
@@ -14,6 +14,7 @@
   OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
   SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 """
+import os
 import six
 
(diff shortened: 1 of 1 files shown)


fatbasstard:
Not easy to test unfortunately. It's deep in the SAM code, tested it by hacking the local code in `venv/`. Could create a test if GovCloud or China would be supported, only then something changes... But that would be a bit like testing another repo.

All current tests still work :wink:

PR #293, opened 2018-08-17
Fix Output check when ImportValue is part of another intrinsic function(s)

fatbasstard:
*Issue https://github.com/awslabs/cfn-python-lint/issues/291*

ImportValue can actually be used by another instrinsic function, e.g. a `!Sub` which is a totally valid scenario. Fixed this and added the sample template from the issue as a "good" testing scenario


fatbasstard:
Should be pretty much impossible to have less items due to the retrieval and filtering up front, but that's all "implicit". I can add an explicit check  🤔

Issue #297, opened 2018-08-17
Dynamic references to secure strings should only be allowed in supported fields


cmmeyer:
*cfn-lint version: (`cfn-lint --version`)*
0.5.1

*Description of issue.*
The new dynamic references to SSM Parameter Store only support secure strings for certain fields as documented here:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/dynamic-references.html

Field list:

Resource | Property Type | Properties
-- | -- | --
AWS::DirectoryService::MicrosoftAD |   | Password
AWS::DirectoryService::SimpleAD |   | Password
AWS::ElastiCache::ReplicationGroup |   | AuthToken
AWS::IAM::User | LoginProfile | Password
AWS::KinesisFirehose::DeliveryStream | RedshiftDestinationConfiguration | Password
AWS::OpsWorks::App | AppSource | Password
AWS::OpsWorks::Stack | CustomCookbooksSource | Password
AWS::OpsWorks::Stack | RdsDbInstances | DbPassword
AWS::RDS::DBCluster |   | MasterUserPassword
AWS::RDS::DBInstance |   | MasterUserPassword
AWS::Redshift::Cluster |   | MasterUserPassword

Please provide as much information as possible:
* Template linting issues:
  * Please provide a CloudFormation sample that generated the issue.
  * If present, please add links to the (official) documentation for clarification.
* Feature request:
  * Please provide argumentation about the missing feature. Context is key!

cmmeyer:
Resolved via PR #299

PR #299, opened 2018-08-20
Add a new rule to check the location of dynamic references

kddejong:
*Issue #, if available:*
#297 

*Description of changes:*
- Add rule E1027 to check if dynamic references are being used with the correct properties.


cmmeyer:
Hi! This is awesome, but the limitation is only for ssm-secure dyn refs.

Issue #150, opened 2018-06-21
Support for AWS StepFunctions 


Nr18:
Would be great if the linter also validates the `DefinitionString` of the [`AWS::StepFunctions::StateMachine`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-stepfunctions-statemachine.html) resource.

The value *must* be in JSON format and the structure is the [Amazon States Language](http://docs.aws.amazon.com/step-functions/latest/dg/concepts-awl.html).

On the *AWS Developers* slack in the [#serverless](https://awsdevelopers.slack.com/archives/C6LDW0BC3/p1529528449000166) channel we discussed it briefly "Simon" stated that he has one written in python that might/can be used for this purpose.

Example(s) of other state linters:

- https://github.com/awslabs/statelint

cmmeyer:
I notice we merged a rule for StateMachines Definitions back in August. Ok to close this Issue?

Nr18:
Makes sense to.

kddejong:
Its worth noting this could be better but its a start.  We can adjust and make improvements as the need arises.

PR #312, opened 2018-08-24
Add rule E2532 to check state machine syntax

kddejong:
*Issue #, if available:*
A start to #150

*Description of changes:*
- Add rule E2532 to start checking state machine syntax.

### DIFF
--- a/src/cfnlint/rules/resources/stepfunctions/StateMachine.py
+++ b/src/cfnlint/rules/resources/stepfunctions/StateMachine.py
@@ -1,0 +1,155 @@
+"""
+  Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+  Permission is hereby granted, free of charge, to any person obtaining a copy of this
+  software and associated documentation files (the "Software"), to deal in the Software
+  without restriction, including without limitation the rights to use, copy, modify,
+  merge, publish, distribute, sublicense, and/or sell copies of the Software, and to
+  permit persons to whom the Software is furnished to do so.
+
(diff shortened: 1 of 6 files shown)


src/cfnlint/rules/resources/stepfunctions/StateMachine.py
```diff
@@ -0,0 +1,155 @@
...
+            matches.append(RuleMatch(path, message))
+            return matches
+
+        matches.extend(self._check_definition_json(def_json, path))
+        return matches
+
+    def check_sub(self, value, path):
+        """Check Sub Object"""
+        matches = list()
+        if isinstance(value, list):
```
cmmeyer: Trying to understand under what circumstances this might be a list and why we only need the first value if it is?

kddejong: My solutions to this would either be:
- not check the state definitions when Fn::Sub is used
- create a way to determine if someone is doing the more complex method and not check it. 
  - JSON blobs addended to the state definition (this may be harder to detect as it may be based on the parse-ability)  This should be really uncommon and in this scenario the rule should be ignored.
  - Using Sub variables for JSON key names.  Easier to detect as it should still be parseable and we would check the key values for things that looked like variables.  

I'm leaning the later as I would expect Subs would be very common for this.

cmmeyer: After talking, it sounds like there are edge cases but this should solve for 80/20 -- and the 20 can just disable the rule and file an issue.  Rest of the rule looks good.

PR #315, opened 2018-08-31
update rule E3002 to handle templated code

### LINKED ISSUE #239: Linter doesn't handle templated code correctly
*cfn-lint version: cfn-lint 0.4.1

*Description of issue.*

The linter doesn't handle templated code [see reference](https://docs.aws.amazon.com/cli/latest/reference/cloudformation/package.html) where `Code: ./sample.zip` is accepted as a valid template but errors in the linter.

* Template linting issues:

image

```
  SampleLambda:
    Type: "AWS::Lambda::Function"
    Properties:
      Role:
        Fn::ImportValue:
          !Sub "${BuildSystemIAMStackName}-SampleLambdaRoleArn"

      Runtime: python2.7
      Handler: lambda_sample.handler
      Code: ./sample.zip
      Timeout: 300
```

Above

What is the guidance on template files being interpreted as CFN yaml files? Should this be picked up by the linter?

kddejong:
*Issue #, if available:*
- Fixes #239 
*Description of changes:*
- Update rule E3002 to handle templated code where a zip file can be passed in.  Only needs to check where there was supposed to be an object and there is a string now.  Some of templated properties are already strings so they don't need exceptions.

### DIFF
--- a/src/cfnlint/rules/functions/Cidr.py
+++ b/src/cfnlint/rules/functions/Cidr.py
@@ -91,6 +91,7 @@ class Cidr(CloudFormationLintRule):
         supported_functions = [
             'Fn::Select',
             'Ref',
+            'Fn::GetAtt',
             'Fn::ImportValue'
         ]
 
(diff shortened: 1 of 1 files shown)


cmmeyer:
So if I'm reading this right, this is for templates that will be used with CloudFormation "package" and "deploy" commands, where these values will be expanded to the final object (Code etc.).

You cannot use this template directly for a CREATE action within CloudFormation.  For this reason, I wonder if we should still throw a warning for these to make sure the end user is aware the value will need expansion?

kddejong:
Yea, I can agree with that.  The folks that are knowingly using `package` and `deploy` could disable the warning as they know what they are getting into.

I would want to make sure we are consistent with that approach.  Like with `sam`.  I think we may be covered there as they have to enter the Transform details which means they have some knowledge of how to deploy it.

cmmeyer:
That's a good point about SAM.  I'd say the rule of thumb would be "anything you can't just run a Create Stack against"...

PR #319, opened 2018-09-02
Add rule E3022 to check resource policy syntax

### LINKED ISSUE #313: Add IAM policy checks
*cfn-lint version: (`cfn-lint --version`) 0.5.0*

*Description of issue.*

Used 'principle' instead of 'principal' in an SNS topic policy, spent 30 minutes finding the error, because the error message from CloudFormation is just an echo of the service error message, which simply says 'null':

> 'AWS::SNS::TopicPolicy SomeTopicPolicy CREATE_FAILED: Invalid parameter: Policy Error: null (Service: AmazonSNS; Status Code: 400; Error Code: InvalidParameter; Request ID: aaaaaaaa-aaaa-aaaa-aaaa-aaaaaaaaaaaa)'

Example template:

```yaml

Resources:

  SomeTopicPolicy:
    Type: AWS::SNS::TopicPolicy
    Properties:
      Topics:
      - some-topic-name
      PolicyDocument:
        Id: some-policy
        Version: 2012-10-17
        Statement:
        - Sid: some-sid
          Effect: Allow
          Principle:
            Service: events.amazonaws.com
          Action:
          - sns:Publish
          Resource: '*'
```

Probably more things can be checked in the structure of IAM/S3/SNS/... policies too, cloudformation itself seems to just take opaque blobs and pass them to the underlying system. Ofc IAM/SNS should have better error messages too.

kddejong:
*Issue #, if available:*
Fixes #313
*Description of changes:*
- Add rule E3022 to check resource policies for syntax correctness

### DIFF
--- a/src/cfnlint/rules/resources/iam/Policy.py
+++ b/src/cfnlint/rules/resources/iam/Policy.py
@@ -14,6 +14,7 @@
   OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
   SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 """
+from datetime import date
 from cfnlint import CloudFormationLintRule
 from cfnlint import RuleMatch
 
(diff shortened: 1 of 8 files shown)


src/cfnlint/rules/resources/iam/ResourcePolicy.py
```diff
@@ -0,0 +1,136 @@
...
+    def __init__(self):
+        """Init"""
+        self.resources_and_keys = {
+            'AWS::SNS::TopicPolicy': 'PolicyDocument',
+            'AWS::S3::BucketPolicy': 'PolicyDocument',
+            'AWS::KMS::Key': 'KeyPolicy',
+            'AWS::SQS::QueuePolicy': 'PolicyDocument',
+            'AWS::ECR::Repository': 'RepositoryPolicyText',
+            'AWS::Elasticsearch::Domain': 'AccessPolicies',
+        }
```
cmmeyer: You should include IoT and VPCEndpoints as well:
              'AWS::EC2::VPCEndpoint': 'PolicyDocument',
              'AWS::IoT::Policy': 'PolicyDocument'

src/cfnlint/rules/resources/iam/ResourcePolicy.py
```diff
@@ -0,0 +1,136 @@
...
+        for resource_type in self.resources_and_keys:
+            self.resource_property_types.append(resource_type)
+
+    def check_policy_document(self, value, path, cfn):
+        """Check policy document"""
+        matches = list()
+
+        valid_keys = [
+            'Version',
```
cmmeyer: We should enforce the most recent policy version ('2012-10-17') in a constant, per: https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_version.html

kddejong:
I think this rule covers the Identity Based Policies.  https://github.com/awslabs/cfn-python-lint/blob/master/src/cfnlint/rules/resources/iam/Policy.py

cmmeyer:
Comparing the two rules in Meld, it seems like the biggest difference is that IAM policies cannot contain a principal:

```
        if 'Principal' in statement:
            message = 'IAM Policy statement shouldn\'t have Principal'
            matches.append(
                RuleMatch(branch[:] + ['Principal'], message))
        if 'NotPrincipal' in statement:
            message = 'IAM Policy statement shouldn\'t have NotPrincipal'
            matches.append(
                RuleMatch(branch[:] + ['NotPrincipal'], message))
```
It would be great to consolidate these into a single rule with resource lists that signify whether a the policy is a resource policy or an identity policy.

src/cfnlint/rules/resources/iam/Policy.py
```diff
@@ -97,60 +127,59 @@ def _check_policy_statement(self, branch, statement):
...
-        """Check IAM Policies Properties"""
+        is_identity_policy = True
+        if resourcetype in self.resources_and_keys:
+            is_identity_policy = False
 
-        matches = list()
+        key = None
+        if resourcetype in self.resources_and_keys:
+            key = self.resources_and_keys.get(resourcetype)
+        else:
```
cmmeyer: This seems dangerous. Should check against both type (resource_and_keys and idp_and_keys) and throw an exception if it's neither.

PR #331, opened 2018-09-07
Support parsing multiple files

kddejong:
*Issue #, if available:*

*Description of changes:*
- Supports linting multiple files

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,19 @@ There are IDE plugins available to get direct linter feedback from you favorite 
 * [Visual Studio Code](https://marketplace.visualstudio.com/items?itemName=kddejong.vscode-cfn-lint)
 * [IntelliJ IDEA](https://plugins.jetbrains.com/plugin/10973-cfn-lint/update/48247)
 
+## Basic Usage
+- `cfn-lint template.yaml`
+- `cfn-lint -t template.yaml`
+
+##### Lint multiple files
+- `cfn-lint template1.yaml template2.yaml`
+- `cfn-lint -t template1.yaml template2.yaml`
+- `cfn-lint path/*.yaml`
+
+##### Specifying the template with other parameters
+- `cfn-lint -r us-east-1 ap-south-1 -- template.yaml`
+- `cfn-lint -r us-east-1 ap-south-1 -t template.yaml`
+
 ## Configuration
 
 ### Command Line
@@ -77,15 +90,6 @@ Metadata:
 
 ### Precedence
 cfn-lint applies the configuration from the CloudFormation Metadata first and then overrides those values with anything specified in the CLI.
-
-## Examples
-### Basic usage
-`cfn-lint --template template.yaml`
-
-### Test a template based on multiple regions
-`cfn-lint --regions us-east-1 ap-south-1 --template template.yaml`
-
-> E3001 Invalid Type AWS::Batch::ComputeEnvironment for resource testBatch in ap-south-1
 
 ### Getting Started Guides
 There are [getting started guides](/docs/getting_started) available in the documentation section to help with integrating `cfn-lint` or creating rules.
--- a/src/cfnlint/core.py
+++ b/src/cfnlint/core.py
@@ -37,17 +37,13 @@ class ArgumentParser(argparse.ArgumentParser):
         self.exit(32, '%s: error: %s\n' % (self.prog, message))
 
 
-def run_cli(filename, template, rules, regions, override_spec, formatter):
+def run_cli(filename, template, rules, regions, override_spec):
     """Process args and run"""
 
     if override_spec:
         cfnlint.helpers.override_specs(override_spec)
 
-    matches = run_checks(filename, template, rules, regions)
-
-    formatter.print_matches(matches)
-
-    return get_exit_code(matches)
+    return run_checks(filename, template, rules, regions)
 
 
 def get_exit_code(matches):
@@ -85,12 +81,21 @@ def comma_separated_arg(string):
     return string.split(',')
 
 
+def space_separated_arg(string):
+    """ Split a comma separated string """
+    return string.split(' ')
+
+
 class ExtendAction(argparse.Action):
     """Support argument types that are lists and can be specified multiple times."""
     def __call__(self, parser, namespace, values, option_string=None):
-        items = getattr(namespace, self.dest, [])
+        items = getattr(namespace, self.dest)
+        items = [] if items is None else items
         for value in values:
-            items.extend(value)
+            if isinstance(value, list):
+                items.extend(value)
+            else:
+                items.append(value)
         setattr(namespace, self.dest, items)
 
 
@@ -104,10 +109,10 @@ def create_parser():
 
     # Alllow the template to be passes as an optional or a positional argument
     standard.add_argument(
-        'template', nargs='?', help='The CloudFormation template to be linted')
+        'templates', metavar='TEMPLATE', nargs='*', help='The CloudFormation template to be linted')
     standard.add_argument(
-        '-t', '--template', metavar='TEMPLATE', dest='template_alt', help='The CloudFormation template to be linted')
-
+        '-t', '--template', metavar='TEMPLATE', dest='template_alt',
+        help='The CloudFormation template to be linted', nargs='+', default=[], action='extend')
     standard.add_argument(
         '-b', '--ignore-bad-template', help='Ignore failures with Bad template',
         action='store_true'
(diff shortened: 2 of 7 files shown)


kddejong:
A few things may need to get done before merging this one.

- [ ] Make sure there is no adverse affect to the people using cfn-lint as a module
- [ ] Double check that the IDE plugins still work

cmmeyer:
I noticed you've removed the WIP tag from this PR -- is it ready for review and merge?

kddejong:
yea I think this is about ready.  In the past there has a been a few argparse things that trip me up.  I'm hoping I can get a few people to put their eyes on it to make sure I didn't miss anything.

cmmeyer:
@alexjurkiewicz and @adamchainz  -- would either of you be willing to do a quick review/test of this PR to see if it helps with your multi-file use cases?

kddejong:
It should work for both... I'll check in on this.

adamchainz:
@cmmeyer comma separated lists shouldn't really pass since commas are allowed in filenames, and there's already a way of passing multiple with space separation

adamchainz:
Takes my lint of 73 templates from 53 seconds to 4 seconds!

But one thing that might be worth documenting is using the `--` separator, since e.g. `--regions` consumes multiple arguments, I have to do `cfn-lint --regions us-east-1 -- cloudformation/*.yml` to make it work. The error message is unclear that there is no template passed without the `--`.

cmmeyer:
It would be nice to print each filename in debug mode -- I got an exception when I pointed the linter at a random directory and it was hard to figure out which file caused the exception. 

(For what it's worth, the exception was when we hit a binary file(.zip).  I'll create a separate issue for that)

```
  File "/usr/local/bin/cfn-lint", line 11, in <module>
    load_entry_point('cfn-lint==0.7.2', 'console_scripts', 'cfn-lint')()
  File "/usr/local/lib/python3.6/site-packages/cfn_lint-0.7.2-py3.6.egg/cfnlint/__main__.py", line 30, in main
    (template, rules, template_matches) = cfnlint.core.get_template_rules(filename, args)
  File "/usr/local/lib/python3.6/site-packages/cfn_lint-0.7.2-py3.6.egg/cfnlint/core.py", line 245, in get_template_rules
    (template, matches) = cfnlint.decode.decode(filename, args.ignore_bad_template)
  File "/usr/local/lib/python3.6/site-packages/cfn_lint-0.7.2-py3.6.egg/cfnlint/decode/__init__.py", line 40, in decode
    template = cfnlint.decode.cfn_yaml.load(filename)
  File "/usr/local/lib/python3.6/site-packages/cfn_lint-0.7.2-py3.6.egg/cfnlint/decode/cfn_yaml.py", line 206, in load
    return loads(fp.read(), filename)
  File "/usr/local/lib/python3.6/site-packages/cfn_lint-0.7.2-py3.6.egg/cfnlint/decode/cfn_yaml.py", line 191, in loads
    loader = MarkedLoader(yaml_string, fname)
  File "/usr/local/lib/python3.6/site-packages/cfn_lint-0.7.2-py3.6.egg/cfnlint/decode/cfn_yaml.py", line 140, in __init__
    Reader.__init__(self, stream)
  File "/usr/local/lib/python3.6/site-packages/yaml/reader.py", line 74, in __init__
    self.check_printable(stream)
  File "/usr/local/lib/python3.6/site-packages/yaml/reader.py", line 144, in check_printable
    'unicode', "special characters are not allowed")
yaml.reader.ReaderError: unacceptable character #x0000: special characters are not allowed
  in "<unicode string>", position 0
```

kddejong:
Thanks for the feedback. Added logging for the filename information and did some extra items around the comma/space parsing.  Also added some more documentation.

src/cfnlint/core.py
```diff
@@ -104,10 +104,11 @@ def create_parser():
 
     # Alllow the template to be passes as an optional or a positional argument
     standard.add_argument(
-        'template', nargs='?', help='The CloudFormation template to be linted')
+        'template', nargs='*', help='The CloudFormation template to be linted')
     standard.add_argument(
-        '-t', '--template', metavar='TEMPLATE', dest='template_alt', help='The CloudFormation template to be linted')
-
+        '-t', '--template', metavar='TEMPLATE', dest='template_alt',
```
alexjurkiewicz: `nargs='+'` and `type=space_separated_arg` seems redundant. The args will be split already when passed in and this might break paths with spaces? (I'm not sure though)

src/cfnlint/decode/__init__.py
```diff
@@ -91,6 +94,12 @@ def create_match_yaml_parser_error(parser_error, filename):
         cfnlint.ParseError(), message=msg)
 
 
+def create_match_file_error(filename, msg):
+    """Create a Match for a parser error"""
+    return cfnlint.Match(
+        1, 1, 1, 2, filename, cfnlint.ParseError(), message=msg)
```
alexjurkiewicz: this function call is a bit cryptic. Should it used named parameters?

src/cfnlint/core.py
adamchainz: yeah don't break paths with spaces

kddejong: They were coming in as an array per character when I removed the `type=space_separated_arg`

And this results in only one `-t` working which doesn't make sense to me either.
```
standard.add_argument(
        '-t', '--template', metavar='TEMPLATE', dest='template_alt',
        help='The CloudFormation template to be linted', nargs='+')
```

This didn't make a lot of sense to me either but I'm probably missing something.

src/cfnlint/decode/__init__.py
kddejong: Updated.  Thanks!

src/cfnlint/core.py
alexjurkiewicz: Here is an example of allowing multiple arguments:

```py
import argparse
parser = argparse.ArgumentParser()

parser.add_argument('templates', nargs='*')
parser.add_argument('-t', dest='template_alt', nargs='+')

args = parser.parse_args()

print(repr(args))
```

And it works:

```
$ python test.py -t "a b" c d "e f" -- "1 2" 3 4 "5 6"
Namespace(template_alt=['a b', 'c', 'd', 'e f'], templates=['1 2', '3', '4', '5 6'])
```

I think the issue is the unnecessary `default=[]` or `action='extend'`?

src/cfnlint/core.py
```diff
@@ -104,10 +104,11 @@ def create_parser():
 
     # Alllow the template to be passes as an optional or a positional argument
     standard.add_argument(
-        'template', nargs='?', help='The CloudFormation template to be linted')
+        'template', nargs='*', help='The CloudFormation template to be linted')
```
alexjurkiewicz: This is a real nitpick, but should probably call this variable 'templates' (with a metavar of TEMPLATE).

README.md
```diff
@@ -80,10 +80,19 @@ cfn-lint applies the configuration from the CloudFormation Metadata first and th
 
 ## Examples
 ### Basic usage
-`cfn-lint --template template.yaml`
+- `cfn-lint template.yaml`
+- `cfn-lint --template template.yaml`
+- `cfn-lint -t template.yaml`
```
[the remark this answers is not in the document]
alexjurkiewicz: To expand on this, for quick start people probably just need to see `cfn-lint template.yaml`. Other forms can be left for the `--help` output.

src/cfnlint/core.py
kddejong: We allowed the extend for multiple `-t`.  I did find an odd issue with the way the extend function worked that I fixed.  That should resolve the strings being split into multiple element issue.

README.md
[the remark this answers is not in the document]
kddejong: Removed `--template` and moved up the usage section in the README so it isn't so far down.  I will look into the adding information into the help.  I think there was a request to make sure we showed the `--` in the README so we didn't cause confusion when using the default parameters with other parameters.  The `-t` examples can be moved though.

alexjurkiewicz:
Top stuff guys!!

adamchainz:
🐇

Saichovsky:
`pip install git+git://github.com/awslabs/cfn-python-lint.git` works, but when I run `pre-commit-validate-config .pre-commit-config.yaml`, it fails with the following Traceback:

```
  File "/usr/bin/pre-commit-validate-config", line 11, in <module>
    sys.exit(validate_config_main())
  File "/usr/lib/python2.7/site-packages/pre_commit/clientlib.py", line 190, in validate_config_main
    load_config(filename)
  File "/usr/lib/python2.7/site-packages/cfgv.py", line 347, in load_from_filename
    contents = f.read()
  File "/usr/lib64/python2.7/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 386: ordinal not in range(128)
```

adamchainz:
@Saichovsky that doesn't seem related to this at all, none of the stack trace frames are in cfn-python-lint. From the stacktrace I guess you have non-unicode characters in your configuration file.

Saichovsky:
@adamchainz Indeed. Reported on the wrong repo. Apologies.

PR #335, opened 2018-09-10
Use list literal instead of list()

adamchainz:
dict, tuple, and list literals are actually single instructions in the CPython virtual machine, so it's always faster and better form to use them over the type constructors.


PatMyron:
Seems like we've reintroduced `list()` since

@adamchainz do you know if there's a way to enforce this going forward?

adamchainz:
Rule C408 in my flake8 plugin [flake8-comprehensions](https://pypi.org/project/flake8-comprehensions/) covers this.

PR #336, opened 2018-09-10
Stop using 'discover'

adamchainz:
I discovered that [discover](https://pypi.org/project/discover/) is a backport of the Python 2.7 unittest discover runner, since only Python 2.7+ is supported (or at least tested, I couldn't find documentation on that), use the standard library version instead.


kddejong:
Looks like a lot more comments in the Travis build.  I want to dig into this first.
https://travis-ci.com/awslabs/cfn-python-lint/jobs/144978155

adamchainz:
Oh yeah `unittest` doesn't do stdout/stderr capture, looks like `discover` does. I'd suggest moving to `pytest` for features like that and more. I use on all my projects - it can run unittest suites with no alterations.

PR #352, opened 2018-09-15
Support get value that handles conditions on the object itself

kddejong:
*Issue #, if available:*

*Description of changes:*
- Moving away from cfnlint.Template to have functions that are acting on dicts/lists inside the objects that make up the CloudFormation template
- Create new functions that are part of the cfnlint custom dict/list types to enumerate and get values that are condition safe

This should allow us to move away from passing cfn around and just functions that are part of the overrode dict and list.  While making this change I noticed when parsing yaml !Functions were not getting translated into the custom dict objects.


src/cfnlint/decode/cfn_json.py
```diff
@@ -265,10 +260,11 @@ def CfnJSONObject(s_and_end, strict, scan_once, object_hook, object_pairs_hook,
             raise JSONDecodeError('Duplicate found {}'.format(err), s, begin, key)
         except NullError as err:
             raise JSONDecodeError('Null Error {}'.format(err), s, begin, key)
-        return result, end
+
```
cmmeyer: You've removed the return from within this object_pairs_hook block -- is that intended? I noticed the return remains in an early block around line 192.

src/cfnlint/decode/node.py
```diff
@@ -0,0 +1,179 @@
...
+            self.start_mark = start_mark
+            self.end_mark = end_mark
+
+        # pylint: disable=bad-classmethod-argument, unused-argument
+        def __new__(self, x, start_mark, end_mark):
+            if sys.version_info >= (3, 0):
+                return cls.__new__(self, x)
+
+            if isinstance(x, six.string_types):
+                return cls.__new__(self, x.encode('ascii', 'ignore'))
```
cmmeyer: ASCII/UTF-8

src/cfnlint/decode/cfn_json.py
kddejong: it was kind of intentional but not done well.  I moved all the conversion to the dict_node class to the object_pairs_hook and then added this back.  It should be now that everything is getting converted to our class of dict_node and returned using that return.

src/cfnlint/decode/node.py
kddejong: #124 should provide some insights.

Also this part is a direct copy from what is there right now and the deleted file str_node.py

Issue #364, opened 2018-09-25
E2521 Parameter MasterUserPassword should exist with MasterUsername for Resources/DBCluster/Properties


cmmeyer:
*cfn-lint version: (`cfn-lint --version`)*
0.7.3

*Description of issue.*
E2521 is properly detecting inclusions for properties 

Please provide as much information as possible:
* Template linting issues:
```
---
AWSTemplateFormatVersion: '2010-09-09'
Description: MU database in a specific environment
Parameters:
  DatabaseMasterUsername:
    Type: String
    NoEcho: true
    Description: Username of database
  DatabaseName:
    Type: String
    Description: Name of database
  DatabaseMasterPassword:
    Type: String
    NoEcho: true
    Description: Password of database

Resources:
  DBCluster:
    Type: AWS::RDS::DBCluster
    Properties:
      MasterUsername: !Ref DatabaseMasterUsername
      MasterUserPassword: !Ref DatabaseMasterPassword
      DatabaseName: !Ref DatabaseName
      Engine: aurora
      EngineMode: serverless
      ScalingConfiguration: 
        AutoPause: true
        MaxCapacity: 8
        MinCapacity: 2
        SecondsUntilAutoPause: 300
      Tags:
      - Key: Name
        Value: !Ref DatabaseName
```

cmmeyer:
Merged PR #366 against this. Closing issue.

Issue #365, opened 2018-09-25
aws::rds::dbcluster SecondsBeforeAutoPause should be SecondsUntilAutoPause


cmmeyer:
*cfn-lint version: (`cfn-lint --version`)*
0.7.3

*Description of issue.*

Please provide as much information as possible:

There is a problem in both the resource specification and documentation when defining an RDS DBCluster ScalingConfiguration.  Both sources list "SecondsBeforeAutoPause" as a valid property but property should be "SecondsUntilAutoPause" per the API documentation.

```
---
AWSTemplateFormatVersion: '2010-09-09'
Description: MU database in a specific environment
Parameters:
  DatabaseMasterUsername:
    Type: String
    NoEcho: true
    Description: Username of database
  DatabaseName:
    Type: String
    Description: Name of database
  DatabaseMasterPassword:
    Type: String
    NoEcho: true
    Description: Password of database

Resources:
  DBCluster:
    Type: AWS::RDS::DBCluster
    Condition: IsClustered
    Properties:
      MasterUsername: !Ref DatabaseMasterUsername
      MasterUserPassword: !Ref DatabaseMasterPassword
      DatabaseName: !Ref DatabaseName
      Engine: aurora
      EngineMode: serverless
      ScalingConfiguration: 
        AutoPause: true
        MaxCapacity: 8
        MinCapacity: 2
#        SecondsBeforeAutoPause: 300
        SecondsUntilAutoPause: 300
      Tags:
      - Key: Name
        Value: !Ref DatabaseName
```

Issue #229, opened 2018-07-18
Create a new rule for RDS


kddejong:
*cfn-lint version: (`cfn-lint --version`)* v0.4.0

*Description of issue.*
Exclusive properties check does not work for the following.
```
"AWS::RDS::DBInstance": {
  "DBSnapshotIdentifier": [
    "CharacterSetName",
    "MasterUsername",
    "MasterUserPassword",
    "StorageEncrypted"
  ]
}
```
This is because `DBSnapshotIdentifier` can be an empty string.  A new check should be created to make sure that these attributes are exclusive if `DBSnapshotIdentifier` is not an empty string.

kddejong:
Should determine how this looks with the new resource provider schemas.

kddejong:
Not fixed in the new provider schemas.

kddejong:
Resolved and this has changed.  This will be covered in v1.

Issue #227, opened 2018-07-18
Ignore 'should not exist' with AWS::NoValue


farski:
This may apply more widely, but as an example: We have a RDS template that can be used for either a new DB or launching from a snapshot. When creating a new DB, a `DBName` property is used, along with `MasterUsername` and `MasterUserPassword`. When using `DBSnapshotIdentifier`, the username and password properties are not allowed. The template uses conditions to determine if there are values for `DBName` _or_ `DBSnapshotIdentifier` or `AWS::NoValue`. The uname/pw also have conditions.

Ideally, it would be great if the linter could follow conditional branches and understand in which situations sets of properties have which values, but in testing this I've noticed that even if the properties are explicitly set to `AWS::NoValue`, it still throws an error. e.g.:

```
DBSnapshotIdentifier: SomeSnapshotIdentifier
MasterUsername: !Ref AWS::NoValue
MasterUserPassword: !Ref AWS::NoValue
```

Maybe a good starting point would be for it to check if an disallowed property is `AWS::NoValue` and treat it as though it doesn't exist if that's the case (which I believe is how CloudFormation handles such properties).

farski:
Also worth noting that even though this is a linter error, and the docs indicate those properties shouldn't be used when using the `DBSnapshotIdentifier` property, CloudFormation itself doesn't actually complain if you include them simultaneously. 

In fact, the docs for `AWS::NoValue` even use RDS as an example and don't account for the removal of the username and password properties.
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/pseudo-parameter-reference.html#cfn-pseudo-param-novalue

kddejong:
Going to remove the exclusive attribute checks for RDS.  It will have to be replaced by another check.

Cause this is also true.  `By specifying this property, you can create a DB instance from the specified DB snapshot. If the DBSnapshotIdentifier property is an empty string or the AWS::RDS::DBInstance declaration has no DBSnapshotIdentifier property, AWS CloudFormation creates a new database. If the property contains a value (other than an empty string), AWS CloudFormation creates a database from the specified snapshot. If a snapshot with the specified name doesn't exist, AWS CloudFormation can't create the database and it rolls back the stack.`

This is the 2nd time this issue has come up.

For AWS NoValue we are going to have a lot more favorable solution with #121 as those properties will be removed when the value is a Ref to AWS::NoValue.  However for this property we are still not be correct because in this case it also supports an empty string.

[...]

PR #366, opened 2018-09-26
Fixed E2521 to check against other properties instead of values

cmmeyer:
*Issue #, if available:*
#364 

*Description of changes:*

E2521 (inclusive properties) was checking required properties against values instead of other properties in resource, causing false positives.

### DIFF
--- a/src/cfnlint/data/AdditionalSpecs/Exclusive.json
+++ b/src/cfnlint/data/AdditionalSpecs/Exclusive.json
@@ -59,12 +59,6 @@
         "MasterUserPassword",
         "StorageEncrypted"
       ],
-      "DBSnapshotIdentifier": [
-        "CharacterSetName",
-        "MasterUsername",
-        "MasterUserPassword",
-        "StorageEncrypted"
-      ],
       "DBClusterIdentifier": [
         "StorageEncrypted"
       ]
--- a/src/cfnlint/data/AdditionalSpecs/Exclusive.json
+++ b/src/cfnlint/data/AdditionalSpecs/Exclusive.json
@@ -59,6 +59,12 @@
         "MasterUserPassword",
         "StorageEncrypted"
       ],
+      "DBSnapshotIdentifier": [
+        "CharacterSetName",
+        "MasterUsername",
+        "MasterUserPassword",
+        "StorageEncrypted"
+      ],
       "DBClusterIdentifier": [
         "StorageEncrypted"
       ]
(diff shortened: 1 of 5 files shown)


src/cfnlint/rules/resources/properties/Inclusive.py
```diff
@@ -46,11 +46,11 @@ def check(self, properties, inclusions, path):
             return matches
 
         for items, p in properties.items_safe(path=path, type_t=(dict)):
-            for k, v in items.items():
+            for k in items.keys():
                 if k in inclusions:
                     for incl_property in inclusions[k]:
-                        if incl_property not in v:
```
kddejong: Wow...  Good catch. Any tests we could add for this?  I'm sure we already had some but I'm surprised the good tests still worked.

src/cfnlint/data/AdditionalSpecs/Exclusive.json
```diff
@@ -59,6 +59,12 @@
         "MasterUserPassword",
         "StorageEncrypted"
       ],
+      "DBSnapshotIdentifier": [
```
kddejong: This used to be in this spec but there was a trick with it.  All the properties can be specified together as long as the DBSnapshotIdentifier is an empty string.

`If the property contains a value (which is not an empty string), AWS CloudFormation creates a database based on the specified snapshot. If no snapshot with the given name exists, AWS CloudFormation can not create the database and will roll back the stack.`

#229 is for creating a separate rule for this
#227 was the original issue

It haven't done a good job testing some of this as I think @farski tried to point out.  Can you specify exclusive attributes together if the value is an empty string or AWS::NoValue?  It could be that the RDS team was the only one to point this out in the documentation or it could be that they are the only ones that support this scenario.  Either way it would be worth a little more digging on our part.

test/fixtures/templates/public/rds-cluster.yaml
```diff
@@ -0,0 +1,35 @@
...
+      MasterUsername: !Ref DatabaseMasterUsername
+      MasterUserPassword: !Ref DatabaseMasterPassword
+      DatabaseName: !Ref DatabaseName
+      Engine: aurora
+      EngineMode: serverless
+      ScalingConfiguration: 
+        AutoPause: true
+        MaxCapacity: 8
+        MinCapacity: 2
+#        SecondsBeforeAutoPause: 300
```
fatbasstard: Can we remove this commented line? Just to keep it clean?

cmmeyer: I removed the whole section, that's there because of a spec bug (#365) which was how I originally stumbled upon this problem.  I also noticed we're not catching the undefined Conditional, I'll remove that line and create another issue for that.

cmmeyer:
Hmm.. I thought I addressed all of @kddejong's change requests, but it still says "Changes Requested" -- not sure what else needs to happen?

PR #375, opened 2018-09-29
Add new rule W3011 to validate DynamoDB has a deletion policy

### LINKED ISSUE #369: Feature Request: Warning if DeletionPolicy is not specified
Sorry for not implementing and submitting pull request but currently have no time for that.

DeletionPolicy is not required in for example DynamoDB Table and defaults to Delete.
That is not a good default value IMHO so always setting DeletionPolicy explicitly would be a nice recommendation.

kddejong:
*Issue #, if available:*
Fix #369
*Description of changes:*
- New rule W3011 to validate DeletionPolicy is specified for AWS::DynamoDB::Table resources as the default policy is to delete them

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,10 +70,14 @@ Optional parameters:
 | -b, --ignore-bad-template | ignore_bad_template | | Ignores bad template errors |
 | -a, --append-rules | append_rules | [RULESDIR [RULESDIR ...]] | Specify one or more rules directories using one or more --append-rules arguments. |
 | -i, --ignore-checks | ignore_checks | [IGNORE_CHECKS [IGNORE_CHECKS ...]] | Only check rules whose ID do not match or prefix these values.  Examples: <br />- A value of `W` will disable all warnings<br />- `W2` disables all Warnings for Parameter rules.<br />- `W2001` will disable rule `W2001` |
+| -c, --include-checks | INCLUDE_CHECKS [INCLUDE_CHECKS ...] | Include rules whose id match these values
 | -d, --debug |  |  | Specify to enable debug logging |
 | -u, --update-specs | | | Update the CloudFormation Specs.  You may need sudo to run this.  You will need internet access when running this command |
 | -o, --override-spec | | filename | Spec-style file containing custom definitions. Can be used to override CloudFormation specifications. More info [here](#customise-specifications) |
 | -v, --version | | | Version of cfn-lint |
+
+### Info Rules
+To maintain backwards compatibility `info` rules are not included by default.  To include these rules you will need to include `-c I` or `--include-checks I`
 
 ### Metadata
 Inside the root level Metadata key you can configure cfn-lint using the supported parameters.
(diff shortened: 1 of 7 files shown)


fatbasstard:
Not sure if I agree on this rule...  It's a really specific enforcement based on company decisions, and what about all other resources that work with Deletion policies? Like S3 buckets for instance?

cmmeyer:
Dangit, @fatbasstard  -- I think you're right. We do have bet practices in the linter, but for the most part they are ones that would cause a failure on provisioning (not providing a CIDR in a parameter for the VPC). 

This one may be pushing us too far into governance style best practices, which is more of a cfn_nag thing.

Anyone else want to chime in here?

vandmo:
I think that it is definitely a rule that some will want to apply.
S3 buckets won't be deleted if they contain objects so that is a more safe default.

I agree that you should not force "best practices" on everyone although I think that a avoiding a dangerous default is more than a best practice. It could potentially save your data.

vandmo:
I think the safest approach is that you explicitly would need to opt-out of anything even slightly dangerous, but that way you **might** end up with a long list of excluded warnings.
IMO a long list of exludes is better than accidentally removing a DynamoDB table.

If the rule would apply to S3 on the other hand it would be more of style thing since the default behavior is more reasonable.
These kind of style rules should probably be rules that you need to opt-in to use.

cmmeyer:
So, @kddejong  and I have been chatting about this.  

What if we added a **new class of rule** for governance/best practices that we turned off by default?
That way existing users could continue with the less opinionated version, but you always have the option to turn on (and extend) governance?

Our working rule level was [INFO] (Ixxxx) which makes sense from a level perspective, but is hard to read.  Other options could be (Bxxxx) for best-practice or (Gxxxx) for governance. 

Thoughts?

vandmo:
I have been thinking in a similar way.
The problem is that there are to few categories.
I think that you should be able to choose if you want to opt-in or opt-out per category.
I also think that an item might belong to more than one category in some cases.

Just opting in per item could make you miss things that you actually want to have.
For example, in this specific case I think it is both "potential mistake" and "destructive".

I would definitely want to opt-in for all future items in these categories, but I might want to opt out for everything that is "minor ugliness".

Then there are of course cases where you need to make a choice, like "indent level" or similar things that are often just a matter of taste. Not sure how these could be handled though :)

These are some categories I have identified.
- potential mistake
- ugliness
- security
- destructive
- errors, will make template useless
- conformity, might fail in the future but works now
- purely matter of taste, like indent level, require alphabetical order, require explicit value even if default is safe (s3 deletion policy)

-

vandmo:
Anyhow, any way I could get this feature, I would be grateful :)

If I could, for example, do: cfn-lint --fail-on-best-practice-violation

And to be clear I think that adding deletion policy to a dynamodb table is a very good best practice while having to add it to an s3 bucket would be kind of annoying.
I would not mind to opt-out explicitly of such an s3 bucket rule if it would keep me from missing new best practices when I bump cfn-lint.

kddejong:
Waiting on #389 then we will switch this error code to Info level.

vandmo:
Cool, thanks :)

Issue #362, opened 2018-09-24
Validate UpdatePolicy for AutoScalingRollingUpdate 


adamchainz:
*cfn-lint version: (`cfn-lint --version`)* 0.7.2

*Description of issue.*

Using a template like:

```yaml
Resources:
  AutoScalingGroup:
    Type: AWS::AutoScaling::AutoScalingGroup
    Properties:
      # ...
      MinSize: "1"
      MaxSize: "1"
      # ...
    UpdatePolicy:
      AutoScalingRollingUpdate:
        MinInstancesInService: 1
```

will fail at deployment time with an error from CloudFormation like:

> AWS::AutoScaling::AutoScalingGroup AutoScalingGroup UPDATE_FAILED: MinInstancesInService must be less than the autoscaling group's MaxSize

It would be nice to have protection against this. I guess there are other constraints, some are [in the docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-updatepolicy.html#cfn-attributes-updatepolicy-rollingupdate). I submitted a [docs PR for this one](https://github.com/awsdocs/aws-cloudformation-user-guide/pull/164).

cmmeyer:
Agreed.  This would be a good business logic addition to compare values.  I'll label this as an enhancement.

kddejong:
Started with rule E3016 to check the basic configuration of the UpdatePolicy (attributes, values are of the correct type), etc.

Moving down the chain to this one next.

kddejong:
Sorry @kalpik I lost track of this.  I have created pull request #468 totally rewritten to handle conditions on each level.  Used your example for adding to tests.  Thanks for reporting this.

digash:
I am getting error E3016 UpdatePolicy doesn't support type AutoScalingRollingUpdate from this AWS example:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/quickref-autoscaling.html#scenario-as-updatepolicy

kddejong:
Closing this issue as it is covered in v1.   Any improvements we can do we will do as requested.

PR #378, opened 2018-10-02
New rule E3016 to check update policy configurations

kddejong:
*Issue #, if available:*
Start to #362
*Description of changes:*
- New rule E3016 to check the update policy configured on resources for basic syntax and type checking

### DIFF
--- a/src/cfnlint/rules/resources/updatepolicy/Configuration.py
+++ b/src/cfnlint/rules/resources/updatepolicy/Configuration.py
@@ -1,0 +1,215 @@
+"""
+  Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+  Permission is hereby granted, free of charge, to any person obtaining a copy of this
+  software and associated documentation files (the "Software"), to deal in the Software
+  without restriction, including without limitation the rights to use, copy, modify,
+  merge, publish, distribute, sublicense, and/or sell copies of the Software, and to
+  permit persons to whom the Software is furnished to do so.
+
+  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+  INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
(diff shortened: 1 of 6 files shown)


src/cfnlint/rules/resources/updatepolicy/Configuration.py
```diff
@@ -0,0 +1,221 @@
...
+            Check primitive types
+        """
+        matches = []
+
+        if isinstance(value, dict):
+            # this could be a ref, getatt, findinmap
+            return matches
+
+        default_message = 'Value for {0} must be of type {1}'
+        if prim_type == 'Boolean':
```
adamchainz: you can just put `bool`, `six.string_types` etc. in the dict above, then avoid this longwinded `if` block with

```python
if not isinstance(value, prim_type):
    matches.append(RuleMatch(path, default_message.format(path[-1], prim_type)))
```

sure you get `int` in the message instead of `Integer`, etc., but I think most people would understand what that means

kddejong: I was able to take this advice and clean it up a little bit (I think).  The hard part I ran into is that six.string_types is a tuple and not a class so that was a little harder to print.

src/cfnlint/rules/resources/updatepolicy/Configuration.py
```diff
@@ -0,0 +1,221 @@
...
+        default_message = 'Value for {0} must be of type {1}'
+        if not isinstance(value, prim_type):
+            if prim_type == six.string_types:
+                s_type = 'String'
+            elif prim_type == int:
+                s_type = 'Integer'
+            elif prim_type == list:
+                s_type = 'List'
+            elif prim_type == bool:
+                s_type = 'Boolean'
```
adamchainz: you could store the names in the dict too to avoid all these ifs - data not code :) Also I'd suggest `type_name` as the var name

src/cfnlint/rules/resources/updatepolicy/Configuration.py
```diff
@@ -0,0 +1,221 @@
...
+            matches.append(
+                RuleMatch(path, default_message.format(path[-1], s_type)))
+
+        if prim_type == list:
```
adamchainz: won't this currently trigger `default_message` twice, once from line 135 and once from line 140? don't you really just need `elif prim_type is list:` here (better to use `is` for comparing classes), and then do the block from line 143?

Issue #186, opened 2018-07-03
Feature: Warning on substitions without !Sub


fatbasstard:
*cfn-lint version: 0.3.2*

It's pretty annoying if you add a substition and forget the `!Sub`:
```
          - Effect: "Allow"
            Action:
              - "s3:ReplicateObject"
              - "s3:ReplicateDelete"
            Resource:
              - "arn:aws:s3:::mycompany-${Environment}-mybucketname/*"
```

I'm not 100% sure if a `${value}` always must be a substitution, since Cfn accepts this as valid string. But this could be a valid rule to give you a warning about this. Either you're missing the `!Sub` or you picked a real annoying string to be used...

fatbasstard:
Or this: 

`RoleName: !Sub "{AppGroup}-lambda"` 🤦‍♂️

cmmeyer:
It's also used a parameter expansion in bash, so this could reek havoc for folks with complex userdata scripts in their EC2 instances:
http://www.gnu.org/software/bash/manual/bash.html#Shell-Parameter-Expansion

fatbasstard:
Totally agree, it's not as simple as it appears due to stuff like that. Just wanted to put it out there anyway to make it visible. These 2 cases literally took me 2 hours of my day that I could spend on more usefull things than waiting for CloudFormation deployments to break 😈

cmmeyer:
I wonder if it makes sense to build a contextual for things like UserData where all bets are off -- I mean, there's nothing to stop you from injecting *another* CloudFormation template into a file via cfn-init in the EC2 metadata.  What would THAT do to our linter?

fatbasstard:
Which properties does this (the parameter extension) apply to? Because we could make it a rule and just skip properties like `UserData`... But check the rest 🤔

cmmeyer:
Off hand, I'm thinking EC2 user data, Lambda function code blocks... others?

SanderKnape:
There might be another (partial) solution for this. For example, looking at the SQS resource (https://docs.aws.amazon.com/sns/latest/api/API_CreateTopic.html), a dollar sign is not allowed for the TopicName. That is something you can easily lint.

I can imagine that for many properties, a dollar sign is not an allowed character.

kddejong:
This one is released now.

PR #387, opened 2018-10-09
Rule to error on unsubstituted parameters

Tro95:
Issue #186

This additional rule will check for the existence of `${parameters}` in the template, and validate whether a corresponding `Fn::Sub` (`!Sub` gets transformed into `Fn::Sub`) exists higher up the path to correctly substitute the parameter.

The reason for making this an error instead of a warning is because the [definition used by this project for warning](https://github.com/awslabs/cfn-python-lint/blob/master/docs/rules.md#warnings) only considers best-practice as opposed to potentially breaking changes.

There are a few complex use cases when this rule should ignore fields to avoid generating a false-positive, such as inline lambda code. There is an `excludes` list defined, which I expect will need adjustment in the future as I definitely haven't captured all possible false-positives.

### DIFF
--- a/src/cfnlint/rules/functions/SubNeeded.py
+++ b/src/cfnlint/rules/functions/SubNeeded.py
@@ -26,6 +26,9 @@ class SubNeeded(CloudFormationLintRule):
     source_url = 'https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-sub.html'
     tags = ['functions', 'sub']
 
+    # Free-form text properties to exclude from this rule
+    excludes = ['UserData', 'ZipFile']
+
     def _match_values(self, searchRegex, cfndict, path, cfn):
         """Recursively search for values matching the searchRegex"""
         values = []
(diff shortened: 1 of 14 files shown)


src/cfnlint/rules/functions/SubNeeded.py
```diff
@@ -0,0 +1,87 @@
...
+        for parameter_string_path in parameter_string_paths:
+            found_sub = False
+
+            # Does the path contain an 'Fn::Sub'?
+            for step in parameter_string_path:
+                if step == 'Fn::Sub' or step in self.excludes:
+                    found_sub = True
+
+            # If we didn't find an 'Fn::Sub' it means a string containing a ${parameter} may not be evaluated correctly
+            if not found_sub:
```
kddejong: Should we also be checking for other scenarios where ${} may exist validly?  Imbedded Lambda code (ZipFile), user data?  I'm trying to think of other scenarios... but that is all I could come up with.   I was thinking in CloudWatch event transforms or API Gateway but I don't think they would show up there.

A possible example:
https://github.com/matsev/cloudformation-api-gateway/blob/master/cloudformation.template#L18

Tro95: That's what the self.excludes list does. See lines 30 and 79. I can definitely see the list expanding in the future as these valid cases are discovered, but I've covered the main two that I could think of (and the `watchmaker.json` file includes the UserData scenario).

kddejong: Woops missed that.  Sorry.

PR #389, opened 2018-10-11
Feature/add info type

kddejong:
*Issue #, if available:*

*Description of changes:*
- Allows for a new rule type of label 'I'.  This rule will be disabled by default and will need to be included by doing `--include-checks I`


test/module/test_rules_collections.py
```diff
@@ -66,3 +73,108 @@ def test_fail_sub_properties_run(self):
...
+        class rule_e0000(CloudFormationLintRule):
+            """Error Rule"""
+            id = 'E0000'
+
+        class rule_w0000(CloudFormationLintRule):
+            """Warning Rule"""
+            id = 'W0000'
+
+        class rule_i0000(CloudFormationLintRule):
+            """Error Rule"""
```
cmmeyer: I don't think you need this class for the default check?

test/module/test_rules_collections.py
```diff
@@ -66,3 +73,108 @@ def test_fail_sub_properties_run(self):
...
```
cmmeyer: Should be "Check Rule" or "Info Rule"

test/module/test_rules_collections.py
kddejong: I do.  I need to setup rules with IDs for testing purposes.

PR #411, opened 2018-10-25
Error rule if subnet referenced in multiple SubnetRouteTableAssociations 

cmmeyer:
*Issue #, if available:*

*Description of changes:*
Check that only one route table association is defined per subnet.  Best effort to evaluate Refs, Fn::GetAtts and Conditions to sniff out duplicates.


src/cfnlint/rules/resources/ectwo/RouteTableAssociation.py
```diff
@@ -0,0 +1,86 @@
...
+                        if isinstance(value, list):
+                            if len(value) == 3:
+                                condition = value[0]
+                                values.extend(self.get_values(value[1], condition))
+                                values.extend(self.get_values(value[2], condition))
+                    if key in ('Ref', 'GetAtt'):
+                        values.extend(self.get_values(value, condition))
+        else:
+            if condition:
+                values.append(condition + '::' + subnetid)
```
kddejong: Any reason we wouldn't just do this as a tuple directly right away?

src/cfnlint/rules/resources/ectwo/RouteTableAssociation.py
```diff
@@ -0,0 +1,86 @@
...
+                for key, value in subnetid.items():
+                    if key in cfnlint.helpers.CONDITION_FUNCTIONS:
+                        if isinstance(value, list):
+                            if len(value) == 3:
+                                condition = value[0]
+                                values.extend(self.get_values(value[1], condition))
+                                values.extend(self.get_values(value[2], condition))
+                    if key in ('Ref', 'GetAtt'):
+                        values.extend(self.get_values(value, condition))
+        else:
```
kddejong: is this else is important.  I'm trying to run scenarios in my head related to Conditions (and lets be clear condition handling is hard especially across resources).  What may be more appropriate is a bale out when things get too crazy.    The scenario here I'm wondering about is what happens if the Resource has a condition and property SubnetId has a condition.  In this case the subnetid could be a dict and a condition parameter was passed in but would never get added to values because of the else.

src/cfnlint/rules/resources/ectwo/RouteTableAssociation.py
```diff
@@ -0,0 +1,86 @@
...
+
+        return matches
+
+    def match(self, cfn):
+        """Check SubnetRouteTableAssociation Resource Properties"""
+        matches = []
+
+        resources = cfn.get_resources(['AWS::EC2::SubnetRouteTableAssociation'])
+        for resource_name, resource in resources.items():
+            properties = resource.get('Properties')
```
kddejong: This many not be condition safe.  
```
Properties:
  Fn::If:
  - myCondition
  - SubnetId: !Ref subneta
  - SubnetId: !Ref subnetb
```

src/cfnlint/rules/resources/ectwo/RouteTableAssociation.py
rjlohan: I actually talked to Chuck offline about this and recommended that Conditions just be taken out of the picture. Just look for duplicates on the inline string/ref target and leave it at that. It's a small step for man.

cmmeyer:
I still suck at git.  This PR is a mess -- if anyone wants to teach me how to fix it I will try be better

PR #415, opened 2018-10-26
New rule E3022 to validate SubnetIDs are only used in one Route Table association

cmmeyer:
*Issue #, if available:*

*Description of changes:*
Subnets can only have a single route table association. If there are two associations to the same subnet things will go poorly.  This tries to find dups across conditions.


rjlohan:
Glorious. Catches exactly the error I wanted it to. 🥇

PR #419, opened 2018-10-26
Catch AttributeError when rules to try to access the wrong attribute

### LINKED ISSUE #403: Multiple 'str_node' object has no attribute 'items' errors with bad Outputs section
0.8.1

*Description of issue.*
Multiple 'str_node' object has no attribute 'items' errors with bad Outputs section:

```
E0002 Unknown exception while processing rule E6001: 'str_node' object has no attribute 'items'
rds-cluster-template.yml:1:1

E0002 Unknown exception while processing rule E6005: 'str_node' object has no attribute 'items'
rds-cluster-template.yml:1:1

E0002 Unknown exception while processing rule E6012: 'str_node' object has no attribute 'items'
rds-cluster-template.yml:1:1

E0002 Unknown exception while processing rule E6004: 'str_node' object has no attribute 'items'
rds-cluster-template.yml:1:1

E0002 Unknown exception while processing rule E6002: 'str_node' object has no attribute 'items'
rds-cluster-template.yml:1:1
```

I recognize this template makes no sense, but we should probably do a better job of cacthing the problem.

Template:

```yaml
---
Resources:
  EC2:
    Type: "AWS::EC2::Instance"
    Properties:
      ImageId: "i-2395829568"
Outputs:
  PathToMasterStackFile
```

kddejong:
*Issue #, if available:*
Fix #403

*Description of changes:*
- Create an items class in the cfn-lint str node object that has items defined but returns nothing.  The assumption being rules should exist to catch the configuration errors but this prevents really improperly configured templates from breaking rules.
- Update rule E6001 so it provides an error when Outputs isn't an object/dict

### DIFF
--- a/src/cfnlint/decode/node.py
+++ b/src/cfnlint/decode/node.py
@@ -15,8 +15,11 @@
   SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 """
 import sys
+import logging
 from copy import deepcopy
 import six
+
+LOGGER = logging.getLogger(__name__)
 
 
 def create_str_node_class(cls):
--- a/src/cfnlint/rules/outputs/Configuration.py
+++ b/src/cfnlint/rules/outputs/Configuration.py
@@ -40,13 +40,16 @@ class Configuration(CloudFormationLintRule):
 
         outputs = cfn.template.get('Outputs', {})
         if outputs:
-            for output_name, output_value in outputs.items():
-                for prop in output_value:
-                    if prop not in self.valid_keys:
-                        message = 'Output {0} has invalid property {1}'
-                        matches.append(RuleMatch(
-                            ['Outputs', output_name, prop],
-                            message.format(output_name, prop)
-                        ))
+            if isinstance(outputs, dict):
+                for output_name, output_value in outputs.items():
+                    for prop in output_value:
+                        if prop not in self.valid_keys:
+                            message = 'Output {0} has invalid property {1}'
+                            matches.append(RuleMatch(
+                                ['Outputs', output_name, prop],
+                                message.format(output_name, prop)
+                            ))
+            else:
+                matches.append(RuleMatch(['Outputs'], 'Outputs do not follow correct format.'))
 
         return matches
--- a/src/cfnlint/__init__.py
+++ b/src/cfnlint/__init__.py
@@ -172,6 +172,22 @@ class RulesCollection(object):
 
         return True
 
+    def run_check(self, check, filename, rule_id, *args):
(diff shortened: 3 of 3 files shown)


kddejong:
I submitted as this would be a possible solution to the problem.  This isn't complete though and we may need to override more functions.

The other option I have is adding these functions and then raising an Error in them.  We could silently catch and log the error then.

src/cfnlint/rules/outputs/Configuration.py
```diff
@@ -40,13 +40,16 @@ def match(self, cfn):
...
+                for output_name, output_value in outputs.items():
+                    for prop in output_value:
+                        if prop not in self.valid_keys:
+                            message = 'Output {0} has invalid property {1}'
+                            matches.append(RuleMatch(
+                                ['Outputs', output_name, prop],
+                                message.format(output_name, prop)
+                            ))
+            else:
+                matches.append(RuleMatch(['Outputs'], 'Outputs need to be an object'))
```
cmmeyer: This message is a little too code-centric.  Perhaps "Outputs do not follow correct format."

src/cfnlint/decode/node.py
```diff
@@ -51,6 +54,12 @@ def __deepcopy__(self, memo):
         def __copy__(self):
             return self
 
+        def items(self):
+            """ Provide items so we don't have failures with really bad coded templates """
+            LOGGER.warning('Strings don\'t support items')
```
cmmeyer: Is there any more context we can give here? Obviously something has gone horribly wrong and it would be nice to provide some info for troubleshooting either the template or the code.  Can we pass the string?"

kddejong: I think so.

kddejong:
Updated this pull request to catch all AttributeErrors and silently continue.  Reducing malformed templates from triggering errors when certain expectations are created.   I wanted to Warn but looks like Warning may mess with IDEs and json based outputs.

adamchainz:
I think catching *all* `AttributeError`s from all rules could cause more pain in the future than it solves now. Developing a rule, one might write a call to a library function and mistype it, e.g. `json.laods`, and then the rule will silently fail. This is the kind of error muting that can make a developer go insane hunting for it 😨

It's always better to push exception handling right around the block of code you know can cause it, especially for such a generic exception as `AttributeError`, e.g.

```python
try:
    data = my_template.an_attribute.another
except AttributeError:
    return []

if looks_wrong(data):
    return [the_error]
```

kddejong:
Yea agreed.  I had a few other approaches for this that I will go after instead.

PR #424, opened 2018-10-29
Move arrays into init function to prevent copying of settings between rules

### LINKED ISSUE #421: Unknown exception from rule E3021
*cfn-lint version: (`cfn-lint --version`)* 0.8.2

*Description of issue.*

```
$ cat test.yml

Resources:

  GlueCrawler:
    Type: AWS::Glue::Crawler
    Properties:
      Role: myrole
      DatabaseName: mydbname
      Targets:
        S3Targets:
        - Path:
            !Sub
            - s3://${BucketName}/
            - BucketName:
                Fn::ImportValue: my-bucket-name
$ cfn-lint --regions us-east-1 --template test.yml
E0002 Unknown exception while processing rule E3021: list index out of range
test.yml:1:1

```

In addition I think the 'unknown exception' message is pretty unhelpful as the stack trace is gone. But at least I could cut my template down to this minimal example.

kddejong:
*Issue #, if available:*
Fix #421
*Description of changes:*
- When setting parent class attributes in the init using self they would be stored and copied to the next rule.  

This rule should have never been triggered with the supplied template and that is because of this copying.  This will force empty lists on initialization. 

### DIFF
--- a/src/cfnlint/__init__.py
+++ b/src/cfnlint/__init__.py
@@ -38,8 +38,10 @@ class CloudFormationLintRule(object):
     tags = []
 
     logger = logging.getLogger(__name__)
-    resource_property_types = []
-    resource_sub_property_types = []
+
+    def __init__(self):
+        self.resource_property_types = []
+        self.resource_sub_property_types = []
 
     def __repr__(self):
         return '%s: %s' % (self.id, self.shortdesc)
--- a/src/cfnlint/rules/functions/DynamicReferenceSecureString.py
+++ b/src/cfnlint/rules/functions/DynamicReferenceSecureString.py
@@ -34,6 +34,7 @@ class DynamicReferenceSecureString(CloudFormationLintRule):
 
     def __init__(self, ):
         """Init """
+        super(DynamicReferenceSecureString, self).__init__()
         specs = cfnlint.helpers.RESOURCE_SPECS.get('us-east-1')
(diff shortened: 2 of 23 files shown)


src/cfnlint/__init__.py
```diff
@@ -38,8 +38,10 @@ class CloudFormationLintRule(object):
     tags = []
 
     logger = logging.getLogger(__name__)
-    resource_property_types = []
-    resource_sub_property_types = []
+
+    def __init__(self):
+        self.resource_property_types = []
+        self.resource_sub_property_types = []
```
adamchainz: oh, fun :)

src/cfnlint/__init__.py
```diff
@@ -38,8 +38,10 @@ class CloudFormationLintRule(object):
     tags = []
```
adamchainz: this might have the same problem?

src/cfnlint/__init__.py
kddejong: yea learned something new here and not a fun thing to find.

src/cfnlint/__init__.py
kddejong: It could.  We have been specifying the tags in the attributes of the sub classes.  If anyone was doing the population using the __init__ this would probably pop up.  I suppose the prudent thing would be to fix it before that happens.

PR #429, opened 2018-10-30
Read configuration file processing cfnlint settings

### LINKED ISSUE #395: Should we have a configuration file
*Description of issue.*
With the new info rules we are starting to expand the amount of command line options and including, excluding rules.  I'm wondering if we need to allow a configuration file we can read and use on execution. 

For instance I want to have info rules always on but I don't want to type in the parameters every time.  Because I'm lazy.  Another use case as the rules expand it would be easier to disable certain rules with a config file instead of creating longer and longer command line parameters.

This would be the least precedence in parameters.  Files and command line parameters will override the configuration file.

kddejong:
*Issue #, if available:*
Fix #395
*Description of changes:*
- Create the ability to read a configuration file instead of using parameters
- Moved template names to use glob which supports cross platform and wildcards in the configuration file
- Rewrote the extend for argparse to override defaults when supplied in the config file and command line
- Order of processing is config file, template file, command line


kddejong:
Need to validate the precedence when using configuration in the template

PR #437, opened 2018-11-05
Add rule to check if referenced Conditions are actually defined

fatbasstard:
Add a rule to check if the conditions specified are actually defined.

Found out this check was missing when trying to deploy the following:

```
  ApplicationLoadbalancerWebACLAssociation:
    Type: "AWS::WAFRegional::WebACLAssociation"
    Condition: false 
    Properties:
      ...
```
(Yes, that's special, I know... 😉 )

Added a new rule that's basically the invert of the `Used` check. Removed the function `ifexists` since that covered a subset of this rule (checking the existence of the Conditions used in `Fn::If` functions)

While creating the rule I discovered the Resource's Conditionals are added twice in the `Used` check...


src/cfnlint/rules/conditions/Used.py
```diff
@@ -43,10 +43,6 @@ def match(self, cfn):
                     ref_conditions.append(iftree[-1][0])
                 else:
                     ref_conditions.append(iftree[-1])
-            # Get resource's Conditions
-            for _, resource_values in cfn.get_resources().items():
-                if 'Condition' in resource_values:
-                    ref_conditions.append(resource_values['Condition'])
```
kddejong: Isn’t this still necessary to see if a Condition is used?

fatbasstard: It was a duplicate. The Resource conditions are added twice 😉

docs/rules.md
```diff
@@ -113,7 +114,7 @@ The following **97** rules are applied by this linter:
 | E7011 <a name="E7011"></a> | Mapping name limit not exceeded | Check the size of Mapping names in the template is less than the upper limit | [Source](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cloudformation-limits.html) | `mappings`,`limits` |
 | E7012 <a name="E7012"></a> | Mapping attribute limit not exceeded | Check if the amount of Mapping attributes in the template is less than the upper limit | [Source](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cloudformation-limits.html) | `mappings`,`limits` |
 | E8001 <a name="E8001"></a> | Conditions have appropriate properties | Check if Conditions are properly configured | [Source](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/conditions-section-structure.html) | `conditions` |
-| I3011 <a name="I3011"></a> | Check DynamoDB tables have a set DeletionPolicy | The default action when removing a DynamoDB Table is to delete it. This check requires you to specifically set a DeletionPolicy and you know the risks | [Source](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-deletionpolicy.html) | `resources`,`dynamodb` |
```
kddejong: Make sure you run the documentation creator with include_checks I

docs/rules.md
```diff
@@ -113,7 +114,7 @@ The following **97** rules are applied by this linter:
...
+| E8002 <a name="E8002"></a> | Check if the referenced Conditions are defined | Making sure the used conditions are actually defined in the Conditions section | [Source](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/conditions-section-structure.html) | `conditions` |
```
kddejong: Any conflicts with 
https://github.com/awslabs/cfn-python-lint/blob/master/src/cfnlint/rules/functions/IfExist.py

docs/rules.md
fatbasstard: Wut... Oh, can I make that the default behaviour?

docs/rules.md
fatbasstard: I think that rule is made obsolete by this one... I'll doublecheck and remove it if that's the case

kddejong: Of course but it would take a little work.   The config pull request may be required first.
1. Right now it spits out the documentation based on the loaded rules.  So we can modify that behavior to document all rules no matter what.  Basically we would overwrite the included and ignore checks when doing a documentation dump so it covers everything.
2. You have a cfnlintrc config file that would always load all the rules.

docs/rules.md
kddejong: Of course but it would take a little work. The config pull request may be required first.

Right now it spits out the documentation based on the loaded rules. So we can modify that behavior to document all rules no matter what. Basically we would overwrite the included and ignore checks when doing a documentation dump so it covers everything.
You have a cfnlintrc config file that would always load all the rules.

PR #441, opened 2018-11-05
Support SSM lists when evaluating refs for lists in E3002

### LINKED ISSUE #439: E3002 error thrown when using parameter lists
*cfn-lint version cfn-lint 0.8.3

*Description of issue. When using SSM parameters that are type list, the linter doesn't properly recognize them as a list and thows an error E3002 

Specifically E3002 Property PreferredAvailabilityZones should be of type List or Parameter should be a list for resource cacheRedisV1
us-east-1.yaml:249:7

The parameter section for this is:
Parameters:
  azList:
    Type: "AWS::SSM::Parameter::Value<List<String>>"
    Description: "The list of AZs from Parameter Store"
    Default: '/regionSettings/azList'

The resource is defined as:
cacheForumRedisV1:
    Type: AWS::ElastiCache::CacheCluster
    Properties:
      PreferredAvailabilityZones: !Ref azList

kddejong:
*Issue #, if available:*
Fix #439 

*Description of changes:*
- Handle SSM List Parameters when evaluating rule E3002 and Refs to parameters.

### DIFF
--- a/test/fixtures/templates/good/resource_properties.yaml
+++ b/test/fixtures/templates/good/resource_properties.yaml
@@ -218,6 +218,10 @@ Outputs:
           - ElasticLoadBalancer
           - DNSName
 Parameters:
+  azList:
+    Type: "AWS::SSM::Parameter::Value<List<String>>"
+    Description: "The list of AZs from Parameter Store"
+    Default: '/regionSettings/azList'
   InstanceType:
     AllowedValues:
     - t1.micro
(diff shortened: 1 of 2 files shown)


test/fixtures/templates/good/resource_properties.yaml
```diff
@@ -218,6 +218,10 @@ Outputs:
           - ElasticLoadBalancer
           - DNSName
 Parameters:
+  azList:
+    Type: "AWS::SSM::Parameter::Value<List>"
```
salexpdx: I just added a note on the issue, GH didn't render what i was trying to show properly. 
"AWS::SSM::Parameter::Value\<List\<String\>\>"

is the correct syntax here.

kddejong: Thanks.

PR #446, opened 2018-11-06
 Support validating custom resource types with --override-spec

### LINKED ISSUE #445: Custom resource types can't be validated with --override-spec
*cfn-lint version:* 0.8.3

*Description of issue:* When using custom resource types and supplying an override spec for the custom type, no property validations seem to be performed.

Example: specify a `Custom::Example` type with a required `ID` property and an optional Boolean `Enabled` property:

test.cf.yaml:
```yaml
Resources:
  GoodExample:
    Type: Custom::Example
    Properties:
      ServiceToken: example.arn
      ID: good
      Enabled: true

  BadExample:
    Type: Custom::Example
    Properties:
      ServiceToken: example.arn
      # Missing required prop `ID`
      Enabled: [invalid, 'must be a boolean']
      Wrong: 'this prop is not in the spec'
```

CustomExampleSpec.json:
```json
{
  "ResourceTypes": {
    "Custom::Example": {
      "Properties": {
        "ServiceToken": {"PrimitiveType": "String", "Required": true, "UpdateType": "Immutable"},
        "ID": {"PrimitiveType": "String", "Required": true, "UpdateType": "Immutable"},
        "Enabled": {"PrimitiveType": "Boolean", "Required": false, "UpdateType": "Mutable"}
      }
    }
  },
  "ResourceSpecificationVersion": "2.11.0"
}
```

Run: `cfn-lint --override-spec CustomExampleSpec.json test.cf.yaml`

Results: No lint errors or warnings

Expected:
```
E3003 Property ID missing at Resources/BadExample/Properties
test.cf.yaml:11:5

E3002 Property should be of type Boolean not List at Resources/BadExample/Properties/Enabled
test.cf.yaml:14:7

E3002 Invalid Property Resources/BadExample/Properties/Wrong
test.cf.yaml:15:7
```

This is occurring because #295 unconditionally forces any type starting with `Custom::` to `AWS::CloudFormation::CustomResource`, which deliberately allows all properties. For unknown custom types, that's incredibly useful. But if you have a known spec for the custom type, it would be good to use that instead.

I think this can be accomplished by changing `if resourcetype.startswith('Custom::')` to add `... and not resourcetype in self.resourcetypes` in both [Properties.py](https://github.com/awslabs/cfn-python-lint/blob/v0.8.3/src/cfnlint/rules/resources/properties/Properties.py#L270) and [Required.py](https://github.com/awslabs/cfn-python-lint/blob/v0.8.3/src/cfnlint/rules/resources/properties/Required.py#L132).

medmunds:
Fixes #445.

*Description of changes:*

If a custom resource type is specified via `--override-spec`, use that for validation (rather than the generic `AWS::CloudFormation::CustomResource` default for `Custom::___` resources).

[Fix is straightforward; it was less clear to me where best to add the tests and fixtures. Please feel free to reorganize tests as you prefer—I kept them a separate commit.]

### DIFF
--- a/src/cfnlint/rules/resources/properties/Properties.py
+++ b/src/cfnlint/rules/resources/properties/Properties.py
@@ -267,7 +267,7 @@ class Properties(CloudFormationLintRule):
         for resourcename, resourcevalue in cfn.get_resources().items():
             if 'Properties' in resourcevalue and 'Type' in resourcevalue:
                 resourcetype = resourcevalue.get('Type', None)
-                if resourcetype.startswith('Custom::'):
+                if resourcetype.startswith('Custom::') and resourcetype not in self.resourcetypes:
                     resourcetype = 'AWS::CloudFormation::CustomResource'
                 if resourcetype in self.resourcetypes:
                     path = ['Resources', resourcename, 'Properties']
--- a/src/cfnlint/rules/resources/properties/Required.py
+++ b/src/cfnlint/rules/resources/properties/Required.py
@@ -129,7 +129,7 @@ class Required(CloudFormationLintRule):
         for resourcename, resourcevalue in cfn.get_resources().items():
             if 'Properties' in resourcevalue and 'Type' in resourcevalue:
                 resourcetype = resourcevalue['Type']
-                if resourcetype.startswith('Custom::'):
+                if resourcetype.startswith('Custom::') and resourcetype not in self.resourcetypes:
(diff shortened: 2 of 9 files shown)


fatbasstard:
Literally yesterday I've had a quick discussion with a colleague about this "missing feature". Awesome, simple change that is a nice enabler! 👍

kddejong:
This is a cool feature.  Thanks!

PR #452, opened 2018-11-12
Remove AttributeError as a catch all

kddejong:
*Issue #, if available:*

*Description of changes:*
- Remove AttributeError as a catch all when processing rules

### DIFF
--- a/src/cfnlint/__init__.py
+++ b/src/cfnlint/__init__.py
@@ -24,6 +24,7 @@ import six
 from yaml.parser import ParserError
 import cfnlint.helpers
 from cfnlint.transform import Transform
+from cfnlint.decode.node import TemplateAttributeError
 
 LOGGER = logging.getLogger(__name__)
 
@@ -176,6 +177,9 @@ class RulesCollection(object):
         """ Run a check """
         try:
             return check(*args)
+        except TemplateAttributeError as err:
+            LOGGER.debug(str(err))
+            return []
         except Exception as err:  # pylint: disable=W0703
             if self.is_rule_enabled('E0002'):
                 message = 'Unknown exception while processing rule {}: {}'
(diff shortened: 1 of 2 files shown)


adamchainz:
It's kind of weird to do this, it effectively duck-types strings and lists as dictionaries. Not sure what I really think about it without full knowledge of the errors you're trying to prevent.

kddejong:
@adamchainz another idea. 

Create this function in our custom classes
```
def __getattr__(self, name):
            value = self.__dict__.get(name)
            if not value:
                # Raise AttributeError if attribute value not found.
                raise CfnAttributeError('%s.%s is invalid' % (self.__class__.__name__, name))

            return value
```

Then we can catch this error like we had been with AttributeError but we know it will specifically thrown from the correct objects.

adamchainz:
That's cleaner yes. The stopped-processing problem existed with the catch-all of `AttributeError`.

Btw `__getattr__` is called only for missing attributes (it's `__getattribute__` that is called for *all* attribute lookups), so you could just use this implementation:

```python
def __getattr__(self, name):
    raise CfnAttributeError(...)
```

One other option is to allow inter-rule dependencies, have one that validates the type structure of e.g. an Output, and then have others depend on that so they can assume the types are all good.

kddejong:
```
One other option is to allow inter-rule dependencies, have one that validates the type structure of e.g. an Output, and then have others depend on that so they can assume the types are all good.
```

Interesting.  I like that idea too.   I may try this out and see what it would like to get implemented.  I like the simplicity and the better guarantee to rules that the structure is what they would expect.  

Thanks for the tip on the `__getattr__`.  I'm going to make those updates and we can discuss if that makes sense as a short term solution.

kddejong:
This pull request now uses the `__getattr__` to raise a custom error we can catch and log to make things a little cleaner.

PR #453, opened 2018-11-12
Need str joins to cover list numbers from Fn::If

kddejong:
*Issue #, if available:*

*Description of changes:*
- Update joins for printing Path to include converting ints to strs.  Helps cover cases where there could be a Fn::If with numbers for the path.


kddejong:
Abbreviation of Lambda to lmbd is because lambda is a keyword in Python.

I think this could cause some problems down the chain.  I'm sure we could shoe horn it in but I think we would have a lot of spots to change to make it work. Since this code is in a lot of spots the other thing could be a helper that does this work for us so we aren't repeating it all over the place.  

Or something we just append to all messages.

cmmeyer:
Duh. Now I remember the lambda discussion from way back. I'm a bad dev. 

On the rest -- Fair enough, but I had to ask.

Issue #417, opened 2018-10-26
Better initial errors for poorly formed YAML/JSON


cmmeyer:
*cfn-lint version: (`cfn-lint --version`)*
0.8.2

*Description of issue.*
Some of the entry-point errors for really broken templates are very cryptic.  We should identify situations where it makes more sense to drop back to general YAML/JSON linting or a generic "JSON/YAML is unparseable" error.

Our error:
`E0000 Null value at line 10 column 37`

CloudFormation `aws cloudformation validate-template` error:
`An error occurred (ValidationError) when calling the ValidateTemplate operation: Template format error: [/Resources/PublicSubnetRouteTableAssociation1] resource definition is malformed`

We should be at least as good.

Template:

```yaml
Parameters:
  PublicRouteTable:
    Type: String
    Default: rt-123
  PublicSubnet01:
    Type: String
    Default: s-123456
Conditions:
  CreateAuxiliaryNetworks: !Equals [ !Ref PulicSubnet01, prod ]
Resources:
  PublicSubnetRouteTableAssociation1:
  Type: AWS::EC2::SubnetRouteTableAssociation
  Properties:
    RouteTableId: {Ref: PublicRouteTable}
    SubnetId: {Ref: PublicSubnet01}

  AuxiliaryPublicSubnetRouteTableAssociation1:
    Type: AWS::EC2::SubnetRouteTableAssociation
    Condition: CreateAuxiliaryNetworks
    Properties:
      RouteTableId: {Ref: PublicRouteTable}
      SubnetId: {Ref: PublicSubnet01}
```

kddejong:
So you want the location in there as well (/Resources/PublicSubnetRouteTableAssociation1)?  Obviously we are providing the location based on column and line # but the message itself is a little lacking.  Thats cause we are detecting the null errors during the parse operations and not as a separate rule.

rjlohan:
I'm not sure even the line # is correct here? It complains about line 10, but it's line 12-15 where my indenting got broken.

kddejong:
The location matches for the `PublicSubnetRouteTableAssociation1` as it has a null value in this case.

kddejong:
I think there is an opportunity here to change some things and make the messages better.   I'm just trying to make sure I understand the goal correctly.

[...]

PR #455, opened 2018-11-13
The Yaml error message for null was off by a line and column

kddejong:
*Issue #, if available:*
#417
*Description of changes:*
- The description on Null Errors when parsing YAML had a line and column that was off by one


kddejong:
I did test JSON too and that was printing correctly.

PR #463, opened 2018-11-15
remove exit calls from integration code path

jaymccon:
*https://github.com/aws-quickstart/taskcat/pull/196*

*Description of changes:*

When using cfn-python-lint as a module, calls to `exit` are problematic, as they kill the host process and don't give me any options on whether I want to handle the error, or do any cleanup.

This pr removes all `exit` calls from the integration code path, as defined in the [docs](https://github.com/awslabs/cfn-python-lint/blob/master/docs/getting_started/integration.md). It replaces them with a couple of exception types so that an integration developer can choose how to handle different failures. Those exception types are then handled in `__main__`, so functionality should not change for users of the cli.

Also adds some tests and removed a few unused imports.


kddejong:
Makes sense.  Going to do a little more testing but at first pass it looks good.

src/cfnlint/core.py
```diff
@@ -142,11 +156,10 @@ def get_template_rules(filename, args):
 def run_checks(filename, template, rules, regions):
     """Run Checks against the template"""
     if regions:
-
-        for region in regions:
-            if region not in REGIONS:
-                LOGGER.error('Supported regions are %s', REGIONS)
-                exit(32)
+        if not set(regions).issubset(set(REGIONS)):
```
kddejong: 👍

src/cfnlint/core.py
```diff
@@ -157,8 +170,8 @@ def run_checks(filename, template, rules, regions):
         try:
             matches.extend(runner.run())
         except Exception as err:  # pylint: disable=W0703
-            LOGGER.error('Tried to process rules on file %s but got an error: %s', filename, str(err))
-            exit(1)
+            msg = 'Tried to process rules on file %s but got an error: %s' % (filename, str(err))
+            UnexpectedRuleException(msg, 1)
```
PatMyron, line 174: Exception should probably be raised:
https://lgtm.com/projects/g/aws-cloudformation/cfn-lint/

PR #469, opened 2018-11-19
Add a rule to test RDS instance sizes against the engine and licensemodel

### LINKED ISSUE #270: RDS instance types
Apologies for another enhancement ticket, but just ran into this problem with a development team.

*Description of issue.*

The team was using:

```
  DatabaseDBClusterParameterGroup:
    Type: "AWS::RDS::DBClusterParameterGroup"
    Properties:
      Family: "aurora5.6"

  DBInstance1a:
    Type: "AWS::RDS::DBInstance"
    Properties:
      DBInstanceClass: "db.m4.xlarge"
```

Which in CloudFormation will lead to the following eventual error:

> RDS does not support creating a DB instance with the following combination: DBInstanceClass=db.m4.xlarge, Engine=aurora, EngineVersion=5.6.10a, LicenseModel=general-public-license. For supported combinations of instance class and database engine version, see the documentation.

Developer points to https://aws.amazon.com/rds/instance-types/ which has the db.m4.xlarge on it. It's only when you scroll to the absolute bottom that you see, in a smaller font mind you, the following:

> Not every instance type is supported for every database engine, version, edition or region. Please see the DB engine-specific pricing pages for details.

Which leads to https://aws.amazon.com/rds/aurora/pricing/ and that shows only r3/r4 is available for Aurora MySQL.

This seems like something that is easily lintable, provided there is some auxiliary files like the CloudFormation spec files. So is there something similar available for instance types available for these aspects?

kddejong:
*Issue #, if available:*
Fix #270
*Description of changes:*
- Check RDS instance types against the region, engine, and license model.  Only checks if they are hardcoded values and not REFed or IFed.

### DIFF
--- a/src/cfnlint/data/AdditionalSpecs/RdsProperties.json
+++ b/src/cfnlint/data/AdditionalSpecs/RdsProperties.json
@@ -1,0 +1,566 @@
+[
+  {
+    "engines": [
+      "aurora",
+      "aurora-mysql"
+    ],
+    "instance_types": [
+      "db.t2.small",
+      "db.t2.medium",
+      "db.r4.large",
+      "db.r4.xlarge",
+      "db.r4.2xlarge",
+      "db.r4.4xlarge",
+      "db.r4.8xlarge",
+      "db.r4.16xlarge"
+    ],
+    "regions": [
+      "us-east-1",
+      "us-east-2",
+      "us-west-1",
+      "us-west-2",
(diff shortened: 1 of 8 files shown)


kddejong:
I'm a little torn on this.  Coming up with a really good defined list of valid instance types is a little difficult.  I would appreciate extra validation of my work on this list.

PR #473, opened 2018-11-20
Implement new approach on the Spec files

fatbasstard:
WIP


fatbasstard:
1. Json patch configuration is pretty complex to write down and maintain, especially with a lot of changes. Well, at least that is/was my initial gut feeling... I will try to get that approch in there.

2. I can also try to apply that mapping to allowed values

3. AllowedPattern was already a use case I had in mind when creating this. Since a pattern can be powerful for all the length and alphanumeric rules some properties had

In short: I'll continue on revamping this. Thanx for the feedback!

fatbasstard:
Ok, just restructured the Spec file loading and changed the property values into a Patch file.

ToDo:

- Integrate Allowed values into the property values
- Implement sorting mechanism (so that the spec patches always come first)?

fatbasstard:
Moved in 1 allowed value into the property spec Patch to see how that works out...

kddejong:
This is looking pretty good now.  How you feeling about it @fatbasstard ?     Lots of changes here so I want to spend some time digging into it more when you're ready.

src/cfnlint/data/ExtendedSpecs/append_spec/allowed_values.json
```diff
@@ -0,0 +1,240 @@
+{
```
kddejong: Is this just a place holder file for now?

src/cfnlint/data/ExtendedSpecs/append_spec/allowed_values.json
```diff
@@ -0,0 +1,240 @@
...
+            "nodejs8.10",
+            "java8",
+            "python2.7",
+            "python3.6",
+            "python3.7",
+            "dotnetcore1.0",
+            "dotnetcore2.0",
+            "dotnetcore2.1",
+            "nodejs4.3-edge",
+            "go1.x"
```
kddejong: Need to update for new Lambda (ruby and custom)

src/cfnlint/data/ExtendedSpecs/append_spec/allowed_values.json
fatbasstard: Yes. Need to merge into the property Values one with the valuetypes

src/cfnlint/data/ExtendedSpecs/append_spec/allowed_values.json
fatbasstard: yeah

fatbasstard:
For now this is fine and keeps the general file structure intact. Should AWS change the file structure we can always change that later on.

src/cfnlint/data/ExtendedSpecs/all/03_value_types.json
```diff
@@ -0,0 +1,332 @@
...
+          "nodejs",
+          "nodejs4.3-edge",
+          "nodejs4.3",
+          "nodejs6.10",
+          "nodejs8.10",
+          "provided",
+          "python2.7",
+          "python3.6",
+          "python3.7",
+          "ruby2.5"
```
kddejong: You are still missing `provided`.  https://docs.aws.amazon.com/lambda/latest/dg/API_CreateFunction.html#SSS-CreateFunction-request-Runtime

cmmeyer:
I went ahead and merged in the new regions. So now we're just missing "provided" for Lambda.

PR #481, opened 2018-11-21
Update package dependencies

### LINKED ISSUE #471: Can't run in Travis CI xenial
I was testing the new Travis CI image:

https://blog.travis-ci.com/2018-11-08-xenial-release
https://docs.travis-ci.com/user/reference/xenial/

My install script ran `pip install --user --upgrade cfn-lint` and then `cfn-lint my.cfn.yaml` which gave this error:

```Traceback (most recent call last):
  File "/home/travis/.local/bin/cfn-lint", line 7, in <module>
    from cfnlint.__main__ import main
  File "/home/travis/.local/lib/python2.7/site-packages/cfnlint/__main__.py", line 19, in <module>
    import cfnlint.core
  File "/home/travis/.local/lib/python2.7/site-packages/cfnlint/core.py", line 24, in <module>
    import cfnlint.maintenance
  File "/home/travis/.local/lib/python2.7/site-packages/cfnlint/maintenance.py", line 19, in <module>
    import requests
  File "/usr/lib/python2.7/dist-packages/requests/__init__.py", line 58, in <module>
    from . import utils
  File "/usr/lib/python2.7/dist-packages/requests/utils.py", line 25, in <module>
    from . import certs
ImportError: cannot import name certs
```

I "fixed" this problem by just changing to Python3 in my build script:

```
pyenv global 3.7
pip3 install --user --upgrade cfn-lint
cfn-lint stack.cfn.yaml
```

But there might be some other lingering issue here

### LINKED ISSUE #479: Don't depend on pathlib2 on newer Python versions
*cfn-lint version: (`cfn-lint --version`)* 0.9.2

*Description of issue.* New dependency `pathlib2` probably isn't necessary from Python 3.4+ as it's just replicating `pathlib` that's in the standard library.

Another project using `pathlib2` is `pytest`, which conditionally depends on it for python versions <3.6, see: https://github.com/pytest-dev/pytest/blob/master/setup.py#L16 . `pathlib` was actually added to Python in 3.4 but they use features that only appear in Python 3.6+ (`__fspath__`).

cfn-lint should use a conditional dependency too to be a good citizen not requiring extra packages.

kddejong:
*Issue #, if available:*
Fix #479
Fix #471 

*Description of changes:*
- Only require pathlib2 in versions of python before 3.6
- Update dependencies on sam translator to 1.8 and fix six dependency to match SAM
- Update requests to fix when really old versions of requests are trying to be used.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -53,11 +53,13 @@ setup(
     packages=find_packages('src'),
     zip_safe=False,
     install_requires=[
-        'pyyaml', 'six',
+        'pyyaml',
+        'six~=1.11',
         'requests>=2.15.0',
-        'aws-sam-translator>=1.6.0',
-        'jsonpatch', 'jsonschema~=2.6.0',
-        'pathlib2>=2.3.0;python_version<"3.6"'
+        'aws-sam-translator>=1.8.0',
+        'jsonpatch',
+        'jsonschema~=2.6',
+        'pathlib2>=2.3.0;python_version<"3.4"'
     ],
     python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*',
     entry_points={
--- a/setup.py
+++ b/setup.py
@@ -52,7 +52,13 @@ setup(
     ]},
     packages=find_packages('src'),
     zip_safe=False,
-    install_requires=['pyyaml', 'six', 'requests', 'aws-sam-translator>=1.6.0', 'jsonpatch', 'jsonschema~=2.6.0', 'pathlib2'],
+    install_requires=[
+        'pyyaml', 'six',
+        'requests>=2.15.0',
+        'aws-sam-translator>=1.6.0',
+        'jsonpatch', 'jsonschema~=2.6.0',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -52,7 +52,14 @@ def get_version(filename):
     ]},
     packages=find_packages('src'),
     zip_safe=False,
-    install_requires=['pyyaml', 'six', 'requests', 'aws-sam-translator>=1.6.0', 'jsonpatch', 'jsonschema~=2.6.0', 'pathlib2'],
+    install_requires=[
+        'pyyaml',
+        'six~=1.11',
+        'requests>=2.15.0',
+        'aws-sam-translator>=1.8.0',
+        'jsonpatch', 'jsonschema~=2.6',
```
adamchainz: one per line missed here?

setup.py
```diff
@@ -52,7 +52,14 @@ def get_version(filename):
...
+        'pathlib2>=2.3.0;python_version<"3.6"'
```
adamchainz: you can probably do `<"3.4"` unless you depend on `__fspath__` or similar like pytest does

PR #492, opened 2018-11-28
Check for lists in Outputs

fatbasstard:
*Issue https://github.com/awslabs/cfn-python-lint/issues/485, if available:*

Add an additional check for lists in outputs. Lists are already checked if they originate from a `!Ref` or `Fn:GetAtt`. Added an explicit check on lists in the values. It's a separate loop/check in existing Rule [E6003](https://github.com/awslabs/cfn-python-lint/blob/master/docs/rules.md#E6003). 


fatbasstard:
That's already in the `myBadGetAttArrayOutput` resouce (assumption here based on the resource name)

PR #494, opened 2018-11-29
Support stdin for processing a template

fatbasstard:
*Issue https://github.com/awslabs/cfn-python-lint/issues/414, if available:*

Allow the processing of a template passed through stdin. while loading the yaml, based on `sys.stdin.isatty()` (Is there data on the stdin) the content is loaded from stdin.

Altered the Template Size rule ('E1002') to check if the file exists. The "original" file content is not availablle anymore after it's loaded as a `dict`. Could be alternative ways to keep track of the file size, but went for this approach to keep it as simple as possible.

*Usage:*

```
cat test/fixtures/templates/bad/generic.yaml | cfn-lint -
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,6 @@ Multiple files can be linted by either specifying multiple specific files:
 - `cfn-lint template1.yaml template2.yaml`
 - `cfn-lint -t template1.yaml template2.yaml`
 
-
 Multiple files can also be specified using wildcards (globbing):
 
 Lint all `yaml` files in `path`:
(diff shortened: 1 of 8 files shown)


kddejong:
I hate to ask but how hard would it be to add JSON support to this too?

fatbasstard:
I parsed some of the quickstart templates and they all got handled by the yaml loader :S 

If you can provide me a JSON template the goes through the JSON loader, I'll add it

kddejong:
https://github.com/awslabs/cfn-python-lint/blob/master/test/fixtures/templates/good/core/config_cfn_lint.json  Any tabbed json file should work.  The test templates we have with json extension should all be tabbed.

fatbasstard:
Check. On it!

PR #507, opened 2018-12-05
Clean up file open errors in py3x and only with testing

kddejong:
*Issue #, if available:*

*Description of changes:*
- Fix tests to not print the following error in Py3x
```
/home/travis/build/awslabs/cfn-python-lint/test/module/cfn_json/test_cfn_json.py:79: ResourceWarning: unclosed file <_io.TextIOWrapper name='fixtures/templates/quickstart/nat-instance.json' mode='r' encoding='UTF-8'>
```

### DIFF
--- a/test/module/cfn_json/test_cfn_json.py
+++ b/test/module/cfn_json/test_cfn_json.py
@@ -15,6 +15,8 @@
   SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 """
 import sys
+from mock import patch
+from six import StringIO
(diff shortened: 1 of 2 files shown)


cmmeyer:
Hey, so 

1. Commit says py27 and PR says py3?
2. This is indistinguishable from magic for me. I could probably google it, but Is this a typical pattern? Why mock?

kddejong:
commit message up to date.

Mock/patch is a way to substitute in something that would be external to what we are coding.  In this case we are faking the call to sys.stdin to provide a StringIO feed instead.

cmmeyer:
Got it. I have no idea how I missed this was in test. Sorry.

PR #515, opened 2018-12-14
Add support for SSM Strings in Cidr check

### LINKED ISSUE #513: Handle SSM parameters
The check E2505 fails when reading the CIDR from SSM using the following code:

```
  VpcCidr:
    Description: 'VPC CIDR Block'
    Type: AWS::SSM::Parameter::Value<String>
    Default: /Infra/VpcCidr
```

The message is misleading, too. It is about security groups and not cidr ranges.

https://github.com/awslabs/cfn-python-lint/blob/3b785df6b262b733bedcaa8c8de9a2c65ac161b8/src/cfnlint/rules/resources/ectwo/Vpc.py#L92

kddejong:
*Issue #, if available:*
Fix #513 
*Description of changes:*
- Fix rule E2505 to allow for SSM parameters

### DIFF
--- a/src/cfnlint/rules/resources/ectwo/Vpc.py
+++ b/src/cfnlint/rules/resources/ectwo/Vpc.py
@@ -54,7 +54,7 @@ class Vpc(CloudFormationLintRule):
             parameter_type = parameter.get('Type', None)
             if parameter_type not in allowed_types:
                 path_error = ['Parameters', value, 'Type']
-                message = 'CidrBlock parameter should be of type [{0}] for {1}'
+                message = 'DefaultTenancy parameter should be of type [{0}] for {1}'
                 matches.append(
                     RuleMatch(
                         path_error,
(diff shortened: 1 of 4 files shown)


src/cfnlint/rules/resources/ectwo/Vpc.py
```diff
@@ -42,7 +43,8 @@ def check_vpc_ref(self, value, path, parameters, resources):
         """Check ref for VPC"""
         matches = []
         allowed_types = [
-            'String'
+            'String',
+            'AWS::SSM::Parameter::Value<String>',
```
cmmeyer: Should we include a test for AWS::SSM::Parameter?

src/cfnlint/rules/resources/ectwo/Vpc.py
```diff
@@ -52,7 +54,7 @@ def check_vpc_ref(self, value, path, parameters, resources):
             parameter_type = parameter.get('Type', None)
             if parameter_type not in allowed_types:
                 path_error = ['Parameters', value, 'Type']
-                message = 'Security Group Id Parameter should be of type [{0}] for {1}'
+                message = 'CidrBlock parameter should be of type [{0}] for {1}'
```
cmmeyer: So was this a typo previously?

PR #523, opened 2018-12-20
Use Conditions to test if a related resource is available

kddejong:
*Issue #, if available:*

*Description of changes:*
- Add new rule W1001 to check if a related resource (ref, getatt) is available when that resource has a condition applied to it
- Add starting methodology to process conditions and analyze them for being related


src/cfnlint/__init__.py
```diff
@@ -18,16 +18,19 @@
...
 import six
 from yaml.parser import ParserError
 import cfnlint.helpers
+import cfnlint.conditions
 from cfnlint.transform import Transform
 from cfnlint.decode.node import TemplateAttributeError
 
 LOGGER = logging.getLogger(__name__)
 
+# pylint: disable=too-many-lines
```
fatbasstard: Perhaps just disable this linter rule for the entire project?

kddejong: yea possibly.  Was kind of debating if we should start splitting out some of those larger classes into their own files.  Then we could remove this one.

src/cfnlint/__init__.py
```diff
@@ -873,6 +877,131 @@ def check_value(self, obj, key, path,
...
+        results = []
+        path_conditions = self.get_conditions_from_path(self.template, path)
+        resource_condition = self.template.get('Resources', {}).get(resource, {}).get('Condition')
+        if resource_condition:
+            # resource conditions are always true.  If the same resource condition exists in the path
+            # with the same True value then nothing else matters
+            test_path_conditions = copy(path_conditions)
+            if not test_path_conditions.get(resource_condition):
+                test_path_conditions[resource_condition] = set()
+            if not test_path_conditions.get(resource_condition):
```
rjlohan: Dupe of L892, and L893 should set this if not already present in L892?

src/cfnlint/conditions.py
```diff
@@ -0,0 +1,316 @@
...
+from copy import copy
+import json
+import logging
+import six
+
+LOGGER = logging.getLogger(__name__)
+
+
+def get_hash(obj):
+    """ Return a hasl of an object """
```
rjlohan: Typo in comment?

src/cfnlint/__init__.py
```diff
@@ -873,6 +877,131 @@ def check_value(self, obj, key, path,
...
+        """
+            Compares a path to resource to see if its available
+            Returns scenarios that may result in the resource doesn't exist
+        """
+        results = []
+        path_conditions = self.get_conditions_from_path(self.template, path)
+        resource_condition = self.template.get('Resources', {}).get(resource, {}).get('Condition')
+        if resource_condition:
+            # resource conditions are always true.  If the same resource condition exists in the path
+            # with the same True value then nothing else matters
```
rjlohan: Nothing Else Matters

src/cfnlint/conditions.py
```diff
@@ -0,0 +1,316 @@
...
+        return None
+
+
+class Condition(object):
+    """ Individual condition """
+    And = None
+    Or = None
+    Not = None
+    Equals = None
+    Influenced_Equals = None
```
rjlohan: I don't know what this means. Can we explain the concept of `Influenced_Equals`?

src/cfnlint/conditions.py
```diff
@@ -0,0 +1,316 @@
...
+                self.Influenced_Equals[equal.Right.Function].add(equal.Left.Function)
+            elif equal.Left.String:
+                self.Influenced_Equals[equal.Right.Function].add(equal.Left.String)
+
+    def process_condition(self, template, value):
+        """ process condition """
+        if isinstance(value, dict):
+            if len(value) == 1:
+                for func_name, func_value in value.items():
+                    if func_name == 'Fn::And':
```
rjlohan: To avoid typos, can we pull out the template tokens like `Fn::And` to CONST type values somewhere?

kddejong: Fair.  I can agree with that.

src/cfnlint/conditions.py
kddejong: Basically all conditions eventually get down to equals.  So Influenced Equals includes all the Equals that make up a Condition.  So if there is sub condition from an and/or/not then we include its equals in the influenced equals.  That way can compare the base components of a condition... for instance all these are all related back to Ref myEnvironment either directly or indirectly.  
```
  isProduction: !Equals [!Ref 'myEnvironment', 'Prod']
  isNotProduction: !Not [{'Condition': 'isProduction'}]
  isProductionOrStaging: !Or [{'Condition': 'isProduction'}, {'Fn::Equals': [!Ref 'myEnvironment', 'Stage']}]
```

src/cfnlint/conditions.py
kddejong: yep will be fixed.

src/cfnlint/__init__.py
kddejong: agreed.  Will be fixed.

PR #536, opened 2018-12-31
 Implement Allowed Values rule

fatbasstard:
*Issue https://github.com/awslabs/cfn-python-lint/issues/50*

Introduce new rule `E3030` to check allowed values as mentioned in the documentation. Properties sometimes have a list of "allowed values" (e.g. https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-vpc.html#cfn-aws-ec2-vpc-instancetenancy).

Allowed Values can now be specified using the new spec patched and ValueTypes. A few of these values are already part of the Spec files (https://github.com/awslabs/cfn-python-lint/blob/master/src/cfnlint/data/ExtendedSpecs/all/03_value_types.json#L145)

This PR will implements the rule that checks these values. Currently 2 types are covered by the rule: Values and Parameters (Ref's to Parameters with Allowed Values)

This replaces the 2 existing hardcoded rules to check for the Lambda Runtimes


kddejong:
Cool.  Thanks @fatbasstard should have time to review this today but first pass looks great.

PR #537, opened 2018-12-31
Improve is_custom_resource check

fatbasstard:
Rule `E3008` breaks if there is a Reference to a non-existing resource:

```
(venv) ➜  cfn-python-lint git:(quick-fixes) cfn-lint test/fixtures/templates/bad/resources_lambda.yaml 
E0002 Unknown exception while processing rule E3008: 'NoneType' object has no attribute 'startswith'
test/fixtures/templates/bad/resources_lambda.yaml:1:1
```

Discovered this when working on another rule. The `is_custom_resource` check now handles `NoneType` (None = Not a custom resource)

### DIFF
--- a/src/cfnlint/helpers.py
+++ b/src/cfnlint/helpers.py
@@ -182,7 +182,7 @@ def set_specs(override_spec_data):
 
 def is_custom_resource(resource_type):
     """ Return True if resource_type is a custom resource """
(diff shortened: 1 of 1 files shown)


kddejong:
None isn't a custom resource so this works as I would expect.

PR #566, opened 2019-01-08
Don't fail on regex that isn't supported by Python

kddejong:
*Issue #, if available:*

*Description of changes:*
- Fix rule E2015 to not fail on an AllowedPattern when the regex isn't supported by Python

### DIFF
--- a/src/cfnlint/rules/parameters/Default.py
+++ b/src/cfnlint/rules/parameters/Default.py
@@ -33,8 +33,11 @@ class Default(CloudFormationLintRule):
             Check allowed value against allowed pattern
         """
         message = 'Default should be allowed by AllowedPattern'
-        if not re.match(allowed_pattern, str(allowed_value)):
-            return([RuleMatch(path, message)])
+        try:
+            if not re.match(allowed_pattern, str(allowed_value)):
+                return([RuleMatch(path, message)])
+        except re.error as ex:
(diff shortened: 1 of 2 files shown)


src/cfnlint/rules/parameters/Default.py
```diff
@@ -33,8 +33,11 @@ def check_allowed_pattern(self, allowed_value, allowed_pattern, path):
             Check allowed value against allowed pattern
         """
         message = 'Default should be allowed by AllowedPattern'
-        if not re.match(allowed_pattern, str(allowed_value)):
-            return([RuleMatch(path, message)])
+        try:
+            if not re.match(allowed_pattern, str(allowed_value)):
+                return([RuleMatch(path, message)])
+        except Exception:   # pylint: disable=W0703
```
fatbasstard: Can we use a more explicit Exception? And perhaps also output the error message in the logline?

kddejong: I think the ways it can fail result in different errors.  We found one error with the `multiple repeat` error.  We can capture them all as we run into them or all Re errors, maybe?

kddejong: Updated with something more useful... hopefully

fatbasstard:
This looks awesome this way! :Sunglasses: :+1:

PR #570, opened 2019-01-09
Feature Add: Scaffolding for arbitrary Match attributes, adding attributes for Type checks.

agamzn:
This PR adds support for arbitrary attributes in the Match objects, allowing for per-type data to be bubbled upwards. This also adds actual_type and expected_type attributes to the PrimitiveType rule (E3012). The scaffolding is a remediation starting point, while the extra attributes allow downstream consumers to perform any type conversions they wish in a programmatic matter.


src/cfnlint/rules/resources/properties/ValuePrimitiveType.py
```diff
@@ -55,30 +55,35 @@ def check_primitive_type(self, value, item_type, path):
...
+                matches.append(RuleMatch(path, message, **extra_args))
         elif item_type in ['Long']:
             if sys.version_info < (3,):
                 integer_types = (int, long,)  # pylint: disable=undefined-variable
             else:
                 integer_types = (int,)
             if not isinstance(value, integer_types):
                 message = 'Property %s should be of type Long' % ('/'.join(map(str, path)))
-                matches.append(RuleMatch(path, message))
+                matches.append(RuleMatch(path, message, **extra_args))
```
kddejong: You're missing setting extra_args in this one.  That should get your tests to green.

src/cfnlint/__init__.py
```diff
@@ -346,10 +346,13 @@ def create_from_directory(cls, rulesdir):
 class RuleMatch(object):
     """Rules Error"""
 
-    def __init__(self, path, message):
+    def __init__(self, path, message, **kwargs):
         """Init"""
         self.path = path
+        self.path_string = '/'.join(map(str, path))
```
kddejong: Whats the benefit to this one?

agamzn: No idea. I didn't add it.

agamzn:
```
 ~/development/qstools/amiupdater-sandbox     master    cfn-lint linux-bastion-old.template
E3012 Property Resources/SSHMetricFilter/Properties/MetricTransformations/0/MetricValue should be of type String
linux-bastion-old.template:402:25

E3012 Property Resources/BastionAutoScalingGroup/Properties/Tags/0/PropagateAtLaunch should be of type Boolean
linux-bastion-old.template:589:25

E3012 Property Resources/BastionSecurityGroup/Properties/SecurityGroupIngress/0/FromPort should be of type Integer
linux-bastion-old.template:781:25

E3012 Property Resources/BastionSecurityGroup/Properties/SecurityGroupIngress/0/ToPort should be of type Integer
linux-bastion-old.template:782:25

E3012 Property Resources/BastionSecurityGroup/Properties/SecurityGroupIngress/1/FromPort should be of type Integer
linux-bastion-old.template:789:25

E3012 Property Resources/BastionSecurityGroup/Properties/SecurityGroupIngress/1/ToPort should be of type Integer
linux-bastion-old.template:790:25

```
```
>>> import cfnlint.core
>>> rules = cfnlint.core.get_rules([], [], [])
>>> def _parse_template(template_path, quiet=False):
...     try:
...         return cfnlint.decode.cfn_yaml.load(template_path)
...     except ScannerError as e:
...         if not quiet:
...             print(PrintMsg.ERROR + 'Linter failed to load template %s "%s" line %s, column %s' % (
...                   template_path, e.problem, e.problem_mark.line, e.problem_mark.column))
...     except FileNotFoundError as e:
...         if not quiet:
...             print(PrintMsg.ERROR + 'Linter failed to load template %s "%s"' % (template_path, str(e)))
...
...
...
...
>>> t = 'example'
>>> template = _parse_template('linux-bastion-old.template')
>>> results = cfnlint.core.run_checks(t, template, rules, ['us-east-1'])
>>> results[0]
[E3012: Check resource properties values] (Property Resources/SSHMetricFilter/Properties/MetricTransformations/0/MetricValue should be of type String) matched 
example:402

>>> vars(results[0])
{'linenumber': 402, 'columnnumber': 25, 'linenumberend': 402, 'columnnumberend': 38, 'filename': 'example', 'rule': E3012: Check resource properties values, 'message': 'Property Resources/SSHMetricFilter/Properties/MetricTransformations/0
/MetricValue should be of type String'}

>>> for r in results:
...     print(r.linenumber)
...
...
402
589
781
782
789
790
>>>
```

From there, the end-user could write a function that took the input path, expected type, actual type.. 

```
for r in results:
    convert_type(r.path, r.expected_type, r.actual_type)
```

I haven't flushed out the necessary logic yet, but having the attributes there is a huge step in the right direction.

src/cfnlint/__init__.py
agamzn: OH, yes I did add it. If... i remember correctly... the path is a list... and turning it into a string means it can be added to an attribute.

kddejong:
Going to merge this.  Going to consider this beta for a little bit until we can build more around how it will be used.

PR #571, opened 2019-01-10
New rule to check the keys in a FindInMap

kddejong:
*Issue #, if available:*

*Description of changes:*
- Add rule E1013 to check the keys in a FindInMap function to see if they are valid
- Rewrite rule E1011 to clean it up


src/cfnlint/rules/functions/FindInMap.py
```diff
@@ -59,6 +59,62 @@ def check_dict(self, obj, tree):
...
+            message = 'FindInMap first key should be a {0}, string, or int at {1}'
+            matches.append(RuleMatch(
+                tree[:] + [1],
+                message.format(
+                    ', '.join(map(str, self.supported_functions)),
+                    '/'.join(map(str, tree)))))
+
+        return matches
+
+    def second_key(self, second_key, tree):
```
cmmeyer: Not a huge deal since CFN only supports two levels, but we could make these identical functions a unified “map_key()” func — we’d lose the “first/second” in the message text though, and I seriously doubt CFN will ever add more levels to the map

src/cfnlint/rules/functions/FindInMapKeys.py
```diff
@@ -0,0 +1,90 @@
...
+"""
+import six
+from cfnlint import CloudFormationLintRule
+from cfnlint import RuleMatch
+
+
+class FindInMapKeys(CloudFormationLintRule):
+    """Check if FindInMap values are correct"""
+    id = 'E1013'
+    shortdesc = 'FindInMap key\'s exist in the map'
```
cmmeyer: Don’t need the \’ here.

src/cfnlint/rules/functions/FindInMapKeys.py
```diff
@@ -0,0 +1,90 @@
...
+                mapping = mappings.get(map_name)
+                if mapping:
+                    if isinstance(first_key, six.string_types):
+                        if mapping.get(first_key):
+                            # Don't double error if they first key doesn't exist
+                            if not mapping.get(first_key, {}).get(second_key):
+                                message = 'FindInMap second key "{0}" doesn\'t exist in map "{1}" under "{2}" at {3}'
+                                matches.append(RuleMatch(
+                                    tree[:] + [2],
+                                    message.format(second_key, map_name, first_key, '/'.join(map(str, tree)))))
```
cmmeyer: Since you have to validate the first key again here anyway, does it make sense to just pass both keys and validate them together in single function?

src/cfnlint/rules/functions/FindInMapKeys.py
kddejong: I need to escape the single quote.  You sure?

src/cfnlint/rules/functions/FindInMapKeys.py
kddejong: Good question.  I could do that I suppose.  I was just trying to break it up for readability and to keep the functions small.

src/cfnlint/rules/functions/FindInMap.py
kddejong: I was thinking about it.  If I combine it into one function I would just pass in an array of the keys.

src/cfnlint/rules/functions/FindInMapKeys.py
kddejong: doh. Fixing.

src/cfnlint/rules/functions/FindInMap.py
```diff
@@ -59,6 +59,62 @@ def check_dict(self, obj, tree):
...
+        """ Check the validity of the first key """
```
cmmeyer: Should be "validity of the second key"

PR #579, opened 2019-01-11
Added configuration/values to OnlyOne and Exclusive

fatbasstard:
While working on the allowed values and uploading a "valid" template came across these errors: 
image

Add 2 errors to the "OnlyOne" rule, see the documentation for more info: 
* https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-route53-recordsetgroup.html#cfn-route53-recordsetgroup-hostedzoneid
* https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-as-group.html#cfn-as-group-launchconfigurationname

Added another to the Exclusive, documentation:
* https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-websiteconfiguration.html#cfn-s3-websiteconfiguration-redirectallrequeststo

Also made some minor tweaks to the `allowed_values` test stacks to make the deployment a bit easier

### DIFF
--- a/src/cfnlint/data/AdditionalSpecs/Exclusive.json
+++ b/src/cfnlint/data/AdditionalSpecs/Exclusive.json
@@ -25,6 +25,13 @@
         "Ebs"
       ]
     },
+    "AWS::S3::Bucket.WebsiteConfiguration": {
+      "RedirectAllRequestsTo": [
+        "ErrorDocument",
+        "IndexDocument",
+        "RoutingRules"
+      ]
+    },
     "AWS::EC2::SecurityGroup.Ingress": {
       "CidrIp": [
         "SourceSecurityGroupName",
--- a/src/cfnlint/data/AdditionalSpecs/OnlyOne.json
+++ b/src/cfnlint/data/AdditionalSpecs/OnlyOne.json
@@ -38,5 +38,25 @@
     ]
   },
   "ResourceTypes": {
+    "AWS::AutoScaling::AutoScalingGroup": [
+      [
+        "InstanceId",
+        "LaunchConfigurationName",
+        "LaunchTemplate",
+        "MixedInstancesPolicy"
+      ]
+    ],
+    "AWS::Route53::RecordSet": [
+      [
+        "HostedZoneId",
+        "HostedZoneName"
+      ]
+    ],
+    "AWS::Route53::RecordSetGroup": [
+      [
+        "HostedZoneId",
+        "HostedZoneName"
+      ]
+    ]
   }
 }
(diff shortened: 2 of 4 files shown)


src/cfnlint/data/AdditionalSpecs/OnlyOne.json
```diff
@@ -38,5 +38,25 @@
     ]
   },
   "ResourceTypes": {
+    "AWS::AutoScaling::AutoScalingGroup": [
```
kddejong: Should this also be in at least one? 

`Required: Conditional. You must specify one of the following: InstanceId, LaunchConfigurationName, LaunchTemplate, or MixedInstancesPolicy.`

src/cfnlint/data/AdditionalSpecs/OnlyOne.json
```diff
@@ -38,5 +38,25 @@
...
+      [
+        "InstanceId",
+        "LaunchConfigurationName",
+        "LaunchTemplate",
+        "MixedInstancesPolicy"
+      ]
+    ],
+    "AWS::Route53::RecordSet": [
```
kddejong: Same.

`Required: Conditional. You must specify either the HostedZoneName or HostedZoneId, but you cannot specify both. If this record set is part of a record set group, do not specify this property.`

src/cfnlint/data/AdditionalSpecs/Exclusive.json
```diff
@@ -25,6 +25,13 @@
         "Ebs"
       ]
     },
+    "AWS::S3::Bucket.WebsiteConfiguration": {
```
kddejong: 👍

src/cfnlint/data/AdditionalSpecs/OnlyOne.json
fatbasstard: This is correct, right? "but you cannot specify both", that makes it "only one"
🤔

kddejong: Yea, I'm wondering if it should be in both files  AtLeastOne and OnlyOne

src/cfnlint/data/AdditionalSpecs/OnlyOne.json
fatbasstard: Well, just specified 2 values and got the following error:

```
Exactly one of [LaunchConfigurationName,InstanceId,LaunchTemplate,MixedInstancesPolicy] needs to be specified
```

src/cfnlint/data/AdditionalSpecs/OnlyOne.json
fatbasstard: But this is then correct right? "You must specify either the HostedZoneName or HostedZoneId, but you cannot specify both"

kddejong: yes this is also correct (from what I can tell).

PR #586, opened 2019-01-15
Parse pricing API to get allowed values

fatbasstard:
Took over the commit from @kddejong to make changes and improve the allowedValues rule using this: The first "region specific" allowedValues 🎉 

Current status/changes:
- API loads Ec2, AmazonMQ and RDS Instance types and created region specific valueType configurations (™️  @kddejong)
- Adeed the new spec patches to the setup.py so they get packaged
- Attached the Valuetypes to a bunch of resources and added them to the allowed_values tests
- Changed the initialization flow of the rules. Instead of always using `us-east-1`, the first region is used (by default that's us-east-1) so the specified region is actually used.
- Added Redshift instance types
- Added some more logging to the script

### DIFF
--- a/src/cfnlint/__init__.py
+++ b/src/cfnlint/__init__.py
@@ -19,15 +19,18 @@ import sys
 import os
 import re
 import traceback
-from copy import deepcopy
+from copy import deepcopy, copy
 from datetime import datetime
 import six
 from yaml.parser import ParserError
 import cfnlint.helpers
+import cfnlint.conditions
 from cfnlint.transform import Transform
 from cfnlint.decode.node import TemplateAttributeError
 
 LOGGER = logging.getLogger(__name__)
+
+# pylint: disable=too-many-lines
 
 
 class CloudFormationLintRule(object):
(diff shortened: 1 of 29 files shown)


src/cfnlint/rules/parameters/AllowedValue.py
```diff
@@ -28,12 +28,11 @@ class AllowedValue(CloudFormationLintRule):
     source_url = 'https://github.com/awslabs/cfn-python-lint/blob/master/docs/cfn-resource-specification.md#allowedvalue'
     tags = ['resources', 'property', 'allowed value']
 
-    def __init__(self, ):
-        """Init """
+    def initialize(self, cfn):
         super(AllowedValue, self).__init__()
-        for resource_type_spec in RESOURCE_SPECS.get('us-east-1').get('ResourceTypes'):
+        for resource_type_spec in RESOURCE_SPECS.get('eu-west-1').get('ResourceTypes'):
```
kddejong: you mean `cfn.regions[0]`?

fatbasstard: Yup, I do;... Changes.

I will fix this in a more "structural" manner using helper functions

PR #624, opened 2019-01-30
Add new rule to block Ref's from Parameters

fatbasstard:
The Default value of a Parameter has to be a string:

image

Same check using the ValidateTemplate:
```
aws cloudformation validate-template --template-body '---                 
Parameters:
  MyEmptyDefault:
    Type: "Number"
    Default: !Ref "AWS::NoValue"
    AllowedValues:
      - "a"
      - "b"
Resources: {}
' 

An error occurred (ValidationError) when calling the ValidateTemplate operation: Template format error: Every Default member must be a string.
```

Python implicitly converts all primitive types to strings, but making that strict seems a bit harsh (That will hurt a lot of people 😱 ).

Complex types like `!Ref` is already checked by other rules like `E1012`, but these exclude the psuedo parameter `AWS:NoValue` (for valid reasons).

To prevent duplicate rules triggering on the same line part of the template, add an explicit check on a `!Ref "AWS:NoValue"` to the Default check.


fatbasstard:
I'll update the PR... Thanx for the feedback 👍

PR #625, opened 2019-01-30
Use a generic regex rule to validate property values

fatbasstard:
A lot of (string) property values are bound to a Pattern. These Patterns (regex) are not always documented, but [sometimes they are](https://github.com/awsdocs/aws-cloudformation-user-guide/search?q=pattern%3A&unscoped_q=pattern%3A)

Based on the "standard" rule approach from the allowed value, I'm looking if the same could be applicable to Patterns.

This PR adds a new rule (`E3031`) that checks Patterns. 
Added the Pattern of the SecurityGroup Description, that is currently hardcoded in rule `E2509`.  This created the following output:

```
(venv) ➜  cfn-python-lint git:(value-regex) ✗ cfn-lint test/fixtures/templates/bad/properties_sg_description.yaml
E2509 GroupDescription length (282) exceeds the limit (255) at Resources/mySecurityGroupLongDescription/Properties/GroupDescription
test/fixtures/templates/bad/properties_sg_description.yaml:11:7

E2509 GroupDescription contains invalid characters (valid characters are: "a-zA-Z0-9. _-:/()#,@[]+=&;\{}!$*"") at Resources/mySecurityGroupInvalidCharacters/Properties/GroupDescription
test/fixtures/templates/bad/properties_sg_description.yaml:15:7

E3031 GroupDescription contains invalid characters (Pattern: ^([a-z,A-Z,0-9,. _\-:/()#,@[\]+=&;\{\}!$*])*$) at Resources/mySecurityGroupInvalidCharacters/Properties/GroupDescription
test/fixtures/templates/bad/properties_sg_description.yaml:15:7

E2509 GroupDescription contains invalid characters (valid characters are: "a-zA-Z0-9. _-:/()#,@[]+=&;\{}!$*"") at Resources/mySecurityGroupValidSub/Properties/GroupDescription/Fn::Sub
test/fixtures/templates/bad/properties_sg_description.yaml:27:7

E2509 GroupDescription contains invalid characters (valid characters are: "a-zA-Z0-9. _-:/()#,@[]+=&;\{}!$*"") at Resources/mySecurityGroupBadMapping/Properties/GroupDescription/Fn::Sub/0
test/fixtures/templates/bad/properties_sg_description.yaml:37:7
```

**Features:**
- New rule to test values against a RegEx
- Support for a "human friendly" RegEx value to display in the error
- Support for strings with `!Sub` (The variable is "just removed"). Maybe not a "catcall", but a good 80%

**Done**
- Added the regex for SG Descriptions and SG Cidr IP's
- Remove the existing SG Description rule without loss of functionality (Length has to be added to the RegEx...)
- Added a test that checks if all the patterns are "valid" patterns

### DIFF
--- a/src/cfnlint/data/ExtendedSpecs/all/04_property_values.json
+++ b/src/cfnlint/data/ExtendedSpecs/all/04_property_values.json
@@ -600,6 +600,20 @@
   },
   {
     "op": "add",
+    "path": "/PropertyTypes/AWS::EC2::SecurityGroup.Egress/Properties/CidrIp/Value",
+    "value": {
+      "ValueType": "CidrIp"
+    }
+  },
+  {
+    "op": "add",
+    "path": "/PropertyTypes/AWS::EC2::SecurityGroup.Ingress/Properties/CidrIp/Value",
+    "value": {
+      "ValueType": "CidrIp"
+    }
+  },
+  {
+    "op": "add",
     "path": "/PropertyTypes/AWS::EC2::SpotFleet.IamInstanceProfileSpecification/Properties/Arn/Value",
     "value": {
       "ValueType": "IamInstanceProfile.Arn"
@@ -1483,6 +1497,27 @@
     "path": "/ResourceTypes/AWS::EC2::SecurityGroup/Properties/VpcId/Value",
     "value": {
       "ValueType": "VpcId"
+    }
+  },
+  {
+    "op": "add",
+    "path": "/ResourceTypes/AWS::EC2::SecurityGroup/Properties/GroupDescription/Value",
+    "value": {
+      "ValueType": "SecurityGroupDescription"
+    }
+  },
+  {
+    "op": "add",
+    "path": "/ResourceTypes/AWS::EC2::SecurityGroupEgress/Properties/CidrIp/Value",
+    "value": {
+      "ValueType": "CidrIp"
+    }
+  },
+  {
+    "op": "add",
+    "path": "/ResourceTypes/AWS::EC2::SecurityGroupIngress/Properties/CidrIp/Value",
+    "value": {
+      "ValueType": "CidrIp"
     }
   },
   {
--- a/src/cfnlint/data/CloudSpecs/ap-northeast-1.json
+++ b/src/cfnlint/data/CloudSpecs/ap-northeast-1.json
@@ -6970,7 +6970,10 @@
           "Documentation": "http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-security-group-rule.html#cfn-ec2-security-group-rule-cidrip",
           "PrimitiveType": "String",
           "Required": false,
-          "UpdateType": "Mutable"
+          "UpdateType": "Mutable",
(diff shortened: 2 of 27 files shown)


src/cfnlint/data/ExtendedSpecs/all/04_property_values.json
```diff
@@ -1223,6 +1223,13 @@
       "ValueType": "VpcId"
     }
   },
+  {
+    "op": "add",
+    "path": "/ResourceTypes/AWS::EC2::SecurityGroup/Properties/GroupDescription/Value",
+    "value": {
+      "Pattern": "^([a-z,A-Z,0-9,. _\\-:/()#,@[\\]+=&;\\{\\}!$*])*$"
```
kddejong: My only real debate here is if should keep this to a ValueType.  Do you think there is a chance that certain things will always follow a pattern that we could reuse across many different things?  For instance with Security Groups we could put the pattern with the Security Group of `sg-someregex` then as we associate to that we can regex the values if they are provided.    In this case it is probably pretty particular to the Group Description attribute... not sure if we would run across this regex again.

fatbasstard: Added in another example that is interesting for this discussion, see the new commit

kddejong: I like having the regex under the Valut Type fields to reduce repeating.  What do you think?

fatbasstard: Well, at first I thought: "overkill". Now that a SecurityGroup CidrIp is popping up 4 times already, I might just agree...

Makes it also possible to (idea) add a Human Readable error. Regex might not always be "user friendly"

fatbasstard: I'll update this PR soon with these changes

fatbasstard:
Rebuild the rule:

- Now uses the `ValueType` approach, similar to the `AllowedValues` (I'm thinking about a way to process the `ValueTypes` in 1 place. So the rules only need to implement the rule/handling. But that's for later)
- Added the feature to specify a "friendly" pattern message. Some Regex definitions are extremely complex, with this the expected value is presented a nicer way.

Example:
```
~ cfn-lint test/fixtures/templates/bad/properties_sg_ingress.yaml
E3031 CidrIp contains invalid characters (Pattern: x.x.x.x/y) at Resources/mySecurityGroupVpc/Properties/SecurityGroupIngress/2/CidrIp
test/fixtures/templates/bad/properties_sg_ingress.yaml:43:9
```

FYI: Still left out the (changed) spec files to keep the changes clear and prevent merging "issues" with keeping the PR open for a long time

fatbasstard:
@kddejong Need to finalize it, I'll remove the WIP tag when done.

Todo:
- replace current Security Group description rule/check
- Add in support for `!sub` (since that's part of the current rule, don't want to lose functionality
- Tweak/optimize current code

fatbasstard:
PR updated and I consider this a "deployable" now

PR #631, opened 2019-02-03
E3035 New rule for DeletionPolicy values

cmmeyer:
*Issue #, if available:*

*Description of changes:*

Add a new rule to validate DeletionPolicy values (Delete,Retain,Snapshot) and make sure Snapshot is used only for allowed resources.


cmmeyer:
The code for UpdateReplacePolicy would be almost identical -- I'm debating whether it makes more sense to have two separate rules or evaluate them together?

cmmeyer:
Removing WIP -- this is ready for review.

src/cfnlint/rules/resources/DeletionPolicy.py
```diff
@@ -0,0 +1,87 @@
...
+from cfnlint import RuleMatch
+
+
+class DeletionPolicy(CloudFormationLintRule):
+    """Check Base Resource Configuration"""
+    id = 'E3035'
+    shortdesc = 'Check DeletionPolicy values for Resources'
+    description = 'Check that the DeletionPolicy values are valid'
+    source_url = 'https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-deletionpolicy.html'
+    tags = ['resources', 'dependson']
```
kddejong: not that we use tags much (yet) is dependson correct here?

cmmeyer: Oh shoot. That's a holdover from the rule I stole.. er.. borrowed as a model. I'll clean that up.

PR #632, opened 2019-02-04
Add rule configuration parameters

### LINKED ISSUE #547: E3012 Property Type Checks are Too Strict
*cfn-lint version: 0.9.2*

There are many properties that demand a "String" type when they are integer in nature. Here are examples:

* `AWS::AutoScaling::AutoScalingGroup/Properties/DesiredCapacity`
* `AWS::AutoScaling::AutoScalingGroup/Properties/MaxSize`
* `AWS::AutoScaling::AutoScalingGroup/Properties/MinSize`
* `AWS::Route53::RecordSet/Properties/TTL`

This list is actually pretty huge, which is why I haven't wanted to create an override spec. Since these are all accepted as numbers by CloudFormation, we should be able to treat that as such.

kddejong:
*Issue #, if available:*
Fix #547 
*Description of changes:*
- Add a process to configure rules
- Update rule E3012 to allow for non strict property checking


src/cfnlint/config.py
```diff
@@ -221,6 +221,24 @@ def comma_separated_arg(string):
...
+    for config in configs:
+        try:
+            rule_id = config.split(':')[0]
+            config_name = config.split(':')[1].split('=')[0]
+            config_value = config.split(':')[1].split('=')[1]
+            if rule_id not in results:
+                results[rule_id] = {}
+            results[rule_id][config_name] = config_value
+        except Exception as e:
+            raise e
```
fatbasstard: this could/should be a better error message. It'a pretty generic try/catch around an expected format in a configuration

src/cfnlint/rules/resources/properties/ValuePrimitiveType.py
```diff
@@ -46,16 +61,35 @@ def initialize(self, cfn):
...
         matches = []
 
         if isinstance(value, dict) and item_type == 'Json':
             return matches
         if item_type in ['String']:
-            if not isinstance(value, (str, six.text_type, six.string_types)):
-                message = 'Property %s should be of type String' % ('/'.join(map(str, path)))
-                matches.append(RuleMatch(path, message))
+            if not isinstance(value, (six.string_types)):
+                if self.config['strict'] is False:
```
fatbasstard: Isn't `if not self.config['strict']` more "pythonic"? (Not sure here, just a question)

fatbasstard:
Sweet.... Don't forget to update the documentation with this. Pretty awesome "rules themself are now configurable" feature 👍

src/cfnlint/rules/resources/properties/ValuePrimitiveType.py
```diff
@@ -30,11 +30,26 @@ class ValuePrimitiveType(CloudFormationLintRule):
     source_url = 'https://github.com/awslabs/cfn-python-lint/blob/master/docs/cfn-resource-specification.md#valueprimitivetype'
     tags = ['resources']
 
+    def configure(self, configs):
+        """Set configuration """
+
+        # Set the defaults
+        self.config['strict'] = True
+
+        if isinstance(configs, dict):
```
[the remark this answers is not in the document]
kddejong: Seems like a decent idea.  Not sure how to do this yet though so open to suggestions.

PR #635, opened 2019-02-04
E3036 New rule for UpdateReplacePolicy values 

cmmeyer:
*Issue #, if available:*

*Description of changes:*
New rule to validate values for UpdateReplacePolicy properties on all resources.


test/rules/resources/test_deletionpolicy.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+class TestResourceDeletionPolicy(BaseRuleTestCase):
+    """Test base template"""
+    def setUp(self):
+        """Setup"""
+        super(TestResourceDeletionPolicy, self).setUp()
+        self.collection.register(DeletionPolicy())
+
+    def test_file_positive(self):
```
rjlohan: I think generally, these tests need to be more explicit. What failure case is expected here? There are several possible; 
* That the DeletionPolicy is an array and not a string? 
* That the DeletionPolicy value is not in the enum
* That `Snapshot` is applied to an invalid type.

Will catch more faults and make tests clearer if we target them a bit more explicitly.

cmmeyer: So ...

1. This is a "Chuck is a crap dev moment" -- the DeletionPolicy code from #631 ended up here as well.

2. I test the three explicit scenarios via the new bad template.  That is historically how we have tested rules. Are we proposing a new, non-template mechanism? Or that I should break the three bad cases into separate templates?

cmmeyer: Ok, I've narrowed this just the changes for UpdateReplacePolicy.  Now help me understand what improvements you're looking for with testing?

PR #654, opened 2019-02-11
Add Config Allowed Values

fatbasstard:
*Issue https://github.com/awslabs/cfn-python-lint/issues/50, if available:*

Add all the allowed values of the [`AWS::Config`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-reference-config.html) Resources

This one turned out to be a bit more complex, since the resource types supported in AWS config are dependant on the support of that service:
- https://docs.aws.amazon.com/config/latest/developerguide/resource-config-reference.html
- https://aws.amazon.com/about-aws/global-infrastructure/regional-product-services/

Changes made for this:
- Split the allowed values into per-region settings
- The `add` operation is slightly different, since an `add` operation on the same path overrides an other one.
- The `AWS Shield` resources are not separated yet due to the following: `The AWS::ShieldRegional::Protection is available in all regions where AWS Shield is supported.`. Shield Standard or Shield Advanced? 🤔  (But that one can be improved in a new PR is needed)


kddejong:
Yikes. Agreed. The per region resource types is kind of scary but makes sense why it was done this way.

PR #660, opened 2019-02-14
Add Linter rule for IAM Permissions

jvanbrunschot-coolblue:
*Description of changes:*

I noticed when developing this rule that it tripped on some policies; Some permissions might exist but are not in the policies.json file; (these permission also trip the *Policy Summary* in the AWS IAM Console

Some examples:
```
appstream:Get*
kinesis:StartStreamEncryption
cognito-idp:DeleteUserClient
apigateway:HEAD
apigateway:OPTIONS
```
image

This will add a new rule to check for  valid IAM permissions in the following resources:
```
'AWS::IAM::ManagedPolicy',
'AWS::IAM::Policy',
'AWS::SNS::TopicPolicy',
'AWS::SQS::QueuePolicy'
```

The Source is coming for the [AWS Policy Generator](https://awspolicygen.s3.amazonaws.com/policygen.html) 
https://awspolicygen.s3.amazonaws.com/js/policies.js


jvanbrunschot-coolblue:
I ran this rule against our codebase..

But test fail due some IAM stuff we don't use.

jvanbrunschot-coolblue:
Yes, it is just a formatted version fo that file. I modify the structure in the rule tho. 

Anyway, I can try to add a maintenance endpoint of script to make updating easier!

jvanbrunschot-coolblue:
@kddejong added the maintenance endpoint

kddejong:
@jvanbrunschot-coolblue and @cmmeyer I hate to keep patching things all over the place but any thoughts on doing this here?

From what I can tell only these from the list above are valid.
```
kinesis:StartStreamEncryption
apigateway:HEAD
apigateway:OPTIONS
```

src/cfnlint/rules/resources/iam/Permissions.py
```diff
@@ -0,0 +1,145 @@
...
+            if isinstance(effective_permissions.get('NotAction'), str):
+                actions.append(effective_permissions.get('NotAction'))
+            else:
+                actions.extend(effective_permissions.get('NotAction'))
+
+        return actions
+
+    def match(self, cfn):
+        """Check IAM Permissions"""
+        matches = list()
```
kddejong: this should be matches = []

test/fixtures/results/quickstart/nist_iam.json
```diff
@@ -78,5 +78,25 @@
...
+                "ColumnNumber": 9,
+                "LineNumber": 163
+            },
+            "End": {
+                "ColumnNumber": 9,
+                "LineNumber": 163
+            }
+        },
+        "Level": "Error",
+        "Message": "Invalid permission \"Get*\" for \"appstream\" found in permissions",
```
kddejong: I'm getting confused a little bit here.  Do errors in IAM policies result in an Error or a failure of the stack being deployed?  Should this be a warning?  I really can't find an appstream Get API call (which if I'm missing one my thought process is completely off so double check my work here).

jvanbrunschot-coolblue: Yes, a warning makes more sense. Invalid IAM policies will just get deployed without errors in CloudFormation. I can turn this into a warning if that makes more sense.

src/cfnlint/rules/resources/iam/Permissions.py
```diff
@@ -0,0 +1,145 @@
...
+
+        return matches
+
+    def get_actions(self, effective_permissions):
+        """return all actions from a statement"""
+
+        actions = []
+
+        if 'Action' in effective_permissions:
+            if isinstance(effective_permissions.get('Action'), str):
```
kddejong: Can we switch this to `six.string_types`?  It helps with different Python versions.

src/cfnlint/rules/resources/iam/Permissions.py
```diff
@@ -0,0 +1,145 @@
...
+                actions.append(effective_permissions.get('Action'))
+            else:
+                actions.extend(effective_permissions.get('Action'))
+
+        if 'NotAction' in effective_permissions:
+            if isinstance(effective_permissions.get('NotAction'), str):
```
kddejong: Can we switch this to `six.string_types`?  It helps with different Python versions.

src/cfnlint/rules/resources/iam/Permissions.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def get_actions(self, effective_permissions):
+        """return all actions from a statement"""
+
+        actions = []
+
+        if 'Action' in effective_permissions:
+            if isinstance(effective_permissions.get('Action'), str):
+                actions.append(effective_permissions.get('Action'))
+            else:
```
kddejong: This may be a weird one but planning for bad templates is hard to do.  Should we be making sure that this is actually a list before extending it as a list?

kddejong:
Thank you so much for work on this.  I'm excited to get it out there.

PR #674, opened 2019-02-26
Use the cfn-lint YAML decoders and force the Path to be a string

### LINKED ISSUE #673: AttributeError: 'PosixPath' object has no attribute 'read_text'
*cfn-lint version:*
0.9.x & 0.11.x

*Description of issue.*
When running `cfn-lint -c I -t <path to template.yml> -a <path to rules>` with a .cfnlintrc file, I receive the error from here:
https://github.com/awslabs/cfn-python-lint/blob/master/src/cfnlint/config.py#L211

*Please provide as much information as possible:*
When running `cfn-lint -c I -t <path to template.yml> -a <path to rules>` with a .cfnlintrc file in my root of my workspace, I get the following error:
```
  File "/REDACTED/cfnlint/config.py", line 211, in _read_config line 211, in _read_config
    config_template = yaml.safe_load(config.read_text())
AttributeError: 'PosixPath' object has no attribute 'read_text'
```
I believe read_text is not supported until Python 3.5.3 per some other Github Issues I saw on this in other repos.

Looking into the code-coverage looks like that line is not covered which is why its not failing the build - https://codecov.io/gh/awslabs/cfn-python-lint/src/master/src/cfnlint/config.py#L211

kddejong:
*Issue #, if available:*
Fix #673 
*Description of changes:*
Use the built-in cfn-lint YAML decoders

### DIFF
--- a/src/cfnlint/config.py
+++ b/src/cfnlint/config.py
@@ -21,8 +21,8 @@ import glob
 import json
 import os
 import six
-import yaml
 import jsonschema
+import cfnlint.decode.cfn_yaml
 from cfnlint.version import __version__
 try:  # pragma: no cover
     from pathlib import Path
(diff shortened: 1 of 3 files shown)


kddejong:
I'm really bummed about my testing here.  Not sure how to tackle it yet though.

kddejong:
got some testing added.  @cmmeyer this is ready

PR #678, opened 2019-02-26
 Add CloudFront allowed values

fatbasstard:
*Issue https://github.com/awslabs/cfn-python-lint/issues/50, if available:*

Add all the allowed values of the [`AWS::CloudFront`]( https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-reference-cloudfront.html)  Resources

* Also added configuration to inclusive/Exclusove and OnlyOne
* The CountryCode list is retrieved from [ISO itself](https://www.iso.org/obp/ui/#iso:pub:PUB500001:en) as mentioned in the docs. Validated with the available options in the Console.


src/cfnlint/data/ExtendedSpecs/all/03_value_types.json
```diff
@@ -555,6 +649,259 @@
           "aws.config"
         ]
       },
+      "CountryCodeAlpha2": {
```
PatMyron, line 652: @kddejong wondering if this should be switched to [`route53domains CountryCode` from `botocore`](https://github.com/boto/botocore/blob/2300e3a87e2835c93d832ee9c8ec7433cfa7fe18/botocore/data/route53domains/2014-05-15/service-2.json#L664)

There would be 21 Removals and 1 Addition (`"AN"`)... might need testing

PR #682, opened 2019-02-27
new rule to test billing mode relationship with provisioned throughput

### LINKED ISSUE #509: Feature request: Check for mutual exclusive attributes for DynamoDB capacity
Currently you can specify provisioned capacity and `BillingMode: PAY_PER_REQUEST` for a DynamoDB table and CloudFormation will process the resource without errors. In that case the provisioned capacity settings are prioritized over the billing mode, resulting in the table still being set to `BillingMode: PROVISIONED`.

I propose that `cfn-lint` checks that  `BillingMode: PAY_PER_REQUEST` and provisioned capacity on a table and its global indexes are mutually exclusive.

kddejong:
*Issue #, if available:*
Fix #509

*Description of changes:*
- Add new rule E3024 to check that BillingMode PAY_PER_REQUEST isn't specified with ProvisionedThroughput


PatMyron:
@kddejong should [`E3024`](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/docs/rules.md#E3024) leverage [`E2523`](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/docs/rules.md#E2523) by just adding here instead?

https://github.com/aws-cloudformation/cfn-python-lint/blob/04c1a4b87eecfa8568bc95e533195a6262eb30a3/src/cfnlint/data/AdditionalSpecs/OnlyOne.json#L65-L73

kddejong:
I don't believe so as E3024 is checking if value of property A is this then value of property B must be this.  We haven't found a way to define that in a spec file as of yet.

PatMyron:
https://github.com/aws-cloudformation/cfn-python-lint/blob/04c1a4b87eecfa8568bc95e533195a6262eb30a3/src/cfnlint/rules/resources/dynamodb/BillingMode.py#L30

true, was thinking [`AWS::DynamoDB::Table.BillingMode`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-dynamodb-table.html#cfn-dynamodb-table-billingmode) wouldn't be specified for `PROVISIONED` since it's the default, but still possible so this covers that

PR #683, opened 2019-02-27
Allow for ignoring of templates when running

### LINKED ISSUE #531: Exclude file from recursive search
*cfn-lint version:  v0.10.2*

I have plenty of CF template files in `iac` folder. I would like them to be recursively linting:
```
templates:
- iac/CloudFormation/**/*.yaml
```

But I would like also to ignore some files from directory. Do you support something like:
```
ignore-templates:
- iac/CloudFormation/notCfTempl.yaml
```

kddejong:
*Issue #, if available:*
Fix #531
*Description of changes:*
- Allow for ignoring certain templates when linting multiple files

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,8 @@ Example:
 ```yaml
 templates:
 - test/fixtures/templates/good/**/*.yaml
+ignore_templates:
+- codebuild.yaml
 include_checks:
 - I
 ```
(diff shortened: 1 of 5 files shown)


fatbasstard:
I've been thinking about this is well, and was thinking about multiples way to solve this. Some ideas that crossed my mind:

- Always ignore invalid templates in the case multiple templates are given
- Add in the `AWSTemplateFormatVersion: '2010-09-09'` test that's currently in the IDE plugins and use that for the "is this a CloudFormation" check
- Always ignore invalid templates, but that could/would be annoying when you're linting 1 file

Apart from that, like the solution as is 👍

kddejong:
@cmmeyer I will add some documentation

PR #686, opened 2019-02-27
Improve the line/column information of a Match with array support

fatbasstard:
The line/column number did not support arrays. As a result the error/warning was always given on the property, not the actual item that causes the error. I started noticing this when working on the Allowed Values rule.

This PR handles this scenario and returns the correct line/error number if the error was in an array.

**Before:**
image

**After:**
image

### DIFF
--- a/src/cfnlint/__init__.py
+++ b/src/cfnlint/__init__.py
@@ -812,12 +812,18 @@ class Template(object):
                 except AttributeError as err:
                     LOGGER.debug(err)
         else:
-            try:
-                for key in text:
-                    if key == path[0]:
-                        result = self._loc(key)
-            except AttributeError as err:
-                LOGGER.debug(err)
+            # If the last item of the path is an integer, and the vaue is an array,
+            # Get the location of the item in the array
+            if isinstance(text, list) and isinstance(path[0], int):
+                result = self._loc(text[path[0]])
+            else:
+                try:
+                    for key in text:
+                        if key == path[0]:
+                            result = self._loc(key)
+                except AttributeError as err:
+                    LOGGER.debug(err)
+
         return result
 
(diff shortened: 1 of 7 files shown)


fatbasstard:
I suddenly get 2 less failures on `test/fixtures/templates/quickstart/vpc-management.json`

Interesting...

🤔 

Looks like the deduplication is causing this:
- The template outputs 30 errors/warnings (Checked in the CLI and in the IDE)
- This is the case with and without this code change
- The test case expects 35 Matches

fatbasstard:
```
FAIL: test_templates (integration.test_quickstart_templates.TestQuickStartTemplates)
Test Successful JSON Parsing
----------------------------------------------------------------------
  File "/Users/frankvanboven/Documents/dev/cfn-python-lint/test/integration/test_quickstart_templates.py", line 111, in test_templates
    assert matched is True, 'Expected error {} at line {}, column {} in matches for {}'.format(c['Rule']['Id'], c['Location']['Start']['LineNumber'], c['Location']['Start']['ColumnNumber'], filename)
AssertionError: Expected error W3005 at line 277, column 5 in matches for test/fixtures/templates/quickstart/nist_high_master.yaml

======================================================================
FAIL: test_success_parse (module.cfn_json.test_cfn_json.TestCfnJson)
Test Successful JSON Parsing
----------------------------------------------------------------------
  File "/Users/frankvanboven/Documents/dev/cfn-python-lint/test/module/cfn_json/test_cfn_json.py", line 74, in test_success_parse
    assert len(matches) == failures, 'Expected {} failures, got {} on {}'.format(failures, len(matches), filename)
AssertionError: Expected 35 failures, got 33 on test/fixtures/templates/quickstart/vpc-management.json
```

PR #702, opened 2019-03-07
Fix an incompatibility with samtranslator 1.10.0

Dunedan:
`cfn-python-lint` suppressed some logging from `samtranslator`. As it
turned out, the logging itself got already removed in `samtranslator`
1.7.0, while they retained the import of the logging module. This import
got finally removed with `samtranslator` 1.10.0, which breaks
`cfn-python-lint` when using version 1.10.0 of `samtranslator`.

This commit removes the suppression of the logging in `cfn-python-lint`,
as no logging happens in the supported `samtranslator` versions anymore
anyway.


cmmeyer:
I'm wondering if we should got the extra step and increment to 1.10.0 in setup.py?  I just tested it locally and it passes for 2.7 and 3.7...

cmmeyer:
Thanks for tracking this down!

PR #717, opened 2019-03-12
Typos

jonathanmorley:
Typo fix

### DIFF
--- a/src/cfnlint/rules/functions/RelationshipConditions.py
+++ b/src/cfnlint/rules/functions/RelationshipConditions.py
@@ -44,7 +44,7 @@ class RelationshipConditions(CloudFormationLintRule):
                 for scenario in scenarios:
                     scenario_text = ' and '.join(
                         ['when condition "%s" is %s' % (k, v) for (k, v) in scenario.items()])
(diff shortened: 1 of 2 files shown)


src/cfnlint/rules/functions/RelationshipConditions.py
```diff
@@ -65,8 +66,9 @@ def match(self, cfn):
                 if value not in PSEUDOPARAMS:
                     scenarios = cfn.is_resource_available(getatt_obj, value)
                     for scenario in scenarios:
-                        scenario_text = ' and '.join(['when condition "%s" is %s' % (k, v) for (k, v) in scenario.items()])
-                        message = 'GetAtt to resource "{0}" that many not be available when {1} at {2}'
+                        scenario_text = ' and '.join(
+                            ['when condition "%s" is %s' % (k, v) for (k, v) in scenario.items()])
```
kddejong: Won't 'when' get repeated a lot of times in this scenario?

kddejong:
Thanks @jonathanmorley for submitting this PR and catching this.  I just have a question on how many 'when's we'll get and where we should pull the 'when' out of.

Issue #725, opened 2019-03-14
E3031 CidrIp contains invalid characters fails when Fn::Sub is present


seporaitis:
*cfn-lint version: (`cfn-lint --version`)*

`cfn-lint 0.16.0`

*Description of issue.*

When `CidrIp` value is `!Sub`ed from `Parameters` - E3031 lint error is raised. Sample template:

```lang=yaml
AWSTemplateFormatVersion: 2010-09-09

Description: AMI Builder Stack

Parameters:

  BuilderCidr:
    Type: String

Resources:

  SecurityGroup:
    Type: AWS::EC2::SecurityGroup
    Properties:
      GroupDescription: Description
      VpcId: vpc-id
      SecurityGroupIngress:
      - IpProtocol: tcp
        FromPort: 1
        ToPort: 65535
        CidrIp: !Sub ${BuilderCidr}
```

Expected output: successful lint
Actual output:

```
E3031 CidrIp contains invalid characters (Pattern: x.x.x.x/y) at Resources/SecurityGroup/Properties/SecurityGroupIngress/0/CidrIp/Fn::Sub
```

> Cfn-lint uses the [CloudFormation Resource Specifications](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html) as the base to do validation. These files are included as part of the application version. Please update to the latest version of `cfn-lint` or update the spec files manually (`cfn-lint -u`)

The problem still persists after running `cfn-lint -u`

fatbasstard:
Variables are replaced with ""... In this case a `!Ref` would be more "logical" btw, but the rule should not break ofcourse

seporaitis:
Ah, you're right about the `!Ref` - I will update my templates as well.

PR #729, opened 2019-03-14
Remove the !Sub checks in the Allowed Pattern rule

fatbasstard:
*Issue https://github.com/aws-cloudformation/cfn-python-lint/issues/725, if available:*

A `!Sub` can also be used as a `!Ref`, in which the string that is checked is empty. The PR handles this scenario (added the test scenario from the issue)


kddejong:
closing in favor of #731

PR #735, opened 2019-03-15
Fix the Matches output (again)

fatbasstard:
The previous hot-fix was not really correct (oops). The IDE plugins _always_ expect a json. So with no matches a `[]`.  The previous fix was incorrect since it didn't output anything.

Slighty reworked the code: The Formatter builds the output. Only if there is output it is printed
 

kddejong:
good catch

Issue #761, opened 2019-03-26
[cfn-lint] E2541:For all currently supported action types, the only valid version string is "1"."


andreaskemvise:
cfn-lint version: cfn-lint 0.17.0

I got the error message "[cfn-lint] E2541:For all currently supported action types, the only valid version string is "1"."

If I use a custom provider I can use a version string which differs to "1"

Cloudformation:```

     Stages:
        - Name: GitHub
          Actions:
            - Name: !Ref GitHubSourceRepo1
              ActionTypeId:
                Category: Source
                Owner: Custom
                Version: "3"
                Provider: GitHUB-custom
              Configuration:
                GitUrl: !Sub git@github.com:${GitHubUser}/${GitHubSourceRepo1}.git
                Branch: !Ref GitBranch1
                SSHKey: !Ref sshkey
                RepoName: !Ref GitHubSourceRepo1
```

**Comment by cmmeyer** [comment]
Historically the ActionId version was required to be “1” for all actions. Looking at the documentation , it’s changed to the following:

```
version
A string that describes the action version.

Type: String

Length Constraints: Minimum length of 1. Maximum length of 9.

Pattern: [0-9A-Za-z_-]+

Required: Yes
```

I’ll need to correct the rule.

kddejong:
Please try release v0.17.1

PR #767, opened 2019-03-27
E2532 allow new string pattern for ActionID version

chuckmeyer:
*Issue #, if available:*
#761 

*Description of changes:*

Modified CodePipeline rule to accept version strings as defined in API docs:
https://docs.aws.amazon.com/codepipeline/latest/APIReference/API_ActionTypeId.html#CodePipeline-Type-ActionTypeId-version

### DIFF
--- a/src/cfnlint/rules/resources/codepipeline/CodepipelineStageActions.py
+++ b/src/cfnlint/rules/resources/codepipeline/CodepipelineStageActions.py
@@ -169,16 +169,11 @@ class CodepipelineStageActions(CloudFormationLintRule):
         if isinstance(version, dict):
             self.logger.debug('Unable to validate version when an object is used.  Skipping')
         elif isinstance(version, (six.string_types)):
-            if len(version) < LENGTH_MIN:
-                message = 'Version string ({0}) is not minimum character length ({1})'
+            if not LENGTH_MIN <= len(version) <= LENGTH_MAX:
+                message = 'Version string ({0}) must be between {1} and {2} characters in length.'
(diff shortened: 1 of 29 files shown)


src/cfnlint/rules/resources/codepipeline/CodepipelineStageActions.py
```diff
@@ -159,15 +161,30 @@ def check_version(self, action, path):
...
         if isinstance(version, dict):
             self.logger.debug('Unable to validate version when an object is used.  Skipping')
-        elif version != '1':
-            message = 'For all currently supported action types, the only valid version string is "1".'
-            matches.append(RuleMatch(
-                path + ['ActionTypeId', 'Version'],
-                message
-            ))
+        elif isinstance(version, (six.string_types)):
+            if len(version) < LENGTH_MIN:
```
kddejong: This makes sense to me.  Only thing would just be simplification question?
Could we have done `if not LENGTH_MIN <= len(version) <= LENGTH_MAX:` and then just squash the message to include the min and the max?

chuckmeyer: Sure! The other thing I noticed is that if you don't hit min length, you don't comply with the regex, so that could get thrown as well.

kddejong:
To note:  We cannot extend the spec in this scenario as we are looking at some embedded json where the spec isn't defined.  So there is no ability to add `AllowedPatternRegex` to the spec as it exists today.

PR #824, opened 2019-04-13
Add Rule: CloudWatchLogGroups are must not duplicated in the Lambda's Events

### LINKED ISSUE #773: CloudWatchLogGroups are must not duplicated in the Lambda's Events section
*cfn-lint version: cfn-lint 0.9.0*

*Description of issue.*

image

In this configuration, Lambda for log analysis (LogSubscriptionFunction in this figure) subscribes and processes multiple Lambda logs output to CloudWatchLogs.
CloudFormation templates can be described as follows.

```.yaml
  LogSubscriptionFunction:
    Properties:
      CodeUri: subscription_function/build/
      Handler: app.lambda_handler
      Runtime: python3.6
      Events:
        FunctionALogGroup:
          Type: CloudWatchLogs
          Properties:
            LogGroupName: !Ref FunctionALogGroup
            FilterPattern: ""
        FunctionBLogGroup:
          Type: CloudWatchLogs
          Properties:
            LogGroupName: !Ref FunctionALogGroup # <--- log group is duplicated !!
            FilterPattern: ""
```

If the log group is duplicated in the Events section, the following error occurs when creating a stack:

> Resource limit exceeded. (Service: AWSLogs; Status Code: 400; Error Code: LimitExceededException; Request ID: 565efb75-522a-11e9-a6c3-61534eead0f1)

I think that it is necessary to have an implementation that causes an error by Linter when LogGroup is described in duplicate.

daisuke-awaji:
Issue #, if available:
Fix #773

Description of changes:
A new rule has been created to display a warning if CloudWatchLogs are duplicated in the Lambda event element.


kddejong:
so this looks pretty good.  It does handle objects inside the LogGroupName (Ref).  However it won't work in crazier condition based scenarios.   The good part is that it doesn't fail so I think for the most part this should work.

src/cfnlint/rules/resources/lmbd/EventsLogGroupName.py
```diff
@@ -0,0 +1,68 @@
...
+        matches = []
+        message = 'You must specify the AWS::Serverless::Function event correctly. ' \
+                  'LogGroups are duplicated. '
+
+        log_group_name_list = self.__get_log_group_name_list(cfn)
+
+        if self.__is_duplicated(log_group_name_list):
+            matches.append(
+                RuleMatch(
+                    'path', message.format()
```
PatMyron, line 38: not sure about this [`RuleMatch`](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/docs/getting_started/rules.md#rule-matches) path

Issue #791, opened 2019-04-02
LoadBalancerAttribute check


kisst:
*cfn-lint version: (`cfn-lint --version`)*
0.17.1
*Description of issue.*

As per the ELBv2 attributes [documentation](https://docs.aws.amazon.com/elasticloadbalancing/latest/APIReference/API_LoadBalancerAttribute.html) states, there is a short list of valid attributes for both ALB and NLB, and an even shorter list for specifically NLB / ALB. 

So this should work:
```
  NetworkLoadBalancer:
    Type: AWS::ElasticLoadBalancingV2::LoadBalancer
    Properties:
      Scheme: internal
      Type: network
      LoadBalancerAttributes:
        - Key: load_balancing.cross_zone.enabled
          Value: 'true'
```
this should not:
```
  NetworkLoadBalancer:
    Type: AWS::ElasticLoadBalancingV2::LoadBalancer
    Properties:
      Scheme: internal
      Type: network
      LoadBalancerAttributes:
        - Key: idle_timeout.timeout_seconds
          Value: '180'
```
Same way, this is valid:
```
  ApplicationLoadBalancer:
    Type: AWS::ElasticLoadBalancingV2::LoadBalancer
    Properties:
      Scheme: internal
      LoadBalancerAttributes:
        - Key: idle_timeout.timeout_seconds
          Value: '180'
```
This should not:
```
  ApplicationLoadBalancer:
    Type: AWS::ElasticLoadBalancingV2::LoadBalancer
    Properties:
      Scheme: internal
      LoadBalancerAttributes:
        - Key: load_balancing.cross_zone.enabled
          Value: 'true'
```

The full list right now 
ALB Attributes allowed values:
 - access_logs.s3.enabled
 - access_logs.s3.bucket
 - access_logs.s3.prefix
 - deletion_protection.enabled
 - idle_timeout.timeout_seconds
 - routing.http2.enabled

NLB attributes allowed values:
 - access_logs.s3.enabled
 - access_logs.s3.bucket
 - access_logs.s3.prefix
 - deletion_protection.enabled
 - load_balancing.cross_zone.enabled

kddejong:
Great suggestion.  I agree we should be able to validate this stuff.

kddejong:
Should be covered with Jamie's work.  Released with v0.21.0

PR #846, opened 2019-04-24
add rule for loadbalancer attributes

jvanbrunschot-coolblue:
*Issue #, if available:* #791 

*Description of changes:*

This will add a check for load balancer attributes. We can also try to add the value validation. 

Also fixed a bug: if you don't specify the type it defaults to "application" the current check, checks explicitly for type == 'application'.


src/cfnlint/rules/resources/elb/Elb.py
```diff
@@ -44,24 +44,48 @@ def check_protocol_value(self, value, path, **kwargs):
...
                     matches.append(RuleMatch(path, 'You must specify at least two SubnetMappings for load balancers with type "application"'))
 
         return matches
 
+    def check_loadbalancer_allowed_attributes(self, properties, path):
+        """ Validate loadbalancer attributes per loadbalancer type"""
+        matches = []
+        only_allowed_alb_values = ['idle_timeout.timeout_seconds', 'routing.http2.enabled']
+        only_allowed_nlb_values = ['load_balancing.cross_zone.enabled']
+        loadbalancer_attributes = properties.get('LoadBalancerAttributes')
```
kddejong: Few things I would change.  This just helps just in case someone really poorly formats their template or they use conditions.  What we usually trip up on is someone using a FindInMap or a Condition in the key or in the list.  These safe guards (I think) will help prevent the rule failing when the unusual but necessary may be done.
```python
loadbalancer_attributes = properties.get('LoadBalancerAttributes')
if isinstance(loadbalancer_attributes, list):   # validate the item exists and is a list (just in case)
  key = item.get('Key')   
  value = item.get('Value')
  if isinstance(key, six.string_types) and isinstance(value, (six.string_types, bool)):   # validate key, value are string values
    if self.is_application_loadbalancer(properties):
      if key in only_allowed_nlb_values:
```

jvanbrunschot-coolblue:
@kddejong Ola; applied your requested changes.

PR #852, opened 2019-04-29
Add new minor version for Amazon MQ

italics:
Amazon MQ nows supports version 5.15.9. Adding this to all regions.

https://aws.amazon.com/about-aws/whats-new/2019/04/amazon-mq-now-supports-activemq-minor-version-5-15-9/

### DIFF
--- a/src/cfnlint/data/CloudSpecs/ap-east-1.json
+++ b/src/cfnlint/data/CloudSpecs/ap-east-1.json
@@ -20774,7 +20774,8 @@
       "AllowedValues": [
         "5.15.0",
         "5.15.6",
-        "5.15.8"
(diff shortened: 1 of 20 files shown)


fatbasstard:
Good one! Only thing is that the spec file changes are "generated", can you perform the following steps:

1. Add the new version to the [Value Type spec patch](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/data/ExtendedSpecs/all/03_value_types.json#L27)
2. Run the spec file update: `cfn-lint -u` (or `cfn-lint --update-specs`.

We need to document our "Extended Specs" process/setup a bit better, I will take care of that.

PR #814, opened 2019-04-07
CodeBuild Service Role takes the Arn or the Name

kddejong:
*Issue #, if available:*
Fix #813
*Description of changes:*
- CodeBuild ServiceRole takes both the Arn of the Role and the Name of the Role
- New rule E3050 to check if a REFed IAM resource has a non default Path

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #866, opened 2019-05-07
For ECS Execution Role to be a ARN

### LINKED ISSUE #855: Stricter checks for ECS task definitions
*cfn-lint version: (`cfn-lint 0.19.1`)*

The CF spec files defines `AWS::ECS::TaskDefinition` as a string:

```json
   "AWS::ECS::TaskDefinition": {
"ExecutionRoleArn": {
          "Documentation": "http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-taskdefinition.html#cfn-ecs-taskdefinition-executionrolearn",
          "PrimitiveType": "String",
          "Required": false,
          "UpdateType": "Immutable"
        },
```

Unfortunately, the actual service expects that to be a full ARN, as the name implies, so it'll take a few minutes before you'll realize your mistake if you just use the short name as used in the UI. It'd be really nice if `cfn-lint` could catch this kind of error. On a related note touching on #238, this also made me think it'd be really helpful if there was a way to require this to be a reference or substitution to catch copy-and-paste errors.

kddejong:
*Issue #, if available:*
Fix #855
*Description of changes:*
- Have AWS::ECS::TaskDefinition ExecutionRoleArn be a IAM Role ARN

### DIFF
--- a/src/cfnlint/data/CloudSpecs/ap-east-1.json
+++ b/src/cfnlint/data/CloudSpecs/ap-east-1.json
@@ -15348,7 +15348,10 @@
           "Documentation": "http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-taskdefinition.html#cfn-ecs-taskdefinition-executionrolearn",
           "PrimitiveType": "String",
           "Required": false,
-          "UpdateType": "Immutable"
+          "UpdateType": "Immutable",
+          "Value": {
+            "ValueType": "IamRole.Arn"
+          }
         },
         "Family": {
           "Documentation": "http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-taskdefinition.html#cfn-ecs-taskdefinition-family",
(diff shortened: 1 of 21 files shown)


acdha:
This is working great for me and while fixing another project I just ran into a slightly confusing experience which is not entirely unlike #814:

```yaml
… task definition…
    ExecutionRoleArn: !Ref ConcordiaTaskRole
```

```
E3008 Property "ExecutionRoleArn" has no valid Refs to Resources at Resources/ConcordiaTask/Properties/ExecutionRoleArn/Ref
cloudformation/infrastructure/fargate-cluster.yaml:225:13
```

This is because the reference doesn't return an ARN so I need to use `GetAtt` instead, and it passes as soon as I correct that:

```yaml
… task definition…
    ExecutionRoleArn: !GetAtt ConcordiaTaskRole.Arn
```

Do you think that confusion is common enough that it'd be worth opening a separate issue to make the error message more informative? It would have been helpful to have the message say something along the lines of “ExecutionRoleArn value must be an ARN, not a Ref” but I don't know how much work that would require.

kddejong:
yea that message could be a little better.  I'm open to suggestions if you have any.

PR #891, opened 2019-05-13
Check for Serverless resources without Transform

orez-:
Previously cfn-lint accepted resources of Type AWS::Serverless::* unconditionally. With this commit it only accepts them with the AWS::Serverless-2016-10-31 Transform.

Let me know if this is the wrong file to define the test in, or if there's some false positive edge case I haven't considered.


fatbasstard:
Nice find 👍 

We indeed "ignore" serverless on certain locations. I'm just not sure if this is right place. One example is what happens when you add multiple resources:

```E3001 Serverless Transform required for Type AWS::Serverless::Function for resource Function in us-east-1
../cloudformation.yaml:30:5

E3001 Serverless Transform required for Type AWS::Serverless::Function for resource AgeFunction in us-east-1
../cloudformation.yaml:53:5
```

Maybe this could be a new rule? Since the check basically is a "if Serverles and no transform" check
🤔 

@kddejong Any thoughts on this?

orez-:
Ah interesting, so you're saying instead of adding a rule match for each Serverless resource just add one rule match on the Transform if there are any Serverless resources. I'll see if I can put that together.

fatbasstard:
@orez- ,

Yup! I'd advice to create a new rule to handle this specific use case. Keeps it clean, can be disabled and is basically a “feature” on it’s own (don't mix rules)

If you need need help on creating rules you can take a look at the [documentation](https://github.com/aws-cloudformation/cfn-python-lint/tree/master/docs/getting_started) or one of the other rules 👍

orez-:
Alright, I moved this check into its own rule. Let me know any more notes or issues, or if I've forgotten anything!

orez-:
I see that my change is failing a code coverage check, I think because the new rule has conditionals to stop checking when the template is misconfigured that aren't being exercised. Should I add test cases for those paths?

kddejong:
Thanks for updating this @orez-

Issue #893, opened 2019-05-14
AWS::ServiceDiscovery::Service validation error when no HealthCheckConfig or HealthCheckCustomConfig


rodrigosaito:
*cfn-lint version: 0.20.1*

*Description of issue.*

Please provide as much information as possible:
* Template linting issues:

The following template is valid but returns an error:
```
Resources:
  ServiceDiscovery:
    Type: AWS::ServiceDiscovery::Service
    Properties:
      Name: service-discovery-name
      Description: service discovery
      DnsConfig:
        DnsRecords:
          - Type: SRV
            TTL: 60
      NamespaceId: ns-id
```

Error:
```
E2523 Only one of [HealthCheckConfig, HealthCheckCustomConfig] should be specified for Resources/ServiceDiscovery/Properties
template.yaml:4:5
```

Official doc: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-servicediscovery-service.html

Still has the issue after `cfn-lint -u`.

fatbasstard:
Good find. It seemed like a logic implementation, but the documentation is actually really confusing now that I'm reading it again.

I removed this case from the rule to undo it. My apologies...

rodrigosaito:
@fatbasstard awesome, thanks for the quick reply!

kddejong:
Released with v0.20.2

PR #895, opened 2019-05-14
Remove OnlyOnly on AWS::ServiceDiscovery::Service

fatbasstard:
*Issue https://github.com/aws-cloudformation/cfn-python-lint/issues/893, if available:*

Remove the OnlyOne configuration of the `AWS::ServiceDiscovery::Service` health check configuration.

Re-reading the documentation I'm really confused:

On `HealthCheckCustomConfig ` it mentions:

```If you specify a health check configuration, you can specify either HealthCheckCustomConfig or HealthCheckConfig but not both.```

So from that case it seems the OnlyOne makes sense.. But what is "a health check configuration" in this case?

Looking at the documentation of `HealthCheckConfig` I'm lost. Remove this...

 
### DIFF
--- a/src/cfnlint/data/AdditionalSpecs/Exclusive.json
+++ b/src/cfnlint/data/AdditionalSpecs/Exclusive.json
@@ -72,6 +72,11 @@
         "ExtendedStatistic"
       ]
     },
+    "AWS::ServiceDiscovery::Service": {
+      "HealthCheckConfig": [
+        "HealthCheckCustomConfig"
+      ]
+    },
     "AWS::EC2::Instance": {
       "NetworkInterfaces": [
         "SubnetId"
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by kddejong** [comment]
So weird.  It has this information in the API docs too.  https://docs.aws.amazon.com/cloud-map/latest/api/API_CreateService.html

**Comment by kddejong** [comment]
We could move this to.   Exclusive as

```
    "AWS::ServiceDiscovery::Service": {
      "HealthCheckConfig": [
        "HealthCheckCustomConfig"
      ]
    }
```

fatbasstard:
Good point @kddejong . API documentation clarifies and indeed: Should be an Exclude instead OnlyOne (since "none given" is valid).

Implemented your suggestion 👍

PR #899, opened 2019-05-14
Change Debug strategy

fatbasstard:
_This is PR is open for discussion_

Change the strategy behind the `--debug` argument. The Debug flag has currently (mainly) been implemented as an extended logging of functions, rules and information. As a result the Debug argument returns in soo many output that it becomes "unusable".

This PR looks at the Debug argument from the "cfn-lint **user**" perspective. If I would run the linter, what would I expect from the debug?

From this I came to the following changes:

- Remove all debug from the helper functions and specific rules
- Make debug from the Config "smart" (Only output if there is data)
- Format JSON outputs generated by the Config and the Transformed template (So it's direct usable and readable)
- Print the full error on "suppressed" errors

_Still a WIP since I'm testing multiple templates for the debug output_

PR #909, opened 2019-05-16
only log when a matching function is called to help reduce logging

kddejong:
*Issue #, if available:*

*Description of changes:*
- Right now the matching function wrapper always log start and stop times.  This moves the logging to only be done if the function itself will be called.


fatbasstard:
Can you take a look at https://github.com/aws-cloudformation/cfn-python-lint/pull/899 ?

Complete related to this and takes it even a step further

fatbasstard:
This now makes sense 😄 Approved!

Issue #898, opened 2019-05-14
Remove deprecated Lambda runtimes


matsev:
*cfn-lint version: (`cfn-lint --version`)*
cfn-lint 0.20.1

*Description of issue.*
Some of the Lambda runtimes that are marked as valid by cfn-lint are no longer supported by CloudFormation, e.g. `nodejs4.3`. Please see the [Lambda runtime](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-function.html#cfn-lambda-function-runtime) documentation for allowed values and the [Runtime Support Policy](https://docs.aws.amazon.com/lambda/latest/dg/runtime-support-policy.html) page for deprecation schedule.

Cfn-lint uses the [CloudFormation Resource Specifications](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html) as the base to do validation. These files are included as part of the application version. Please update to the latest version of `cfn-lint` or update the spec files manually (`cfn-lint -u`)

rhbecker:
Similarly, I get errors when declaring the newly available `nodejs10.x`, even after I execute `cfn-lint -u` ...

> W2030 You must specify a valid allowed value for LambdaRuntime (nodejs10.x).
> Valid values are ['dotnetcore1.0', 'dotnetcore2.0', 'dotnetcore2.1', 'go1.x', 'java8', 'nodejs', 'nodejs4.3-edge', 'nodejs4.3', 'nodejs6.10', 'nodejs8.10', 'provided', 'python2.7', 'python3.6', 'python3.7', 'ruby2.5']

fatbasstard:
@rhbecker 10 is already added (https://github.com/aws-cloudformation/cfn-python-lint/issues/896 and https://github.com/aws-cloudformation/cfn-python-lint/pull/890) and will be available next release.

rhbecker:
ah, didn't dig deep enough ... thanks, @fatbasstard

fatbasstard:
@rhbecker No problem, we need to get a new release out ASAP...  Working on it!

fatbasstard:
@matsev Took some time to respond because I wanted to check something before "just removing it".

**A bit TL;DR, but my opinion/findings:**

We can/should not remove these values from the "Allowed Values" rule. This because the [runtime lifecycle is a bit special](https://docs.aws.amazon.com/lambda/latest/dg/runtime-support-policy.html):

[...]

PR #910, opened 2019-05-16
Add new rule to check for deprecated Lambda Runtimes (W2530)

fatbasstard:
*Issue https://github.com/aws-cloudformation/cfn-python-lint/issues/898 if available:*

Add a rule to give a Warning is a deprecated Runtime is used:

**features**
- A Warning since we cannot detect creates/updates in Lambda functions based on a static file.
- Time aware rule, no cfn-python-lint release required for the rule to become enable.
- Warning is raised when the `Deprecated (Update)` date is passed
- Support for "Rule Configuration": if `check_eol_date=true` is specified, the rule triggers on the EOL date. (CLI example `W2530:check_eol_date=true`
- Warning gives advice and the new Runtime to use. Decised to go for the "latest" version (in case of Node: `NodeJs10.x`)

**Might be interesting to make this `experimental` first **

Updated the tests to reflect this (we used `nodejs6.3` in most of our tests). Minor "risk" in this is that once NodeJS10 becomes dedicated, the tests will start failing. Accepting this for now as an acceptable risk.


kddejong:
The only thing I keep thinking about is personally I would like to know that I may have an issue before the issue occurs.  I don't want an error as I still have time to fix (past the EOL Date) it but I want to be able to visualize this issue before it becomes an issue.

Should there be different class rules for this?  
Should it be an Error if the deprecated date has occurred?
Could it be an Info level warning if the EOL has occurred?

fatbasstard:
@kddejong Check on my explanation in the Issue as well.

- The config toggle allows this somewhat. Trigger on the EOL date, or on the Deprecated date.
- Error is not an option, because (What I understand from it) the Lambda cannot be updated, but the rest of the template can. An Error is not true in that case
- Could remove the config and add that as an Informational rule (Just a decision)

cmmeyer:
I like the idea of moving the Lambda runtime lifecycle info into its own file (or enriching the runtime data in the spec) and then we could have two rules around eol and deprecated).

fatbasstard:
Easy to move into a file...

fatbasstard:
Closing in favor of https://github.com/aws-cloudformation/cfn-python-lint/pull/915

PR #914, opened 2019-05-17
Check bucket policy

### LINKED ISSUE #753: Improve s3 bucket policy linting
cfn-lint 0.16.0

*Description of issue.*
Add better linting for s3 bucket policies

I have the following template:
```
  ReportBucketPolicy:
    Properties:
      Bucket: !Ref 'ReportBucket'
      PolicyDocument:
        Statement:
          - Action:
              - s3:ListBucket
            Condition:
              StringEquals:
                aws:PrincipalOrgID: !Ref 'OrgId'
            Effect: Allow
            Principal: '*'
            Resource:
              - !Sub '${ReportBucket.Arn}/*'
            Sid: AllowOrgAccess
        Version: '2012-10-17'
    Type: AWS::S3::BucketPolicy
```
I initially had a typo so that I forgot the `$` in the `Sub()` so that it was like this: `- !Sub '${ReportBucket.Arn}/*'`

Can cfn-lint be made to catch that the resource specified was invalid?

The above template also has an issue because the resource shouldn't have a `*` for `s3:ListBucket`. This is the error that CloudFormation returns: `Action does not apply to any resource(s) in statement`

Is it within cfn-lint's scope to validate PolicyDocuments?

kddejong:
*Issue #, if available:*
Fix #753
*Description of changes:*
- Start to check S3 Bucket Policy for validity.  Check object/bucket calls and if the appropriate resource is included

### DIFF
--- a/src/cfnlint/rules/resources/s3/BucketPolicy.py
+++ b/src/cfnlint/rules/resources/s3/BucketPolicy.py
@@ -1,0 +1,238 @@
+"""
+  Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+
+  Permission is hereby granted, free of charge, to any person obtaining a copy of this
+  software and associated documentation files (the "Software"), to deal in the Software
+  without restriction, including without limitation the rights to use, copy, modify,
+  merge, publish, distribute, sublicense, and/or sell copies of the Software, and to
+  permit persons to whom the Software is furnished to do so.
+
+  THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED,
+  INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+  PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+  HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+  OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+  SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+"""
+import re
+import json
+import six
+from cfnlint import CloudFormationLintRule
+from cfnlint import RuleMatch
+from cfnlint.helpers import convert_dict
+
+
+class BucketPolicy(CloudFormationLintRule):
+    """Check if a S3 Bucket has a bucket policy"""
+    id = 'E3027'
+    shortdesc = 'Validate a S3 bucket\'s policy'
+    description = 'Check that a S3 Bucket Policy is syntactically correct'
+    source_url = 'https://docs.aws.amazon.com/AmazonS3/latest/dev/using-iam-policies.html'
+    tags = ['resources', 's3']
+
+    def __init__(self):
+        """Init"""
+        super(BucketPolicy, self).__init__()
+        self.resource_property_types = ['AWS::S3::BucketPolicy']
+
+    # pylint: disable=W0613
+    def check_value(self, value, path, start_mark, end_mark, bucket_resources, custom_resources):
+        """Count them up """
+
+        if isinstance(value, six.string_types):
+            try:
+                value = convert_dict(json.loads(value), start_mark, end_mark)
+            except Exception as ex:  # pylint: disable=W0703,W0612
+                return []
+
+        if isinstance(value, dict):
+            return self._check_policy(value, path[:], bucket_resources, custom_resources)
+
+        return []
+
+    def _check_policy(self, policy, path, bucket_resources, custom_resources):
+        """ Check the entire policy """
+        matches = []
+
+        key = 'Statement'
+        for statements, statements_path in policy.get_safe(key, default=[], path=path[:]):
+            if isinstance(statements, list):
+                for statement, statement_path in statements.items_safe(path=statements_path[:]):
+                    matches.extend(
+                        self._check_statement(
+                            statement, statement_path[:],
+                            bucket_resources, custom_resources))
+
+        return matches
+
+    def _check_statement(self, value, path, bucket_resources, custom_resources):
+        """ Check each statement"""
+        matches = []
+
+        actions = value.get('Action')
+        if not isinstance(actions, (list, six.string_types)):
+            return matches
+        resources = value.get('Resource')
+        if not isinstance(resources, (list, six.string_types, dict)):
+            return matches
+
+        if isinstance(actions, six.string_types):
+            actions = [actions]
+        if isinstance(resources, (six.string_types, dict)):
+            resources = [resources]
+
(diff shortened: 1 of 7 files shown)


src/cfnlint/rules/resources/s3/BucketPolicy.py
```diff
@@ -0,0 +1,238 @@
...
+            matches.append(RuleMatch(path[:] + ['Resource'], message))
+        elif not found_bucket_resources and found_bucket_actions:
+            message = 'S3 Bucket Policy has bucket actions and no specified bucket resource'
+            matches.append(RuleMatch(path[:] + ['Resource'], message))
+
+        return matches
+
+    def _get_statement_actions(self, actions):
+        """ Breaks down the actions in a statement to each category """
+        object_actions = [
```
fatbasstard: Can't we grab this data from the IAM policies? (from the experimental rule)

src/cfnlint/rules/resources/s3/BucketPolicy.py
```diff
@@ -0,0 +1,238 @@
...
+import re
+import json
+import six
+from cfnlint import CloudFormationLintRule
+from cfnlint import RuleMatch
+from cfnlint.helpers import convert_dict
+
+
+class BucketPolicy(CloudFormationLintRule):
+    """Check if Resources have a circular dependency"""
```
fatbasstard: 🤔

fatbasstard:
No test class?

src/cfnlint/rules/resources/s3/BucketPolicy.py
kddejong: Looking through that policy definition it doesn't look like it distinguishes between objects and the bucket.  So I don't believe so.  I wish I could though.

PatMyron: @kddejong thoughts on something like
```
object_actions = list(filter(re.compile("s3:.*Object").match, list_of_all_policies)) + [
            's3:AbortMultipartUpload',
            's3:ListMultipartUploadParts',
            's3:BypassGovernanceRetention',
]
```

kddejong: Let me compare the results.

kddejong: Updated to be more dynamic.

src/cfnlint/rules/resources/s3/BucketPolicy.py
```diff
@@ -0,0 +1,181 @@
...
+        for resource in resources:
+            if isinstance(resource, dict):
+                if len(resource) == 1:
+                    for func, func_value in resource.items():
+                        if func == 'Fn::Sub':
+                            if isinstance(func_value, six.string_types):
+                                if re.match('.+/.+', func_value):  # pylint: disable=R1703
+                                    found_object_resources = True
+                                else:
+                                    found_bucket_resources = True
```
fatbasstard: In this case, a `Resource: *` is counted as a "Bucket Resource"... 

Semantic not really true, is the the intended behaviour? Maybe treat `*` specifically or different?

src/cfnlint/rules/resources/s3/BucketPolicy.py
```diff
@@ -0,0 +1,181 @@
...
+
+        return found_object_actions, found_bucket_actions, True
+
+    def _get_statement_resources(self, resources, bucket_resources, custom_resources):
+        """ Get Statement Resources """
+        found_bucket_resources = False
+        found_object_resources = False
+        for resource in resources:
+            if isinstance(resource, dict):
+                if len(resource) == 1:
```
fatbasstard: Only 1 resource? Is this true? What if:

```
- Resource:
  - "bucketA"
  - "bucketB/object"
```

🤔

test/fixtures/templates/bad/resources/s3/bucket_policy.yaml
```diff
@@ -0,0 +1,114 @@
...
+            Effect: "Allow"
+            Resource:
+            - !GetAtt myBucket.Arn
+          - Action: s3:GetBucketLocation
+            Effect: "Allow"
+            Resource:
+            - !Sub "${myBucket.Arn}/*"
+          ### No Extra failures for this.  Prevents new actions from causing this rule to automaticall fail
+          ### or other new features
+          - Action: s3:aNewCall
```
fatbasstard: Maybe interesting to add this scenario to a new (information or warning) rule?

test/fixtures/templates/bad/resources/s3/bucket_policy.yaml
```diff
@@ -0,0 +1,114 @@
...
+Resources:
+  myBucket:
+    Type: AWS::S3::Bucket
+  BpBadActionWithBucketResource:
+    Type: AWS::S3::BucketPolicy
+    Properties:
+      Bucket: !Ref myBucket
+      PolicyDocument:
+        Statement:
+          - Action: s3:GetObject
```
fatbasstard: What about wilcard usages?

Like `s3:get*`?

src/cfnlint/rules/resources/s3/BucketPolicy.py
kddejong: In this case I think this is right.  Its because it can be 
- `Resource: !Sub "arn:aws:s3:::${myBucket}"`
- `Resource: !Ref myBucket` 
etc

src/cfnlint/rules/resources/s3/BucketPolicy.py
kddejong: Trying to think of the different scenarios here.  Since this is in a sub it gets really interesting.
- `Resource: !Sub "*"`
- `Resource: !Sub "arn:aws:s3:::${myBucket}"` where myBucket is a Resource
- `Resource: !Sub "arn:aws:s3:::${myBucket}"` where myBucket is a parameter (this could be any string... we should skip this?)
- `Resource: !Sub "arn:aws:s3:::*"`

Maybe I can't safely determine what is going on here to know what to do.  I think there is a possibility if I know myBucket is a resource that I can still make the appropriate determination.  Thoughts?

test/fixtures/templates/bad/resources/s3/bucket_policy.yaml
kddejong: Agreed.  I think we have this covered with the experimental rule right now.  I would like to get that thing out of experimental.

src/cfnlint/rules/resources/s3/BucketPolicy.py
kddejong: That being said I don't think I covered lists very well so I may have to add some logic there.

test/fixtures/templates/bad/resources/s3/bucket_policy.yaml
kddejong: added a test for it to get this covered.

src/cfnlint/rules/resources/s3/BucketPolicy.py
```diff
@@ -0,0 +1,181 @@
...
+import re
+import json
+import six
+from cfnlint import CloudFormationLintRule
+from cfnlint import RuleMatch
+from cfnlint.helpers import convert_dict, load_resources
+
+class BucketPolicy(CloudFormationLintRule):
+    """Check if a S3 Bucket has a bucket policy"""
+    id = 'E3027'
```
PatMyron: rule ID got taken in the meantime

PR #915, opened 2019-05-17
New rules for Lambda runtime EOL

kddejong:
*Issue #, if available:*

*Description of changes:*
- New rule E2531 to check if a lambda runtime is deprecated
- New rule W2531 to check if a lambda runtime is EOL

### DIFF
--- a/src/cfnlint/rules/resources/lmbd/DeprecatedRuntime.py
+++ b/src/cfnlint/rules/resources/lmbd/DeprecatedRuntime.py
@@ -15,83 +15,25 @@
   SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 """
 from datetime import datetime
+from cfnlint.helpers import load_resources
 from cfnlint import CloudFormationLintRule
-from cfnlint import RuleMatch
 
 
 class DeprecatedRuntime(CloudFormationLintRule):
     """Check if EOL Lambda Function Runtimes are used"""
-    id = 'W2530'
-    shortdesc = 'Check if EOL Lambda Function Runtimes are used'
-    description = 'Check if an EOL Lambda Runtime is specified and give a warning if used. '
-    source_url = 'https://docs.aws.amazon.com/lambda/latest/dg/runtime-support-policy.html'
-    tags = ['resources', 'lambda', 'runtime']
 
     def __init__(self):
         """Init"""
         super(DeprecatedRuntime, self).__init__()
-        self.config_definition = {
-            'check_eol_date': {
-                'default': False,
-                'type': 'boolean'
-            }
-        }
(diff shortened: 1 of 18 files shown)


src/cfnlint/rules/resources/lmbd/DeprecatedRuntime.py
```diff
@@ -0,0 +1,97 @@
...
+  PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+  HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+  OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE
+  SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+"""
+from datetime import datetime
+from cfnlint import CloudFormationLintRule
+
+
+class DeprecatedRuntime(CloudFormationLintRule):
```
fatbasstard: Doesn't this run as a standalone rule? Or is it filtered out based on the fact that there is no Meta data specified?

src/cfnlint/rules/resources/lmbd/DeprecatedRuntime.py
```diff
@@ -0,0 +1,97 @@
...
+    """Check if EOL Lambda Function Runtimes are used"""
+
+    def __init__(self):
+        """Init"""
+        super(DeprecatedRuntime, self).__init__()
+        self.resource_property_types.append('AWS::Lambda::Function')
+
+    current_date = datetime.today()
+
+    deprecated_runtimes = {
```
fatbasstard: Maybe move this to a JSON file (or make it part of the Spec files, but that might be interesting for later)?

kddejong: yea, I like that idea but any thoughts about how we would fold this into the spec file?

fatbasstard: For now just a standalone (like `OnlyOne.json`), to keep it consistent?

src/cfnlint/rules/resources/lmbd/DeprecatedRuntime.py
kddejong: It is filtered out so it doesn't run.

src/cfnlint/rules/resources/lmbd/DeprecatedRuntime.py
kddejong: Moved.

PR #921, opened 2019-05-21
Introduce new LogLevel

fatbasstard:
Split logging into 2 loglevels:

* **--debug**: Detailed debug information from a "developing on the linter" POV
* **--info**: Information from a "using the linter" POV

Default loglevel of the linter is now moved to "NOTSET", elevated the loglevel of the SAM transformation to `--info`

### DIFF
--- a/src/cfnlint/config.py
+++ b/src/cfnlint/config.py
@@ -33,15 +33,17 @@ except ImportError:  # pragma: no cover
 LOGGER = logging.getLogger('cfnlint')
 
 
-def configure_logging(debug_logging):
+def configure_logging(debug_logging, info_logging):
     """Setup Logging"""
     ch = logging.StreamHandler()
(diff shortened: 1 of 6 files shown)


src/cfnlint/config.py
```diff
@@ -33,15 +33,17 @@
...
-def configure_logging(debug_logging):
+def configure_logging(debug_logging, info_logging):
     """Setup Logging"""
     ch = logging.StreamHandler()
     ch.setLevel(logging.DEBUG)
 
     if debug_logging:
         LOGGER.setLevel(logging.DEBUG)
-    else:
+    elif info_logging:
```
cmmeyer: As written it seems like info logging would win. Perhaps reverse so highest level wins?

fatbasstard:
Alternative approach on https://github.com/aws-cloudformation/cfn-python-lint/pull/899

PR #927, opened 2019-05-24
AWS CDK template support

joshlartz:
When AWS CDK synthesizes a CloudFormation template, it places a custom type in the resources section called `AWS::CDK::Metadata` which the linter flags a E3001 about. 

Reference: https://docs.aws.amazon.com/cdk/latest/guide/tools.html
> The AWS::CDK::Metadata resource looks like the following.
```CDKMetadata:
  Type: "AWS::CDK::Metadata"
  Properties:
    Modules: "@aws-cdk/core=0.7.2-beta,@aws-cdk/s3=0.7.2-beta,lodash=4.17.10"```

This PR adds support for it in the custom handling section as well as adding a spec for it.

### DIFF
--- a/src/cfnlint/data/CloudSpecs/ap-east-1.json
+++ b/src/cfnlint/data/CloudSpecs/ap-east-1.json
@@ -13692,6 +13692,16 @@
         }
       }
     },
+    "AWS::CDK::Metadata": {
+      "Documentation": "https://docs.aws.amazon.com/cdk/latest/guide/tools.html",
+      "Properties": {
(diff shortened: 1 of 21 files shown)


joshlartz:
Sure. I wasn’t entirely sure if you would want to consider this part of the valid spec (yet), so I opted for the override method that was being used for custom resources. I’ll update with your suggestion.

joshlartz:
Added it to the spec, ran `cfn-lint -u`, and committed the results. Thanks for the help!

PR #931, opened 2019-05-24
Check Join for lists in the correct sections

### LINKED ISSUE #557: [Feature Request] Proper Linting of List-to-String Fn::Join Actions When Concat-Elements Include List-objects
**cfn-lint version: (`cfn-lint --version`)**

cfn-lint 0.10.2

**Description of issue:**

I make frequent use of nested-templates. Typically, a parent template will contain a call to a child-stack that creates one or more `SecurityGroup` objects and then passes those `SecurityGroup` objects to an EC2 or `AutoScalingGroup`/`LaunchConfiguration` child template as a comma-delimited string using a method similar to the following:

~~~json
"SecurityGroupIds": {
    "Fn::Join": [
        ",",
        [
            { "Fn::GetAtt": [ "SgRes", "Outputs.AppSg" ] },
            { "Fn::GetAtt": [ "SgRes", "Outputs.NasSg" ] },
            { "Fn::GetAtt": [ "SgRes", "Outputs.RdsSg" ] }
        ]
    ]
},
~~~

Recently, I received a request to add the ability to ensure a user-specifiable "management" `SecurityGroup` is applied to the resultant EC2s. So, I added a `Parameter` to my parent template similar to:

~~~json
"MgtSgs": {
    "AllowedPattern": "^$|^sg-[0-9a-z]{8}$|^sg-[0-9a-z]{17}$",
    "Default": "",
    "Description": "Security group(s) that need management access to the Gluster nodes' CLIs",
    "Type": "List<AWS::EC2::SecurityGroup::Id>"
},
~~~

I use the `List<AWS::EC2::SecurityGroup::Id>` type because my template users prefer to be able to select from a pull-down when using the template from the CFn web-console. Otherwise, I'd consider using a basic `String` type. When I first updated my `"SecurityGroupIds"` join-block, I'd tried:

~~~json
"SecurityGroupIds": {
    "Fn::Join": [
        ",",
        [
            { "Ref": "MgtSgs" }
            { "Fn::GetAtt": [ "SgRes", "Outputs.AppSg" ] },
            { "Fn::GetAtt": [ "SgRes", "Outputs.NasSg" ] },
            { "Fn::GetAtt": [ "SgRes", "Outputs.RdsSg" ] }
        ]
    ]
},
~~~

and

~~~json
"SecurityGroupIds": {
    "Fn::Join": [
        ",",
        [
            {
                "Fn::Join": [
                    ",",
                    [
                        { "Ref": "MgtSgs" }
                    ]
                ]
            },
            { "Fn::GetAtt": [ "SgRes", "Outputs.AppSg" ] },
            { "Fn::GetAtt": [ "SgRes", "Outputs.NasSg" ] },
            { "Fn::GetAtt": [ "SgRes", "Outputs.RdsSg" ] }
        ]
    ]
},
~~~

Both of which `cfn-lint` said were valid. However, when attempting to use either of the above to deploy a stack, the launches failed with:

~~~
An error occurred (ValidationError) when calling the CreateStack operation: Template error: 
   every Fn::Join object requires two parameters, (1) a string delimiter and (2) a list of 
   strings to be joined or a function that returns a list of strings (such as Fn::GetAZs) to be
   joined.
~~~

This error is obviously the correct result, and the fix (after much hair-pulling and knob-twizzling) was to change to doing:

~~~json
"SecurityGroupIds": {
    "Fn::Join": [
        ",",
        [
            {
                "Fn::Join": [
                    ",",
                    { "Ref": "MgtSgs" }
                ]
            },
            { "Fn::GetAtt": [ "SgRes", "Outputs.AppSg" ] },
            { "Fn::GetAtt": [ "SgRes", "Outputs.NasSg" ] },
            { "Fn::GetAtt": [ "SgRes", "Outputs.RdsSg" ] }
        ]
    ]
},
~~~

It would be great if `cfn-lint` could have flagged the faulty methods I'd tried. I'd likely still have experienced errors, but would have had them at the edit-stage rather than the attempt-to-deploy stage (though a meaningful error at the edit-stage would have saved that).

kddejong:
*Issue #, if available:*
Fix #557 
*Description of changes:*
Check for lists and strings in the right location

### DIFF
--- a/src/cfnlint/rules/functions/Join.py
+++ b/src/cfnlint/rules/functions/Join.py
@@ -17,6 +17,7 @@
 import six
 from cfnlint import CloudFormationLintRule
 from cfnlint import RuleMatch
+from cfnlint.helpers import RESOURCE_SPECS
 
 
 class Join(CloudFormationLintRule):
@@ -26,6 +27,17 @@ class Join(CloudFormationLintRule):
     description = 'Making sure the join function is properly configured'
     source_url = 'https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-join.html'
     tags = ['functions', 'join']
+
+    def __init__(self):
+        """Initialize the rule"""
+        super(Join, self).__init__()
+        self.list_supported_functions = []
+        self.singular_supported_functions = []
+        for intrinsic_type, intrinsic_value in RESOURCE_SPECS.get('us-east-1').get('IntrinsicTypes').items():
+            if 'List' in intrinsic_value.get('ReturnTypes', []):
+                self.list_supported_functions.append(intrinsic_type)
+            if 'Singular' in intrinsic_value.get('ReturnTypes', []):
+                self.singular_supported_functions.append(intrinsic_type)
 
     def _get_parameters(self, cfn):
         """Get all Parameter Names"""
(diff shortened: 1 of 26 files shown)


src/cfnlint/rules/functions/Join.py
```diff
@@ -27,81 +27,175 @@ class Join(CloudFormationLintRule):
...
-            'Ref'
+            'Ref',
         ]
 
+        template_parameters = self._get_parameters(cfn)
+        get_atts = cfn.get_valid_getatts()
+
+        if isinstance(join_string_objs, dict):
+            if isinstance(join_string_objs, dict):
+                if len(join_string_objs) == 1:
```
cmmeyer: ❓

kddejong: Fail. I'll get that fixed.

kddejong:
@chuckmeyer moved the intrinsic definitions to the spec

PR #933, opened 2019-05-25
Fill in Function alias to have a parameter value

### LINKED ISSUE #582: AutoPublishAlias with a !Ref value marked as false positive
*cfn-lint version: 0.11.1*

I'm running into an issue with `AutoPublishAlias`. When using a `!Ref` as its value, I get the following error message:
```
E0001 Resource with id [SkillFunction] is invalid. 'AutoPublishAlias' must be a string or a Ref to a template parameter
foo.yml:1:1
```

Here's the minimal template to reproduce the issue:
```yaml
---

Parameters:
  Stage:
    Description: Environment stage (deployment phase)
    Type: String
    AllowedValues:
      - beta
      - prod

Resources:
  SkillFunction:
    Properties:
      CodeUri: '.'
      Handler: main.handler
      Runtime: python3.7
      Timeout: 30
      MemorySize: 128
      AutoPublishAlias: !Ref Stage
```

The error message is actually very helpful (👏 for that), but as you can see, I am using a `!Ref` to a template parameter, so this should not cause a linting error.

kddejong:
*Issue #, if available:*
Fix #582
*Description of changes:*
- Pass parameter values to Transform as in some cases they are needed

### DIFF
--- a/src/cfnlint/transform.py
+++ b/src/cfnlint/transform.py
@@ -38,6 +38,7 @@ class Transform(object):
         self._filename = filename
         self._template = template
         self._region = region
+        self._parameters = {}
 
         self._managed_policy_map = self.load_managed_policies()
         self._sam_parser = parser.Parser()
@@ -70,6 +71,13 @@ class Transform(object):
             if resource_type == 'AWS::Serverless::Function':
 
                 Transform._update_to_s3_uri('CodeUri', resource_dict)
+                auto_publish_alias = resource_dict.get('AutoPublishAlias')
+                if isinstance(auto_publish_alias, dict):
+                    if len(auto_publish_alias) == 1:
+                        for k, v in auto_publish_alias.items():
+                            if k == 'Ref':
+                                if v in self._template.get('Parameters'):
+                                    self._parameters[v] = 'Alias'
             if resource_type in ['AWS::Serverless::LayerVersion']:
                 if resource_dict.get('ContentUri'):
                     Transform._update_to_s3_uri('ContentUri', resource_dict)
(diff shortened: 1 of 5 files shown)


src/cfnlint/transform.py
```diff
@@ -70,6 +71,13 @@ def _replace_local_codeuri(self):
             if resource_type == 'AWS::Serverless::Function':
 
                 Transform._update_to_s3_uri('CodeUri', resource_dict)
+                auto_publish_alias = resource_dict.get('AutoPublishAlias')
```
cmmeyer: Is this the only property that allows/requires a fake value for a !Ref? Does it make sense to do this for all properties on a Serverless resource?

src/cfnlint/transform.py
```diff
@@ -70,6 +71,13 @@ def _replace_local_codeuri(self):
...
+                if isinstance(auto_publish_alias, dict):
+                    if len(auto_publish_alias) == 1:
+                        for k, v in auto_publish_alias.items():
+                            if k == 'Ref':
```
cmmeyer: Do we need to handle !Join and !Sub as well?  Might be better to think of this as a general function to populate fake values for all params in a resource (yeah.... I know...)

src/cfnlint/transform.py
kddejong: From the SAM docs.
`
AutoPublishAlias             Ref of a CloudFormation Parameter  Alias resources created by SAM uses a LocicalId <FunctionLogicalId+AliasName>. So SAM either needs a string for alias name, or a Ref to template Parameter that SAM can resolve into a string.
`

src/cfnlint/transform.py
kddejong: From the SAM Docs.
`
AutoPublishAlias             Ref of a CloudFormation Parameter  Alias resources created by SAM uses a LocicalId <FunctionLogicalId+AliasName>. So SAM either needs a string for alias name, or a Ref to template Parameter that SAM can resolve into a string.
`

src/cfnlint/transform.py
kddejong: I don't see any other parameter REFs in https://raw.githubusercontent.com/awslabs/serverless-application-model/master/docs/cloudformation_compatibility.rst

src/cfnlint/transform.py
jlhood: @cmmeyer is right. This fix shouldn't be specific to `AutoPublishAlias`. My expectation was the fix should put dummy values in for any required template parameters (no default specified). That will future-proof this against other similar issues.

Issue #1002, opened 2019-07-06
Property Check Failing for AWS::Cognito::UserPool in ap-southeast-1


beck3905:
*cfn-lint version: (`cfn-lint --version`)*
v0.22.0

*Description of issue.*
AWS::Cognito::UserPool was added to src/cfnlint/data/ExtendedSpecs/ap-southeast-1/07_ssm_service_addition.json as a resource type.  Now, my template containing the following fails:

```
MyUserPool:
    Type: AWS::Cognito::UserPool
    Properties:
      UserPoolName: UserPoolName
      Policies:
        PasswordPolicy:
          MinimumLength: 8
      UsernameAttributes:
        - email
      Schema:
        - AttributeDataType: String
          Name: email
          Required: false
```

The failure occurs on `AWS::Cognito::UserPool.Policies` and rule E3002.  There is a KeyError raised here: https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/rules/resources/properties/Properties.py#L203.  My guess is that this issue is caused by something missing in the resource types or property types referenced by cfn-lint.

kddejong:
Released with v0.22.2

PR #1003, opened 2019-07-08
Fix Property Check Failing for AWS::Cognito::UserPool in ap-southeast-1

sambattalio:
*Issue #, if available:*
#1002

*Description of changes:*
I updated ap-southeast-1.json with the 'types' from the ap-southeast-2.json file. From local testing, it looks like that this fixes the issue mentioned above. 

Was previously throwing "E0002: Error processing rule on the template (Unknown exception while processing rule E3002: 'AWS::Cognito::UserPool.Policies') on files such as [this test file](https://github.com/awslabs/serverless-application-model/blob/develop/tests/translator/output/api_with_openapi_definition_body_no_flag.json).

Please let me know if there is anything I should change on this! Thanks


kddejong:
Thanks for submitting this.  I will take a look shortly.

PR #1004, opened 2019-07-08
Fix CloudSpecs data for Backup resources

zjullion:
*Issue #, if available:*
N/A

*Description of changes:*
 * Remove duplicated Backup resources from a number of regions
 * Rename resources to match actual names in CloudFormation
 * Adjust several doubles to be longs

Note: many of these changes conflict with current official CloudFormation documentation, as the docs are wrong. See this support forum thread: https://forums.aws.amazon.com/message.jspa?messageID=907303#907303

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1006, opened 2019-07-09
Fix backup cloudspecs

kddejong:
*Issue #, if available:*
- Replace #1004

*Description of changes:*
- Remove manual backup patching
- Make updates to backups based on issues with the documentation


cmmeyer:
To confirm, all this change is doing is refreshing the enriched specs files against  the latest version?

kddejong:
Yes and should include the fixes that are in 1006.  We had originally patched in all of the backups stuff. So part of this is taking that out too.

kddejong:
@zjullion the long story here is that backup wasn't originally included so we manually patched the entire thing in there.  They finally got it included in the spec so some of this pull request is removing our manual patching but then including your fixes as well.  I didn't manually patch to https because that seemed like overkill for what we needed here.

kddejong:
Also since I built this manually I did it entirely based on the documentation so I'm curious what your thoughts are on the spec file based on the patching you area doing.  It could be that the spec file is correct already but the documentation is wrong (or I just messed up with the original patch I created... which is why I'm happy to remove it).

zjullion:
Oops, sorry, I had started reviewing before you posted your comment and didn't see the bit about `https`. Do what you like with that, it's by no means a major issue.

kddejong:
Yea, I wish it was https and maybe its just me being lazy.  I'm going to talk to some internal folks about getting that changed to https over us patching it.  I don't see why it wouldn't be https but creating a bunch of patches to switch it may not be completely valuable to us at this time.

Issue #1007, opened 2019-07-09
Add option to force specific rules to be evaluated


jonathanmorley:
We are trying to use cfn-lint to enforce certain template patterns across the organization.

Currently users can opt-out of certain rules with Metadata in their templates (which they control).

We would like to be able to provide a list of rules which are forced to be run, and do not respect the Metadata-exclusions.

chuckmeyer:
Hi Jonathan, 

The current scope of cfn-lint is as a validation tool, not a governance tool. Other tools like cfn-nag may be better for this function.  I like to think of cfn-lint as "can I deploy this?" and cfn-nag as "should I deploy this?".

We may add a governance mode in the future, so we'll keep this in our backlog as an enhancement (feel free to dig in if you have a solution!), but I just wanted to give you the current state of the tool and the ecosystem.

atkinsonm:
@patmyron it looks like this was closed by #1243

PR #1016, opened 2019-07-10
Mandatory checks

yangas:
*Issue #, if available:* #1007 

*Description of changes:* Adding extra flag for mandatory checks which should supersede exclusions defined within templates.

**Disclaimer:** This is realistically more like pseudocode at this point. I'm not sure what the best way to get the args at the **get_directive** call but the way I'm doing it doesn't really work.


src/cfnlint/__init__.py
```diff
@@ -595,8 +595,13 @@ def get_directives(self):
...
         for _, resource_values in self.template.get('Resources', {}).items():
-            ignore_rule_ids = resource_values.get('Metadata', {}).get('cfn-lint', {}).get('config', {}).get('ignore_checks', [])
-            for ignore_rule_id in ignore_rule_ids:
+            ignore_rule_ids = set(resource_values.get('Metadata', {}).get('cfn-lint', {}).get('config', {}).get('ignore_checks', []))
+            mandatory_rule_ids = set([])
+            try:
+                mandatory_rule_ids = set(cfnlint.config.ConfigMixIn.mandatory_checks)
+            except:
+                pass
+            for ignore_rule_id in (ignore_rule_ids - mandatory_rule_ids):
```
kddejong: Ignore rules can be the prefix of the string as well.  So you can do E1 and that would ignore all E1XXX rules.  

https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/__init__.py#L206-L207

kddejong: The other thing we may need to figure out here is how to handle the directives section of exceptions.  So people can put in an exception in a resources metadata.  Those exceptions are filtered out here.  In this case the rule still runs but we filter out the results based on these directives.

https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/__init__.py#L1356-L1369

kddejong:
@yangas this is a good start and I'm willing to help here finish it up if you would like.  Just let me know.

yangas:
> @yangas this is a good start and I'm willing to help here finish it up if you would like. Just let me know.

Currently on a trip but will give it another go when I get back!

src/cfnlint/config.py
```diff
@@ -476,6 +481,11 @@ def ignore_checks(self):
         """ ignore_checks """
         return self._get_argument_value('ignore_checks', True, True)
 
+    @property
+    def mandatory_checks(self):
+        """ mandatory_checks """
+        return self._get_argument_value('mandatory_checks', True, True)
```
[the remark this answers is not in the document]
yangas: Updated accordingly, I agree it should not be overriden within a template.

src/cfnlint/__init__.py
```diff
@@ -591,11 +591,25 @@ def get_valid_getatts(self):
...
+                result.append(ignore_rule)
+        return result
+
     def get_directives(self):
         """ Get Directives"""
         results = {}
         for _, resource_values in self.template.get('Resources', {}).items():
             ignore_rule_ids = resource_values.get('Metadata', {}).get('cfn-lint', {}).get('config', {}).get('ignore_checks', [])
+            try:
+                mandatory_rule_ids = cfnlint.config.ConfigMixIn.mandatory_checks
```
yangas: @kddejong Do you know what the correct way to do this in this scope would be? Since doing it like this returns the object reference since it's an un-initialized class.

src/cfnlint/config.py
```diff
@@ -225,6 +225,17 @@ def comma_separated_arg(string):
     return string.split(',')
 
 
+def comma_separated_rule_names(string):
```
yangas: Added this with the hope that it would simplify things for evaluation when it comes to rule evaluations. But since one of the 2 supports wildcards essentially it would still present a need to unpack the values that match the prefix in the case of an overlap.
eg.
```
ignore_rules: [E3]
mandatory_rules: [E3001]
```
In the above scenario we would compare them and would need to leave all other rules to ignore except for the E3001.

@kddejong has there been any consideration on expanding the prefixes into all the rules they represent? This would result in higher overhead from iteration perspective but should convert comparisons into being explicit.

yangas:
@kddejong any thoughts or comments you can give on the above? Would appreciate some guidance

src/cfnlint/config.py
kddejong: @yangas are you saying should we expand `E3` into the full rule numbers?  I could absolutely see the value in that.  

We have to make sure we account for scanning multiple templates in one execution of cfn-lint.  Since the ignore rule can be specified in each template.

kddejong:
So I think I got this working but I'm kind of annoyed with the amount of passing of parameters I had to do.  I think some of your ideas here are better long term solutions.  So I'm a little stuck on doing whats necessary for this issue (based on the options today) or evaluating how we should covert some of this in the long term.  If its a priority to you I will get what I have in and can create another issue to work on cleaning it up.

Sorry this has taken a while to get to.

yangas:
> So I think I got this working but I'm kind of annoyed with the amount of passing of parameters I had to do. I think some of your ideas here are better long term solutions. So I'm a little stuck on doing whats necessary for this issue (based on the options today) or evaluating how we should covert some of this in the long term. If its a priority to you I will get what I have in and can create another issue to work on cleaning it up.
> 
> Sorry this has taken a while to get to.

No problem, just saw this, awesome!

I don't think it's that urgent on our end since we've built an intermediary approval process around our usage to resolve this need in the short-term.

src/cfnlint/__init__.py
```diff
@@ -150,16 +151,18 @@ def matchall_resource_sub_properties(self, filename, cfn, resource_properties, p
...
 
         # Whether "experimental" rules should be added
         self.include_experimental = include_experimental
 
         # Make Ignore Rules not required
         self.ignore_rules = ignore_rules or []
+        self.mandatory_rules = mandatory_rules or []
         self.include_rules = include_rules or []
         self.configure_rules = configure_rules or []
+        self.rules_trie = StringTrie()
```
yangas: @kddejong What are your thoughts on using a data structure such as trie in order to leverage for comparisons? This, I think, would be a good kick off point for doing prefix unwrapping and make it simpler to handle overlap between prefixed exclusions/mandatory_checks.

kddejong:
Closing as replaced by #1243

kddejong:
Closing as replaced by #1243

Issue #1022, opened 2019-07-12
[Feature request] cfn-lint --regions *


PatMyron:
This would currently be equivalent to something like this, but the list would grow as cfn-lint supports more regions:

```
cfn-lint --regions us-east-1 us-east-2 us-west-1 us-west-2 ca-central-1 eu-central-1 eu-west-1 eu-west-2 ap-northeast-1 ap-northeast-2 ap-southeast-1 ap-southeast-2 ap-south-1 sa-east-1 eu-west-3 ap-northeast-3 us-gov-west-1 us-gov-east-1 eu-north-1 ap-east-1
```

This could be incredibly helpful for those striving to deploy in every region, like AWS service teams

kddejong:
will be in the next release

PR #1026, opened 2019-07-15
Feat: Add ALL_REGIONS option for -r flag

sambattalio:
*Issue #, if available:*
#1022
*Description of changes:*
Added extra conditional in ConfigMixIn to catch "ALL_REGIONS". Based the change on how the default value is structured. Also added a test case for this.

Let me know if there are any adjustments/changes that can be made on this.


kddejong:
seems simple enough.

Issue #1033, opened 2019-07-17
SAM transform is only applied by cfn-lint if not declared as a list


boonew2:
*cfn-lint version: (`cfn-lint --version`)*
0.22.0
*Description of issue.*
cfn-lint works as expected if the SAM transform is specified like this:
```
Transform: AWS::Serverless-2016-10-31
```
but if multiple transforms are needed (or just declared as a list) like:
```
Transform: 
  - AWS::Serverless-2016-10-31
```
the template is not transformed before the validation occurs and that causes an assortment of false positives.
Please provide as much information as possible:
Working Example:
```
AWSTemplateFormatVersion: '2010-09-09'
Transform: AWS::Serverless-2016-10-31
Globals:
  Function:
    Timeout: 3
    Runtime: nodejs8.10
    Environment:
      Variables:
        VARIABLE: VALUE

Resources:
  Function:
    Type: AWS::Serverless::Function
    Properties:
      CodeUri: src/
      Handler: index.handler
      Events:
        API:
          Type: Api
          Properties:
            Path: /
            Method: GET
```
Example that yells about `E1001 Top level item Globals isn't valid`:
```
AWSTemplateFormatVersion: '2010-09-09'
Transform: 
  - AWS::Serverless-2016-10-31
Globals:
  Function:
    Timeout: 3
    Runtime: nodejs8.10
    Environment:
      Variables:
        VARIABLE: VALUE

Resources:
  Function:
    Type: AWS::Serverless::Function
    Properties:
      CodeUri: src/
      Handler: index.handler
      Events:
        API:
          Type: Api
          Properties:
            Path: /
            Method: GET
```

Cfn-lint uses the [CloudFormation Resource Specifications](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html) as the base to do validation. These files are included as part of the application version. Please update to the latest version of `cfn-lint` or update the spec files manually (`cfn-lint -u`)

kddejong:
Thanks for submitting this.  Working on it.

boonew2:
@kddejong 
Was the decision to not do the SAM transform at all if multiple transforms were specified just because it didn't seem like the template would be in a knowable state without mystery transforms  being applied? 

The specific use case i had was a transform running before after SAM to paper over a limitation in it ( https://github.com/awslabs/serverless-application-model/issues/717 ):
```
AWSTemplateFormatVersion: 2010-09-09
Transform:
[...]

PR #1056, opened 2019-07-26
Applies SAM transform if present in transform declaration

boonew2:
*Issue #, if available:*
#1033 
*Description of changes:*
Tweaks the first change ( https://github.com/aws-cloudformation/cfn-python-lint/pull/1042 ) a bit so that the sam transform is applied if it is anywhere in the transform. Without the unsupported transforms applied it _might_ end up in a bad state still, but by running the SAM transform it should have a better chance at not throwing false positives!

Let me know if i should open a bonus issue since #1033 is already closed

I'm going to look at how the code works some more and might submit another PR in this area to refactor the transform section to not be as coupled to the SAM transform exclusively if that seems like a good idea.


boonew2:
I probably should tweak the test templates so that they will fail before this change actually. Working on that now

src/cfnlint/__init__.py
```diff
@@ -1327,28 +1327,22 @@ def __init__(
...
         if transform_type:
-            if isinstance(transform_type, list):
-                if len(transform_type) != 1:
-                    return matches
-                if transform_type[0] != 'AWS::Serverless-2016-10-31':
-                    return matches
-            if transform_type != 'AWS::Serverless-2016-10-31':
+            if isinstance(transform_type, list) and sam_transform not in transform_type:
+                return matches
+            if transform_type != sam_transform:
```
kddejong: I think you want this
```
if isinstance(transform_type, six.string_types) and transform_type != sam_transform:
                return matches
``` 

other wise the list won't match the string.

kddejong: To give more to this when you have this.
```
Transform: 
  - UnsupportedTransform
  - AWS::Serverless-2016-10-31
```

The first `if isinstance(transform_type, list):` won't return matches because it found `AWS::Serverless-2016-10-31` which means it hits the second if.  Since the list above won't match `AWS::Serverless-2016-10-31` ti returns matches and doesn't do the transform.  I believe your second `if` is to cover the case when its just a string so we should include that check as well.  Thoughts?

boonew2: 🤦‍♂  Yeah you are right 🤦‍♂

boonew2: @kddejong 
Are you good with the change to normalize the transform_type check as a list to avoid the type checking below? It feels cleaner to me, but i can just add the `isinstance(transform_type, six.string_types)` if that is more inline with project conventions

kddejong: Yea I'm good with that.

spockNinja:
Thanks a ton for this fix. We just ran into this issue, as we are mixing a custom transform with SAM. I'm looking forward to seeing this change hit pypi.

kddejong:
I'm hoping we will have it out there some time tomorrow.  Just waiting on approval of the release PR.

kddejong:
Updated!  Should be available for you.

PR #1067, opened 2019-07-30
Add rule I1022 to recommend Sub over Join when empty delimiter

### LINKED ISSUE #917: Prefer Fn::Sub over Fn::Join for simple substitution
*cfn-lint version: (`cfn-lint --version`)* 0.20.1

*Description of issue.*

`!Sub` is generally more readable than `!Join` for simple substitution. A common design pattern is to use `!Join` with an empty string delimiter `''`. `!Sub` can accomplish this same function but is often easier to read which aids others in interpreting a template. There are other cases but empty string is the most obvious.

kddejong:
*Issue #, if available:*
Fix #917 
*Description of changes:*
- Add rule I1022 to recommend Sub over Join when join is using empty delimiter


src/cfnlint/rules/functions/SubNotJoin.py
```diff
@@ -0,0 +1,44 @@
...
+
+        join_objs = cfn.search_deep_keys('Fn::Join')
+
+        for join_obj in join_objs:
+            if isinstance(join_obj[-1], list):
+                join_operator = join_obj[-1][0]
+                if isinstance(join_operator, six.string_types):
+                    if join_operator == '':
+                        matches.append(RuleMatch(
+                            join_obj[0:-1], 'Prefer using Fn::Sub over Fn::Join with an empty delimiter'))
```
chuckmeyer: Does it make sense to link out to how to do this or give some sort of breadcrumb to help the author correct this?

Issue #971, opened 2019-06-20
Color output in the terminal


direvus:
*cfn-lint version: (`cfn-lint --version`)*

cfn-lint 0.21.5

*Description of issue.*

Colourising the output of cfn-lint in the terminal would be a real advantage for readability.

Output in the style of [ack](https://beyondgrep.com/) would be brilliant, we could show the actual lines that were flagged, together with line numbers, maybe some surrounding context.

PatMyron:
Circling back on this, [the Visual Studio Code extension](https://marketplace.visualstudio.com/items?itemName=kddejong.vscode-cfn-lint) sorts by rule level with colored icons if anyone is looking for a solution available today
<img width="30" alt="Screen Shot 2019-12-13 at 6 38 47 PM" src="https://user-images.githubusercontent.com/7014355/70842384-d2cc4d00-1dd7-11ea-80ce-9bac8e7a75ec.png">

direvus:
Nice!

PR #1103, opened 2019-08-20
WIP: Add PrettyFormatter for colored output in a terminal.

kadrach:
See #971.
Looking for some feedback before I go further down this rabbit hole.

*Description of changes:*
Added some colors to the output if not connected to a tty.

While grep-like context is not hard to add, all grep variants support a bazillion output options, adding significant complexity. 

I don't have a Windows system to hand to test if adding `colorama` is all that is required.

Also, this may be a breaking change, so perhaps this should just be a `-f pretty` output (but should that be default at some point?).

image

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -82,6 +82,8 @@ setup(
         'jsonschema~=3.0',
         'pathlib2>=2.3.0;python_version<"3.4"',
         'setuptools',
+        'termcolor',
+        'colorama;platform_system=="Windows"'
     ],
     python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*',
     entry_points={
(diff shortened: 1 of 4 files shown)


PatMyron:
@kadrach I really like the idea of different colors for different level rules!

kadrach:
@kddejong Can I assume the thumbs up means "no changes required" (tests aside)? As implemented, this may introduce breaking changes (the "default" output of the CLI changes).

kddejong:
@kadrach That's a good question.  I would hope anyone using this that would care about the formatting output would be using JSON.  However, I'm not sure we can assume this.  What about a different output type?

kddejong:
Replaced with #1742.  Thanks for submitting this.

PR #1116, opened 2019-09-01
Add max/min number types to values in spec

kddejong:
*Issue #, if available:*

*Description of changes:*
- Add value references for new number min/max types


src/cfnlint/data/ExtendedSpecs/all/03_value_types/aws_rds.json
```diff
@@ -22,12 +22,5 @@
...
       "NumberMin": 0
     }
-  },
-  {
-    "op": "add",
-    "path": "/ValueTypes/AWS::RDS::DBInstance.Iops",
-    "value": {
-      "NumberMin": 1000
-    }
   }
```
PatMyron, line 25: lost this constraint

PR #1118, opened 2019-09-03
Add link to emacs integration

soren:
Added link to an Emacs wiki page, that contains a description of how to integrate cfn-lint with Emacs.


chuckmeyer:
Hey, @soren -- thanks for the link! I notice this seems to only cover YAML templates. Any chance it could be extended to include JSON templates as well?

chuckmeyer:
Feedback I got from a coworker who uses emacs:

> it correctly shows squiggles for me but for some reason its not parsing the messages out
that tells me the error regex is not quite right
but it seems close

soren:
> Hey, @soren -- thanks for the link! I notice this seems to only cover YAML templates. Any chance it could be extended to include JSON templates as well?

Of course, I will look into adding JSON support.

soren:
I've added support for JSON templates :-)

soren:
> Feedback I got from a coworker who uses emacs:
> 
> > it correctly shows squiggles for me but for some reason its not parsing the messages out
> > that tells me the error regex is not quite right
> > but it seems close

I'm not exactly sure what issue your coworker has.

I get squiggles and when I move the mouse over the squiggles the corresponding message is show in a tooltip. To show all message below the buffer, you'll have to press `C-c ! l` that run `flycheck-list-errors`. You can also click "FlyC" in the mode line to display a menu with options:

image

PR #1172, opened 2019-10-24
consolidating on PSEUDOPARAMS from cfnlint.helpers

PatMyron:
`AWS::NoValue` was missing from 

https://github.com/aws-cloudformation/cfn-python-lint/blob/c73209fe2ba24c511f040a6eecb6da878b2d5ac6/src/cfnlint/rules/functions/Sub.py#L104-L112

Not sure if that was intentional or not


kddejong:
Digging into these failures

src/cfnlint/rules/functions/Sub.py
```diff
@@ -101,22 +102,12 @@ def _test_parameter(self, parameter, cfn, parameters, tree):
...
-            'AWS::URLSuffix',
-        ]
-
         odd_list_params = [
             'CommaDelimitedList',
             'AWS::SSM::Parameter::Value<CommaDelimitedList>',
         ]
 
-        valid_params = valid_pseudo_params
+        valid_params = PSEUDOPARAMS
```
kddejong: You need to switch this to `valid_params = copy(PSEUDOPARAMS)` you are getting hit by a Python thing here.

At the top.
`from copy import copy`

PatMyron: Good catch

PR #1146, opened 2019-09-29
Remove unneeded patches as of 2019.9.29

kddejong:
*Issue #, if available:*
Fix #1130 
*Description of changes:*
- Update patch specs to what is in spec version 6.2.0

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1203, opened 2019-11-18
remove AWS::EC2::VPCEndpointService spec patch

PatMyron:
https://github.com/aws-cloudformation/cfn-python-lint/pull/1146

[`us-east-1` specification](https://d1uauaxba7bl26.cloudfront.net/latest/gzip/CloudFormationResourceSpecification.json) now has:

```
    "AWS::EC2::VPCEndpointService": {
      "Documentation": "http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-vpcendpointservice.html",
      "Properties": {
        "NetworkLoadBalancerArns": {
          "PrimitiveItemType": "String",
          "Type": "List",
          "Required": true,
          "Documentation": "http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-vpcendpointservice.html#cfn-ec2-vpcendpointservice-networkloadbalancerarns",
          "UpdateType": "Mutable"
        },
        "AcceptanceRequired": {
          "Required": false,
          "Documentation": "http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-vpcendpointservice.html#cfn-ec2-vpcendpointservice-acceptancerequired",
          "PrimitiveType": "Boolean",
          "UpdateType": "Mutable"
        }
      }
    },
```


kddejong:
It doesn't look like that resource has made it into another regions specs.  By merging this in we will drop it from every region besides us-east-1.  Is that what we want?

PatMyron:
In some other regions' specs as well:
https://github.com/ScriptAutomate/aws-cfn-resource-specs/blob/2e0dd4570e6b6e145c1234fb55ef78e77787c1c3/supported-regions-per-resource.json#L29311-L29316

Definitely still missing from some regions that are supported though. I'll reach out to that team again to publish their correct regional availability

I was thinking most people use the default (`us-east-1`), so I'd like it to always have the latest from the official `us-east-1` specification and also don't want to assume the resource type is in every region

PatMyron:
@kddejong reached out to their team. We can wait for them to publish their correct regional availability before merging this

PatMyron:
@kddejong resource specification version 9.0.0 out now with correct `AWS::EC2::VPCEndpointService` regional availability

Issue #1209, opened 2019-11-21
Support appending rules from installed packages


zroger:
I would like to create a python package of my organization's custom rules, which could be installed alongside `cfn-lint` and then referenced from the CLI and/or `.cfnlintrc` by import path rather than directory path.

For example, with a custom package named `myrules` and a module within that package named `recommended`:

```shell
# install via pip
pip install cfn-lint
pip install myrules

# invoke via CLI
cfn-lint -a myrules.recommended ...
```

```yaml
# or in .cfnlintrc
append_rules:
  - myrules.recommended
```

I'm happy to put together a PR if this is something that would be considered.

kddejong:
I like this idea.  I wonder how hard it would be to accomplish.  I'm open to it.

kddejong:
I  believe this is complete.  Closing.  We can reopen if needed.

PR #1216, opened 2019-11-22
Support module paths in append_rules

zroger:
Addresses Issue #1209 

This makes it so that the `append_rules` option accepts module import paths as well as directory paths. If the value is a directory, then the existing loading operation takes place. Otherwise, the path is used as an import path, and the resulting module is scanned for rules classes in the same way as any module found by the directory scanner.


kddejong:
Very cool work.  Should have this in shortly.

PR #1057, opened 2019-07-27
Mandatory checks

kddejong:
*Issue #, if available:*

*Description of changes:*
- Allow for the configuration of mandatory checks.  These checks cannot be ignored

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1243, opened 2019-12-04
Mandatory checks

jonathanmorley:
*Issue #, if available:* #1007

*Description of changes:* Adding extra flag for mandatory checks which should supersede exclusions defined within templates, or in config.

This replaces #1016 and #1057  

### DIFF
--- a/docs/getting_started/integration.md
+++ b/docs/getting_started/integration.md
@@ -17,7 +17,7 @@ template = cfnlint.decode.cfn_yaml.load(filename)
 cfnlint.core.configure_logging(None)
 
 # Initialize the ruleset to be applied (no overrules, no excludes)
-rules = cfnlint.core.get_rules([], [], [])
+rules = cfnlint.core.get_rules([], [], [], [], False, [])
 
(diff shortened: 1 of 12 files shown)


kddejong:
Thank you so much for doing this.  Removing the package dependencies is helpful as that causes some extra work on my side.

I'm working on some code cleanup around the core libraries.  I want to get this in so I can restructure it.  I was going to build it out as part of that but this will save me a lot.

PatMyron:
Does `--mandatory-checks` work for Informational rules? Struggling to get it working with [this](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/test/fixtures/templates/bad/approaching_limit_size.yaml):

```
cfn-lint approaching_limit_size.yaml -c I | grep 'I1003'
I1003 The template description (994 bytes) is approaching the limit (1024 bytes)
cfn-lint approaching_limit_size.yaml -i E W -m I1003 | grep 'I1003'
cfn-lint approaching_limit_size.yaml -m I1003 | grep 'I1003'
```

PR #1244, opened 2019-12-04
Remove UUOC

awkspace:
*Description of changes:*
This commit removes a useless use of cat from the stdin example.


PatMyron:
@awkspace `cat` obviously isn't needed, but the original seems more readable to me. Do you have specific concerns around the usage of cat?

PR #1272, opened 2019-12-23
Adding extra attributes to specific rules

andrew-glenn:
*Description of changes:*

This PR adds `actual_type` and `expected_type` attributes to specific rules/conditions/test-cases. This is an extension of #570, where support for these attributes was added. 

The intent here is to expose certain metadata that will be useful for auto-remediation. I have a basic auto-remediation script in place locally for type-checks (expected int, found string, so forth).  Currently, it's function over form, but this PR will help refine a few less-common use cases. 


PatMyron:
@andrew-glenn auto-remediation is interesting, I like the idea. Do you have any examples of templates and their messages before/after these changes?

PR #1289, opened 2020-01-08
Remove support of EOL 3.4 add in 3.8

kddejong:
*Issue #, if available:*

*Description of changes:*
- Remove support for Python 3.4.  
- Add in support for Python 3.8

PyYAML a dependent package has removed support.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,46 +2,47 @@ language: python
 sudo: true
 dist: xenial
 services:
-- docker
+  - docker
 stages:
-- Linting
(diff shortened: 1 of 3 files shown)


kddejong:
Current download proportion rates for Python 3.4 are less than 1%.  They will have to pin an older version to continue to download.  

@PatMyron maybe we pin pyyaml once to a version supported by Python 3.4  Release. Then EOL.
That way users can have a version that will work until they can migrate.

PR #1292, opened 2020-01-08
switch to importlib from imp for python 3.x

### LINKED ISSUE #1285: Fix 'imp' Warning on Python 3
*cfn-lint version: (`cfn-lint --version`)* 0.26.2

*Description of issue.*

I see this warning on Python 3.7:

```
.../cfnlint/helpers.py:11: DeprecationWarning: the imp module is deprecated in favour of importlib; see the module's documentation for alternative uses
    import imp
```

On Python 3 you can use `importlib` to the same effect as the usage of `imp`.

kddejong:
*Issue #, if available:*
Fix #1285 
*Description of changes:*
- Switch to importlib from imp for python 3.x

### DIFF
--- a/src/cfnlint/helpers.py
+++ b/src/cfnlint/helpers.py
@@ -283,46 +283,58 @@ def create_rules(mod):
     return result
 
 
+if sys.version_info < (3,):
+    def import_filename(pluginname, root):
+        """ import_filename imports a module from a file"""
+        fh = None
+        try:
+            fh, filename, desc = imp.find_module(pluginname, [root])
+            mod = imp.load_module(pluginname, fh, filename, desc)
+            return mod
+        finally:
+            if fh:
+                fh.close()
+
+        return None
+
+else:
+    loader_details = (
+        importlib.machinery.SourceFileLoader,  # pylint: disable=no-member
+        importlib.machinery.SOURCE_SUFFIXES  # pylint: disable=no-member
+    )
+
+    def import_filename(pluginname, root):
+        """ import_filename imports a module from a file"""
+        mod_finder = importlib.machinery.FileFinder(  # pylint: disable=no-member
+            root, loader_details)
+
+        mod_spec = mod_finder.find_spec(pluginname)
+        if mod_spec is not None:
+            if sys.version_info < (3, 5):
+                # for python 2.7 disabling pylint checks
+                mod = mod_spec.loader.load_module()  # pylint: disable=no-member
+                return mod
+            # for python 2.7 disabling pylint checks
+            mod = importlib.util.module_from_spec(mod_spec)  # pylint: disable=no-member
+            mod_spec.loader.exec_module(mod)
+            return mod
+
+        return None
+
+
 def load_plugins(directory):
     """Load plugins"""
     result = []
-    fh = None
 
     def onerror(os_error):
         """Raise an error"""
         raise os_error
 
     for root, _, filenames in os.walk(directory, onerror=onerror):
-        if sys.version_info < (3,):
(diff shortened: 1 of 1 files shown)


src/cfnlint/helpers.py
```diff
@@ -290,15 +293,30 @@ def onerror(os_error):
...
+                importlib.machinery.SourceFileLoader,  # pylint: disable=no-member
+                importlib.machinery.SOURCE_SUFFIXES  # pylint: disable=no-member
+            )
+            mod_finder = importlib.machinery.FileFinder(  # pylint: disable=no-member
+                root, loader_details)
+            for filename in fnmatch.filter(filenames, '[A-Za-z]*.py'):
+                pluginname = filename.replace('.py', '')
+                mod_specs = mod_finder.find_spec(pluginname)
+                if mod_specs is not None:
+                    mod = mod_specs.loader.load_module()
```
kddejong: I think load_module is deprecated too.

kddejong:
@adamchainz how does this look?

src/cfnlint/helpers.py
```diff
@@ -290,15 +293,36 @@ def onerror(os_error):
...
-        for filename in fnmatch.filter(filenames, '[A-Za-z]*.py'):
-            pluginname = filename.replace('.py', '')
-            try:
-                fh, filename, desc = imp.find_module(pluginname, [root])
-                mod = imp.load_module(pluginname, fh, filename, desc)
-                result.extend(create_rules(mod))
-            finally:
-                if fh:
-                    fh.close()
+        if sys.version_info < (3,):
```
adamchainz: split the changed code into a function `import_filename` or something? Then you can define it per version, e.g. https://github.com/scoutapp/scout_apm_python/blob/master/src/scout_apm/compat.py#L34

then you can have a single `for filename` loop :)

src/cfnlint/helpers.py
```diff
@@ -290,15 +293,36 @@ def onerror(os_error):
...
+                try:
+                    fh, filename, desc = imp.find_module(pluginname, [root])
+                    mod = imp.load_module(pluginname, fh, filename, desc)
+                    result.extend(create_rules(mod))
+                finally:
+                    if fh:
+                        fh.close()
+        else:
+            loader_details = (
+                importlib.machinery.SourceFileLoader,  # pylint: disable=no-member
```
adamchainz: I'd recommend against pylint since it makes you do write all these hoop jumping comments

100% agree with Luke Plant's blog post here https://lukeplant.me.uk/blog/posts/pylint-false-positives/

adamchainz:
In general looking correct, presume your tests cover it

src/cfnlint/helpers.py
kddejong: I'll take a look into what this change could look like.  Thanks for the recommendation.

src/cfnlint/helpers.py
```diff
@@ -280,25 +283,58 @@ def create_rules(mod):
...
         raise os_error
 
     for root, _, filenames in os.walk(directory, onerror=onerror):
         for filename in fnmatch.filter(filenames, '[A-Za-z]*.py'):
-            pluginname = filename.replace('.py', '')
-            try:
-                fh, filename, desc = imp.find_module(pluginname, [root])
-                mod = imp.load_module(pluginname, fh, filename, desc)
+            mod = import_filename(filename.replace('.py', ''), root)
+            if mod is not None:
```
adamchainz: much nicer :)

kddejong: Thanks for the time and feedback.

Issue #1294, opened 2020-01-08
W2531 ignores resource-based metadata


tazatwell:
*cfn-lint version: 0.25.0

Hi, I have the following cfn template (just a sample, I can't provide my organization's template) named `sample-template.yaml`:

```
# Metadata:  # nodeJS8.10 EOL warning
#   cfn-lint:
#     config:
#       ignore_checks:
#       - W2531
AWSTemplateFormatVersion: '2010-09-09'
Transform: 'AWS::Serverless-2016-10-31'
Resources:
  Lambda:
    Metadata:  # nodeJS8.10 EOL warning
      cfn-lint:
        config:
          ignore_checks:
          - W2531
    Type: 'AWS::Serverless::Function'
    Properties:
      Handler: "index.handler"
      Runtime: "nodejs8.10"
```

When I run `cfn-lint sample-template.yaml`, I receive the following warning:

```
~/Documents/workspace/temp|
⇒  cfn-lint sample-template.yaml
W2531 EOL runtime (nodejs8.10) specified. Runtime is EOL since 2019-12-
31 and updating will be disabled at 2020-02-03, please consider to upda
te to nodejs10.x
sample-template.yaml:9:3
~/Documents/workspace/temp|
⇒  
```

When I un-comment the root-level Metadata, then the warning is gone:

```
Metadata:  # nodeJS8.10 EOL warning
  cfn-lint:
    config:
      ignore_checks:
      - W2531
AWSTemplateFormatVersion: '2010-09-09'
Transform: 'AWS::Serverless-2016-10-31'
Resources:
  Lambda:
    Metadata:  # nodeJS8.10 EOL warning
      cfn-lint:
        config:
          ignore_checks:
          - W2531
    Type: 'AWS::Serverless::Function'
    Properties:
      Handler: "index.handler"
      Runtime: "nodejs8.10"
```

```
~/Documents/workspace/temp|
⇒  cfn-lint sample-template.yaml
~/Documents/workspace/temp|
⇒  
```

I have no other files in my dir:

```
~/Documents/workspace/temp|
⇒  tree .
.
└── sample-template.yaml

0 directories, 1 file
~/Documents/workspace/temp|
⇒  
```

It seems as though for `W2531`, cfn-lint does not process resource-based metadata, but it _does_ process root-based metadata. I believe this is a bug.

I can provide more info on request.

Thank you!

kddejong:
[...]

PR #1297, opened 2020-01-09
Updated README with small note on resource metadata

tazatwell:
_First time contributor here, let me know if there's PR guidelines I should be aware of :)_

*Issue #, if available:*

#1294 

*Description of changes:*

Added a quick note on some metadata being lost when passed to aws-sam-translator.


tazatwell:
referenced here: #1294

PR #1307, opened 2020-01-13
Build a standalone package for offline installation

jakub-bochenski:
*Description of changes:*

A convenience for offline or automated installations on Linux or macOS.
Same as the bundle for AWS CLI, which it's based on - https://github.com/aws/aws-cli/blob/1.16.314/scripts/make-bundle

It would be useful for me to have this published officially, even if only on Github (no need to host it on S3 like the AWS CLI bundle).

```
The goal of this script is simple:
  * Create a self contained install of the CLI that
  has requires no external resources during installation.
It does this by using all the normal python tooling
(virtualenv, pip) but provides a simple, easy to use
interface for those not familiar with the python
ecosystem.
```


jakub-bochenski:
Would be nice to at least get some kind of reply. Do you even consider merging this?

kddejong:
Sorry for the delay.  Looking into this today.  

So to add on to this would we also add something to pipeline to push the packages to git?

jakub-bochenski:
Thanks!

> So to add on to this would we also add something to pipeline to push the packages to git?

Probably yes. Though I have no idea about the pipeline setup in this project, that's why I didn't do it.

kddejong:
Was able to validate this works.  Double checking all the code one more time.

PR #1316, opened 2020-01-17
Filter down objects before doing heavy computations on condition scenarios

kddejong:
*Issue #, if available:*

*Description of changes:*
- Filter down objects before running computationally heavy condition scenarios

### DIFF
--- a/src/cfnlint/__init__.py
+++ b/src/cfnlint/__init__.py
@@ -767,7 +767,7 @@ class Template(object):  # pylint: disable=R0904
 
         return result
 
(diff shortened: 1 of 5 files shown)


kddejong:
before
`cfn-lint test4.yaml  14.03s user 4.13s system 98% cpu 18.404 total`
after
`cfn-lint test4.yaml  0.96s user 0.20s system 86% cpu 1.335 total`

This isn't a full fix for possible performance issues but is a significant improvement for certain scenarios.

PR #1321, opened 2020-01-20
E6001: Fn::Transform can be used in output values

pitkley:
*Description of changes:*

cfn-lint v0.27.0 threw an error on a working template of ours, stating that `Fn::Transform` is not an allowed function for an `Output` `Value`. Transforms are actually working just fine in outputs, hence my addition of the function to the `valid_funcs` list.

Please let me know if anything is missing!

-----


pitkley:
Thanks for the quick merge, @kddejong! :+1:

kddejong:
thanks for the PR.  I'll try to get a patched version out during my afternoon.

PR #1334, opened 2020-01-29
Add python 3.8 support

chenrui333:
*Description of changes:*

Add python 3.8 support


kddejong:
This looks good to me.  Thanks for doing this.

chenrui333:
@kddejong I just found that you did the similar thing in #1289, but i guess we can take a granular step of add/delete.

kddejong:
exactly.  Completely cool with that approach.  I need to do some more planning on the 3.4 side of the world anyways.

chenrui333:
Totally agree!

kddejong:
@chenrui333 looks like an issue outside of your control.  I'll try to get a fix in shortly then get this in after.

chenrui333:
Thanks @kddejong!

Issue #1335, opened 2020-01-29
ScalingConfiguration when engine is not Aurora is an invalid property


atkinsonm:
*cfn-lint version: (`cfn-lint --version`)* 0.26.2

*Description of issue.*

Sample CloudFormation error on `Create|UpdateStack`:

You can't specify scaling configuration for non Aurora Serverless cluster. (Service: AmazonRDS; Status Code: 400; Error Code: InvalidParameterValue; Request ID: bff47f04-6bbc-492c-a74a-b46b1988f957)

It would be nice for cfn-lint to surface this attribute error so that it can be caught before the  `Create|UpdateStack` action.

chrisoverzero:
In `cfn-lint --version` 0.28.0, this test appears to be firing false failures:

```yaml
Database:
  Type: AWS::RDS::DBCluster
  Properties:
    BackupRetentionPeriod: 30
    DBClusterParameterGroupName: default.aurora-postgresql10
    DBSubnetGroupName: !Ref DatabaseSubnetGroup
    EnableHttpEndpoint: true
    Engine: aurora-postgresql
    EngineMode: serverless
    EngineVersion: '10.7'
    KmsKeyId: !Ref EncryptionKey
    MasterUsername: !Sub '{{resolve:secretsmanager:${DatabaseSecret}::username}}'
    MasterUserPassword: !Sub '{{resolve:secretsmanager:${DatabaseSecret}::password}}'
    ScalingConfiguration:
      AutoPause: false
      MaxCapacity: 4
      MinCapacity: 2
    StorageEncrypted: true
```

This resource produces E3028.

kddejong:
@chrisoverzero I think we resolved the E3028 error in this template.

kddejong:
Rule `E3028` solves this issue

PR #1338, opened 2020-01-30
E3028: cannot specify ScalingConfiguration for non Aurora Serverless AWS::RDS::DBCluster

PatMyron:
https://github.com/aws-cloudformation/cfn-python-lint/issues/1335

Like https://github.com/aws-cloudformation/cfn-python-lint/pull/1284, this rule needs to be ignored if `EngineMode` is not a literal string because the Linter cannot evaluate Parameter/Dynamic References


src/cfnlint/rules/resources/rds/AuroraScalingConfiguration.py
```diff
@@ -0,0 +1,41 @@
...
+        self.resource_property_types = ['AWS::RDS::DBCluster']
+
+    def check(self, properties, path, cfn):
+        """Check itself"""
+        matches = []
+        property_sets = cfn.get_object_without_conditions(properties)
+        for property_set in property_sets:
+            properties = property_set.get('Object')
+            if properties.get('EngineMode') != 'serverless':
+                if properties.get('ScalingConfiguration'):
```
PatMyron: @atkinsonm is this the logic you're looking for?

`AWS::RDS::DBCluster.ScalingConfiguration` can't be set when `AWS::RDS::DBCluster.EngineMode` isn't `serverless`?

atkinsonm: Yes that's correct. Thanks for the fast reply and fix!

src/cfnlint/rules/resources/rds/AuroraScalingConfiguration.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def __init__(self):
+        """Init"""
+        super(AttributeMismatch, self).__init__()
+        self.resource_property_types = ['AWS::RDS::DBCluster']
+
+    def check(self, properties, path, cfn):
+        """Check itself"""
+        matches = []
+        property_sets = cfn.get_object_without_conditions(properties)
```
kddejong: Can we change this to?
`property_sets = cfn.get_object_without_conditions(properties, ['EngineMode', 'ScalingConfiguration'])`

This will prevent some performance issues where people over condition all the properties inside this resource type.

PR #1361, opened 2020-02-17
Add initial Dockerfile

gliptak:

After this merged, I will update the docs

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,6 @@
+FROM python:3.8-alpine
+
+RUN pip install cfn-lint
+
+ENTRYPOINT ["cfn-lint"]
+CMD ["--help"]


Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM python:3.8-alpine
```
kddejong: Validating this part internally for software licensing.

gliptak: Alpine for smaller size. Could use ```3.8-slim``` (Debian Buster) instead

kddejong:
weird error... I think this may be a misconfiguration with our linting.

gliptak:
@kddejong current ```astroid``` doesn't have ```Python 2``` support

https://pypi.org/project/astroid/
https://travis-ci.org/PyCQA/astroid

gliptak:
Same for https://pypi.org/project/pylint/

gliptak:
@kddejong after rebasing PR build is green

PR #1369, opened 2020-02-18
Validate CertificateManager::Certificate::DomainValidationOptions

sbol-coolblue:
*Issue #, if available:*
N/A

*Description of changes:*
This PR introduces validation of [`AWS::CertificateManager::Certificate.DomainValidationOptions`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-certificatemanager-certificate-domainvalidationoption.html) to ensure that the ValidationDomain is a superdomain of the domain being validated.

Eg: "subdomain.example.com" can be validated by itself or "example.com", but not by "example.net".

Deployment of the stack fails if this condition is not met, so I've introduced this as an error-level rule.


PatMyron:
@sbol-coolblue running something like this will autogenerate [rule documentation](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/docs/rules.md):

```
pip install -e .
cfn-lint --update-documentation
```

sbol-coolblue:
> @sbol-coolblue running something like this will autogenerate [rule documentation](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/docs/rules.md):
> 
> ```
> pip install -e .
> cfn-lint --update-documentation
> ```

Thanks, ran it and added as an additional commit. I had wrongly assumed this command to run as part of some sort of automated pipeline.

sbol-coolblue:
Rebased to be in sync with changes to rules.md (thanks guys for adding the clarification on how to run the documentation update there!)

PatMyron:
@sbol-coolblue my main concern on this PR is gracefully handling template syntax features in these properties like functions and references. Obviously, some template syntax features like [dynamic references](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/dynamic-references.html) could never be deterministically evaluated before stack operation runtime, but I think the Linter should always err on the side of reducing noise in situations it cannot fully evaluate yet

This concern is definitely not specific to this rule and seems to broadly apply to most rules:
https://github.com/aws-cloudformation/cfn-python-lint/pull/1284
https://github.com/aws-cloudformation/cfn-python-lint/pull/1338

I'd love a way for this concern to be abstracted above the individual rule level so each rule doesn't have to re-implement handling these concerns from scratch

kddejong:
@sbol-coolblue is there a reason we only check the first element in the DomainValidationOptions list?  Looks like that list supports up to 100 values.

kddejong:
One example of that is.
`Unknown exception while processing rule E3503: can only concatenate str (not "dict_node") to str` will occur if there is anything for ValidationDomain that isn't a string.

kddejong:
I used your branch to make a few tweaks and submitted back a new PR to fix a few of those issues.  #1421

sbol-coolblue:
Thanks for picking up the branch, Kevin! Unexpected time-consuming activities at work reduced my availability for working on this issue - glad to see the fix merged!

kddejong:
No problem.  Thanks for your help with this new rule.

Issue #1249, opened 2019-12-05
Faster --update-specs


clar-cmp:
*cfn-lint version: 0.25.5*

Right now, `cfn-lint --update-specs` doesn't really do any checking to see if the existing specs are up to date and naively patches all of them. It would be beneficial to have some sort of local checksum which would prevent this, to increase overall speed when there are no updates.

PatMyron:
[using the zipped resource specifications] (https://github.com/aws-cloudformation/cfn-python-lint/pull/1148) should've helped a bit, but there's definitely more we could do. Will also look into making sure the resource specification updating happen in parallel across regions

kddejong:
@PatMyron do you know if there are any hashes for those specs that we could use to compare before downloading?

PatMyron:
> @PatMyron do you know if there are any hashes for those specs that we could use to compare before downloading?

[`aws s3api head-object`](https://docs.aws.amazon.com/cli/latest/reference/s3api/head-object.html) provides `"An ETag is an opaque identifier assigned by a web server to a specific version of a resource found at a URL."`, but I'd rather not require credentials

clar-cmp:
Honestly, I was thinking that there would be a single file that the updater would download first that lists the versions of each spec and checksums.

kddejong:
I'm curios @clar-cmp and @tro95.  Are you doing the `--update-specs` as part of CI/CD?  Are you testing your templates across all regions?  The other option for speeding this up even more may be to use the region parameter with the `--update-specs`  From my basic testing this PR does a pretty good job speeding things up thought... so just curious.

Tro95:
I am not using cfn-lint as part of any CI/CD flow, and nor do I necessarily need to check if my templates work across all regions. My main use cases are for assisting manual dev effort, and ensuring basic compatibility with all of the eu-* regions and us-east-1.

For context, slow updates don't cause me any pain. I worked on this PR simply because it was something to do on a weekend, and it seemed more interesting than writing a couple of additional rules.

PatMyron:
https://github.com/aws-cloudformation/cfn-python-lint/pull/1682 will slow down `--update-specs` by a minute or two again sometimes in order to update much more information

this should currently be infrequent since [nothing is currently updated if there are not newer Resource Specifications](https://github.com/aws-cloudformation/cfn-python-lint/pull/1554/)

PatMyron:
[...]

PR #1383, opened 2020-02-24
Checks the ETag of the files being requested to prevent downloading a…

Tro95:
Checks the ETag of the files being requested to prevent downloading and replacing files with no updates. https://github.com/aws-cloudformation/cfn-python-lint/issues/1249

*Issue #, if available:* 1249

*Description of changes:* Added caching functionality to the `get_url_content()` helper function, that stores and compares ETag values in the response headers.


src/cfnlint/helpers.py
```diff
@@ -165,10 +166,52 @@
     'AWS::Redshift::Cluster'
 ]
 
+CACHING_DIR = os.path.join(os.path.dirname(__file__), 'cache/')
```
Tro95: This can easily be changed to a different directory or moved into a config option if desired. I'm also unsure of the consequences of using `__file__` rather than `cfnlint.__file__` as is used in some places, but there doesn't seem to be a consistent approach being used, and there were circular dependency issues with the latter option.

test/unit/module/helpers/test_get_url_content.py
```diff
@@ -11,6 +11,9 @@
 from test.testlib.testcase import BaseTestCase
 from mock import patch, MagicMock
 import cfnlint.helpers
+import random
+import string
```
Tro95: These can both be removed if people were fine with hard-coding the ETag value in the tests. I just used them to generate a random string.

Tro95:
With this pull request I'm seeing a reduction from ~33 seconds to download all the specifications to ~20 seconds when none of them need redownloading. Not exactly the kind of increase I was hoping for, but still an improvement, and I'm sure someone can make it even better.

PatMyron:
> With this pull request I'm seeing a reduction from ~33 seconds to download all the specifications to ~20 seconds when none of them need redownloading. Not exactly the kind of increase I was hoping for, but still an improvement, and I'm sure someone can make it even better.

Especially as the number of regions continues to grow, I think updating the regions in parallel will be the main key to keeping `cfn-lint --update-specs` completion times within a couple seconds

https://github.com/aws-cloudformation/cfn-python-lint/blob/52864add9eac85a42cc6866cdf05c4d3b4dcd149/src/cfnlint/maintenance.py#L50

PatMyron:
> With this pull request I'm seeing a reduction from ~33 seconds to download all the specifications to ~20 seconds when none of them need redownloading. Not exactly the kind of increase I was hoping for, but still an improvement, and I'm sure someone can make it even better.

Especially as the number of regions continues to grow, I think updating the regions in parallel will be the main key to keeping `cfn-lint --update-specs` completion times within a couple seconds

https://github.com/aws-cloudformation/cfn-python-lint/blob/52864add9eac85a42cc6866cdf05c4d3b4dcd149/src/cfnlint/maintenance.py#L50

src/cfnlint/helpers.py
kddejong: Wondering if it should be under the data folder?  Would we want to commit that file as we do the `update-specs` so the most recent version is included in the package?

kddejong:
I think parallelization is a great idea here.  Another option would be to combine the region and update specs together.  Allowing for a user to filter down to the specs they care about.  By default we would have to keep the all which may make this a little trickier to implement.

src/cfnlint/helpers.py
Tro95: I'm happy for it to live under the data folder; the primary reason for using a separate cache directory was just to signify the temporary, non-essential nature of the files. It definitely could be commited and used like a lock file, and would save an initial fetch of every spec file when it's first run.

Tro95:
I'm happy to give paralleisation a go, is there any particular parellelisation module you'd prefer to use? The constant I/O of the downloads metadata file concerns me, both from a mutex point-of-view, and the inefficiency of reading and writing a whole file for a single line change. What are your thoughts on having an individual metadata file per a canonical url, i.e. `<sha256 of url>.meta.json` holding data specifically related to that url, rather than a single, centralised file?

Tro95:
I've added parallelisation and changed some basic things like the directory the caching files were being pulled into.

Updating all files speed:
```
$ time cfn-lint --update-specs

real	0m11.139s
user	0m22.762s
sys	0m2.131s
```
Subsequent update with no specs to update (as per cache check):
```
$ time cfn-lint --update-specs

real	0m5.726s
user	0m11.829s
sys	0m1.781s
```

src/cfnlint/maintenance.py
```diff
@@ -21,19 +22,37 @@
...
-        # Patch the files
-        spec = patch_spec(spec, 'all')
-        spec = patch_spec(spec, region)
+def update_resource_spec(region, url):
+    """ Update a single resource spec """
+    filename = os.path.join(os.path.dirname(cfnlint.__file__), 'data/CloudSpecs/%s.json' % region)
 
-        with open(filename, 'w') as f:
-            json.dump(spec, f, indent=2, sort_keys=True, separators=(',', ': '))
+    multiprocessing_logger = multiprocessing.log_to_stderr()
```
Tro95: One thing to note is I couldn't get logging to work inside the parallelised processes. Everything inside the parallel process is only debug logging, however it's definitely something that could be improved upon.

miparnisari: Yesterday night  [I tried out a similar implementation to yours, but using multiple threads](https://github.com/aws-cloudformation/cfn-python-lint/pull/1410), and I think logging works there.

PatMyron:
Is `cfn-lint --update-specs` still generating [`src/cfnlint/data/CloudSpecs/`](https://github.com/aws-cloudformation/cfn-python-lint/tree/master/src/cfnlint/data/CloudSpecs) changes [like the ones here](https://github.com/aws-cloudformation/cfn-python-lint/pull/1355)? I wasn't able to generate any using this branch:

```shell
$ pip3 install -e . --user
$ cfn-lint --update-specs
$ git status
On branch cfn-issue-1249
Your branch is up to date with 'origin/cfn-issue-1249'.
nothing to commit, working tree clean
```

Tro95:
It should be making changes (assuming there are any to make), so I'm unsure why it's not generating any for you.
```
$ git show
commit 706369e0e0aacbe90c150c1a7b761d880ce601c4 (HEAD -> cfn-issue-1249, dev/cfn-issue-1249)
Merge: ab46e27a be3867c6
Author: Kevin DeJong <kddejong@amazon.com>
Date:   Sat Mar 14 07:25:39 2020 -0500

    Merge branch 'master' into cfn-issue-1249

$ git status
On branch cfn-issue-1249
Your branch is up to date with 'dev/cfn-issue-1249'.

nothing to commit, working tree clean
$ pip3 install -e . --user
$ cfn-lint --update-specs
$ git status
On branch cfn-issue-1249
Your branch is up to date with 'dev/cfn-issue-1249'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git checkout -- <file>..." to discard changes in working directory)

	modified:   src/cfnlint/data/CloudSpecs/ap-east-1.json
	modified:   src/cfnlint/data/CloudSpecs/ap-northeast-1.json
	modified:   src/cfnlint/data/CloudSpecs/ap-northeast-2.json
	modified:   src/cfnlint/data/CloudSpecs/ap-northeast-3.json
	modified:   src/cfnlint/data/CloudSpecs/ap-south-1.json
	modified:   src/cfnlint/data/CloudSpecs/ap-southeast-1.json
	modified:   src/cfnlint/data/CloudSpecs/ap-southeast-2.json
	modified:   src/cfnlint/data/CloudSpecs/ca-central-1.json
	modified:   src/cfnlint/data/CloudSpecs/cn-north-1.json
	modified:   src/cfnlint/data/CloudSpecs/cn-northwest-1.json
	modified:   src/cfnlint/data/CloudSpecs/eu-central-1.json
	modified:   src/cfnlint/data/CloudSpecs/eu-north-1.json
	modified:   src/cfnlint/data/CloudSpecs/eu-west-1.json
	modified:   src/cfnlint/data/CloudSpecs/eu-west-2.json
	modified:   src/cfnlint/data/CloudSpecs/eu-west-3.json
	modified:   src/cfnlint/data/CloudSpecs/me-south-1.json
	modified:   src/cfnlint/data/CloudSpecs/sa-east-1.json
	modified:   src/cfnlint/data/CloudSpecs/us-east-1.json
	modified:   src/cfnlint/data/CloudSpecs/us-east-2.json
	modified:   src/cfnlint/data/CloudSpecs/us-gov-east-1.json
	modified:   src/cfnlint/data/CloudSpecs/us-gov-west-1.json
	modified:   src/cfnlint/data/CloudSpecs/us-west-1.json
	modified:   src/cfnlint/data/CloudSpecs/us-west-2.json

Untracked files:
  (use "git add <file>..." to include in what will be committed)

	src/cfnlint/data/DownloadsMetadata/

no changes added to commit (use "git add" and/or "git commit -a")
```

PatMyron:
@Tro95 sorry, I forgot I was testing the [homebrew](https://github.com/aws-cloudformation/cfn-python-lint/issues/1434) version recently, was able to generate changes here after uninstalling that

kddejong:
@PatMyron this looking good to merge?

src/cfnlint/helpers.py
```diff
@@ -166,10 +167,63 @@
...
+            make_second_request = False
+
+            # Request the URL with the body too
+            res = urlopen(url)
+
+        # If we have an ETag value stored and it matches the returned one,
+        # then we already have a copy of the most recent version of the
+        # resource, so don't bother fetching it again
+        if etag and res.info().get('ETag'):
+            if etag == res.info().get('ETag'):
```
miparnisari, line 208: You can combine the two `if`s.

src/cfnlint/helpers.py
```diff
@@ -166,10 +167,63 @@
...
+            # We're going to request the full body, so don't do this twice
+            make_second_request = False
+
+            # Request the URL with the body too
+            res = urlopen(url)
+
+        # If we have an ETag value stored and it matches the returned one,
+        # then we already have a copy of the most recent version of the
+        # resource, so don't bother fetching it again
+        if etag and res.info().get('ETag'):
```
miparnisari, line 207: Suggestion: rename `etag` to `cached_etag`

src/cfnlint/helpers.py
```diff
@@ -166,10 +167,63 @@
...
+            if etag == res.info().get('ETag'):
+                LOGGER.debug('We already have a cached version of url %s with ETag value of %s', url, etag)
+                return None
+
+        # If we reach here, then we don't have the latest version of the resource
+        LOGGER.debug('We do not have a cached version of url %s yet', url)
+
+        # Write the etag back to our cache
+        if res.info().get('ETag'):
+            metadata['etag'] = res.info().get('ETag')
```
miparnisari, line 217: Suggestion: rename `metadata` to `cache`

src/cfnlint/helpers.py
```diff
@@ -166,10 +167,63 @@
...
+            metadata['url'] = url # To make it obvious which url the Tag relates to
+            save_download_metadata(metadata, metadata_filename)
```
miparnisari, line 219: Suggestion: rename function to `update_cache`

src/cfnlint/helpers.py
```diff
@@ -166,10 +167,63 @@
     'AWS::Redshift::Cluster'
 ]
 
+CACHING_DIR = os.path.join(os.path.dirname(__file__), 'data', 'DownloadsMetadata')
 
-def get_url_content(url):
+
+def get_url_content(url, caching=False):
     """Get the contents of a spec file"""
```
miparnisari, line 174: To be more specific, the function returns `None` if the spec file is in the cache.

src/cfnlint/helpers.py
```diff
@@ -166,10 +167,63 @@
...
+    metadata = {}
+    encoded_url = hashlib.sha256(url.encode()).hexdigest()
+    metadata_filename = os.path.join(CACHING_DIR, encoded_url + '.meta.json')
+
+    # To help with Python2's inability to make HEAD requests
+    make_second_request = True
+
+    if caching:
+        # Load in the cache
+        metadata = get_download_metadata(metadata_filename)
```
miparnisari, line 185: Suggestion, rename function to `load_cache`

src/cfnlint/helpers.py
```diff
@@ -166,10 +167,63 @@
     'AWS::Redshift::Cluster'
 ]
 
+CACHING_DIR = os.path.join(os.path.dirname(__file__), 'data', 'DownloadsMetadata')
 
-def get_url_content(url):
+
+def get_url_content(url, caching=False):
```
miparnisari, line 173: This function is doing too many things... it's loading up the cache, checking to see if an update needs to happen, and if it does, it is downloading the file, and then updating the cache. Can we split it up into smaller chunks?

src/cfnlint/maintenance.py
```diff
@@ -21,19 +22,43 @@
...
 
+    multiprocessing_logger = multiprocessing.log_to_stderr()
+
+    multiprocessing_logger.debug('Downloading template %s into %s', url, filename)
+
+    spec_content = get_url_content(url, caching=True)
+
+    # Check to see if no content was returned, indicating we have the latest version already
+    if not spec_content:
+        multiprocessing_logger.debug('We already have the most recent version of %s stored in %s', url, filename)
```
miparnisari, line 50: Is this message necessary? `get_url_content` is already logging this

src/cfnlint/maintenance.py
```diff
@@ -21,19 +22,43 @@
...
+    """ Update a single resource spec """
+    filename = os.path.join(os.path.dirname(cfnlint.__file__), 'data/CloudSpecs/%s.json' % region)
 
+    multiprocessing_logger = multiprocessing.log_to_stderr()
+
+    multiprocessing_logger.debug('Downloading template %s into %s', url, filename)
+
+    spec_content = get_url_content(url, caching=True)
+
+    # Check to see if no content was returned, indicating we have the latest version already
```
miparnisari, line 48: This comment hints that there could be a function `should_perform_update` that returns true if the cache is old, false otherwise

src/cfnlint/data/DownloadsMetadata/1c9ead4af49b3a8f39632f5a30578ead5310da0b5a68ae4cf93b4be6a9a05278.meta.json
```diff
@@ -0,0 +1 @@
+{"etag": "\"f71b5ea17b03894692ac60518e30db24\"", "url": "https://s3.us-gov-west-1.amazonaws.com/cfn-resource-specifications-us-gov-west-1-prod/latest/gzip/CloudFormationResourceSpecification.json"}
```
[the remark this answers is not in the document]
Tro95: It makes sense to commit them, otherwise someone with the latest specs running `--update-specs` would be forced to redownload them one time to generate these files.

src/cfnlint/helpers.py
Tro95: I prefer the term metadata because what's being stored isn't the cached file itself, but metadata about the file that gets used to check if we have the latest version. There may be future use cases where we want to store additional metadata about the files that isn't related to caching.

src/cfnlint/maintenance.py
Tro95: Good call, I separated the update logic into a separate method `url_has_newer_version`.

src/cfnlint/data/DownloadsMetadata/1c9ead4af49b3a8f39632f5a30578ead5310da0b5a68ae4cf93b4be6a9a05278.meta.json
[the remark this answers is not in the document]
miparnisari: > someone with the latest specs running --update-specs would be forced to redownload them one time to generate these files

Having people download the files one more time is a small price to pay IMO. I prefer to keep the codebase clean. They'll reap the benefits later :)

src/cfnlint/helpers.py
miparnisari: Yeah, `load_metadata` sounds better

src/cfnlint/maintenance.py
miparnisari: I saw logs so I'm resolving this conversation.

src/cfnlint/data/DownloadsMetadata/1c9ead4af49b3a8f39632f5a30578ead5310da0b5a68ae4cf93b4be6a9a05278.meta.json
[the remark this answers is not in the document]
Tro95: I'll remove them from this PR, but I think we should consider adding them in future. If someone is using cfn-pythn-lint in a CI/CD flow where they pull the latest version each build, running `--update-specs` will force a download every time even though they might already have the latest.

PatMyron:
@kddejong @miparnisari if the [resource specifications](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html) haven't changed, some of the other scripts we run won't be applied without modifying [`src/cfnlint/data/DownloadsMetadata`](https://github.com/aws-cloudformation/cfn-python-lint/tree/master/src/cfnlint/data/DownloadsMetadata):
```shell
pip3 install -e .
scripts/update_specs_from_pricing.py # requires Boto3 and Credentials
cfn-lint --update-specs # won't apply changes if metadata matches
```

Tro95:
> @kddejong @miparnisari if the [resource specifications](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html) haven't changed, some of the [other data sources](https://github.com/aws-cloudformation/cfn-python-lint/pull/1480) won't be applied without modifying [`src/cfnlint/data/DownloadsMetadata`](https://github.com/aws-cloudformation/cfn-python-lint/tree/master/src/cfnlint/data/DownloadsMetadata):
> 
> ```shell
> pip3 install -e .
> scripts/update_specs_from_pricing.py # requires Boto3 and Credentials
> cfn-lint --update-specs # won't apply changes if metadata matches
> ```

@PatMyron Is the problem the fact that `--update-specs` does both the downloading of updated specs, and the merging of the specs with the patches produced by the scripts? If so, the fix should be relatively straight-forward, by replacing https://github.com/aws-cloudformation/cfn-python-lint/blob/cbf972fcb3d45dd71bdac4575e81433cbd36dcc6/src/cfnlint/maintenance.py#L46-L53 with something like 

```python
    # Check to see if we already have the latest version, otherwise download it
    if url_has_newer_version(url):
        spec_content = get_url_content(url, caching=True)
        spec = json.loads(spec_content)
    else:
        with open(filename, 'r') as f:
            spec = json.load(f)
```
This would allow the downloading of updated spec files to be skipped if there are no updates, whilst patching the specs everytime regardless. Alternatively, should we separate the patching of the specs, and include the patching with the scripts you are running?

PatMyron:
> Is the problem the fact that `--update-specs` does both the downloading of updated specs, and the merging of the specs with the patches produced by the scripts?

Yeah pretty much. `--update-specs` also merges the specs with manually written patches in addition to patches produced by the scripts

> Alternatively, should we separate the patching of the specs, and include the patching with the scripts you are running?

Need some way for manually written patches to be applied in addition to patches produced by the scripts

PR #1409, opened 2020-03-09
When using aurora dbinstance make sure certain properties aren't used

### LINKED ISSUE #1395: Check for AllocatedStorage when DBInstance engine is not Aurora
*cfn-lint version: (`cfn-lint --version`)* 0.28.2
*Using latest spec*

*Description of issue.*

`cfn-lint` does not currently warn/error when an `AWS::RDS::DBInstance` is missing the `AllocatedStorage` property. This property is required when using non-Aurora engines, and is not when using Aurora.

kddejong:
*Issue #, if available:*
fix #1395 
*Description of changes:*
- Create rule E3029 to validate that instance properties aren't  used when using a Aurora engine

### DIFF
--- a/src/cfnlint/rules/resources/rds/AuroraDBInstanceProperties.py
+++ b/src/cfnlint/rules/resources/rds/AuroraDBInstanceProperties.py
@@ -1,0 +1,66 @@
+"""
+Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+SPDX-License-Identifier: MIT-0
+"""
+import six
+from cfnlint.rules import CloudFormationLintRule
+from cfnlint.rules import RuleMatch
+
+
+class AuroraDBInstanceProperties(CloudFormationLintRule):
+    """Aurora DB instances have a lot properties that can't be set and vice and versa"""
+    id = 'E3029'
+    shortdesc = 'Aurora instances don\'t require certain properties'
+    description = 'Certain properties are not reuqired when using the Aurora engine for AWS::RDS::DBInstance'
+    source_url = 'https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-rds-database-instance.html'
+    tags = ['resources', 'rds']
+    aurora_not_required_props = [
+        'AllocatedStorage',
+        'BackupRetentionPeriod',
+        'CopyTagsToSnapshot',
+        'DeletionProtection',
+        'EnableIAMDatabaseAuthentication',
+        'MasterUserPassword',
+        'StorageEncrypted',
+    ]
+    aurora_engines = [
+        'aurora',
+        'aurora-mysql',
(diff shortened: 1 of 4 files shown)


andrewmatthews:
Where did you source this list from? According to this [AWS blog](https://aws.amazon.com/about-aws/whats-new/2019/03/copy-tags-from-an-amazon-aurora-postgresql-cluster-to-a-database-snapshot/), and the AWS Sample CloudFormation Template for [Aurora](https://github.com/aws-samples/aws-aurora-cloudformation-samples/blob/master/cftemplates/Aurora-Postgres-DB-Cluster.yml), `CopyTagsToSnapshot` is definitely supported. This merge has broken our CI linting, even though the CloudFormation stack works correctly.

kddejong:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-rds-database-instance.html#cfn-rds-dbinstance-copytagstosnapshot
`
Not applicable. Copying tags to snapshots is managed by the DB cluster. Setting this value for an Aurora DB instance has no effect on the DB cluster setting. 
`

While I will agree that Aurora supports `CopyTagsToSnapshot` its where that property is set.  For Aurora it has to be set on the cluster and not on the instance.  I've been testing some Aurora clusters and when I set the instance property it will not set it on the cluster.  So while the template will pass the desired outcome will not be reached.

neerolyte:
Any advice on fixing stacks broken by this?

If we update `cfn-lint` to a version with this check in it we have to drop the `CopyTagsToSnapshot` option from the CloudFormation to pass linting, but then updating a pre-existing stack results in a:

```
The specified DB Instance is a member of a cluster. Modify the DB engine version for the DB Cluster using the ModifyDbCluster API (Service: AmazonRDS; Status Code: 400; Error Code: InvalidParameterCombination; Request ID: ...)
```

I've checked a few times making sure no other template changes are made.

Is there a practical way to go from having it in the template to not that doesn't involve rebuilding the DB and migrating data?

P.s I ran in to this having followed https://aws.amazon.com/blogs/database/deploy-an-amazon-aurora-postgresql-db-cluster-with-recommended-best-practices-using-aws-cloudformation/ which still lists the property in the Aurora yaml:

```
$ curl -s https://s3.amazonaws.com/aws-database-blog/artifacts/aws-aurora-cloudformation-samples/cftemplates/Aurora-Postgres-DB-Cluster.yml | grep CopyTags
      CopyTagsToSnapshot: true  
      CopyTagsToSnapshot: true 
```

kddejong:
@PatMyron I may need your opinion on this.  I'm trying to replicate this issue and I cannot replicate it.  @neerolyte this was the only value that changed correct? 

@PatMyron looks like this issue was brought up here too.  https://github.com/aws-cloudformation/aws-cloudformation-coverage-roadmap/issues/211

PR #1391, opened 2020-02-29
Build a graph of resources. Use it to detect circular dependencies

miparnisari:
Changes: 

- Use `networkx` to model the template as a graph of resources
- Use this graph to check for cycles in rule E3004
- Fix #1386 by adding `DependsOn` edges to the graph

While debugging I found it very useful to use [`networkx.drawing.nx_pydot.write_dot(self.graph, path)`](https://networkx.github.io/documentation/stable/reference/generated/networkx.drawing.nx_pydot.write_dot.html#networkx.drawing.nx_pydot.write_dot) which creates a visual representation of the graph. [I think it would be useful for customers as well](https://www.reddit.com/r/aws/comments/8xdreu/can_one_generate_diagrams_from_cloudformation/). But I wasn't sure whether to include it in this PR because it needs another dependency, `pydot` and it would need some work to make it a fancy graph.

PR #1411, opened 2020-03-10
Add --build-graph option

miparnisari:
https://github.com/aws-cloudformation/cfn-python-lint/pull/1391

*Description of changes:*
Add support for `cfn-lint -t mytemplate.json -g`. If `pydot` or `pygraphviz` are installed, it will output the graph in the same directory as the template linted: `mytemplate.json.dot`, and then run the linter. If neither of those modules are found, it will print an error but the linting will continue.

TO DO: 
- [x] add test
- [x] make sure that `self._get_argument_value('build_graph', False, False)` is correct
- [x] `mytemplate.json.dot` looks ugly -> i'm marking this as done but i'm open to ideas


miparnisari:
> Not a big deal, but still creates an empty `.dot` file if those dependencies are missing 🤷‍♂️

Ah good catch... it's because of https://github.com/networkx/networkx/issues/3865

PR #1417, opened 2020-03-14
update ttl to be long

### LINKED ISSUE #1414: TTL should be of type String
*cfn-lint version: 0.28.4*

*Description of issue.*
The lint is giving me the error Property Resources/Dns/Properties/TTL should be of type String even though TTL is actually a number.

* Template linting issues:

```
Description: My Description

Resources:
  Dns:
    Type: AWS::Route53::RecordSet
    Properties:
      HostedZoneId: !ImportValue HostedZone
      Name: !Join ['', ['my.subdomain.', !ImportValue Zone, '.']]
      Type: CNAME
      ResourceRecords:
        - !ImportValue LoadBalancerDns
      TTL: 3600
```

  * Validate if the issue still exists with the latest version of `cfn-lint` and/or the latest Spec files
* Feature request:
  * Please provide argumentation about the missing feature. Context is key!

Cfn-lint uses the [CloudFormation Resource Specifications](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html) as the base to do validation. These files are included as part of the application version. Please update to the latest version of `cfn-lint` or update the spec files manually (`cfn-lint -u`)

kddejong:
*Issue #, if available:*
fix #1414 
*Description of changes:*
- Update TTL to be long as it is in the API
https://docs.aws.amazon.com/cli/latest/reference/route53/change-resource-record-sets.html


JLH993:
> _Issue #, if available:_
> fix #1414
> _Description of changes:_
> 
> * Update TTL to be long as it is in the API
>   https://docs.aws.amazon.com/cli/latest/reference/route53/change-resource-record-sets.html
> 
> By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

I have a couple of issues here that I would like to surface:

1. There are insufficient documentation updates that go with this change. 
2. I am now receiving errors while linting my previously working CF templates with records set in them. 

If this is a required change, can you update the documentation where it states this needs to be a string and nothing else? (https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-route53-recordset.html#cfn-route53-recordset-ttl) Also can examples be added? I would have to imagine I am not the only one with this issue. 

Thanks,
Jason

anilkumarmyla:
We have several of our cfn templates fail because of this, is this change correct since the official documentation claims TTL is a string?

deleugpn:
Is it possible to allow both options? I reported this issue and I would not like to change all of my templates to String since TTL is obviously a number.

kddejong:
@deleugpn this is a tricky one but the answer to your question is yes.  You can disable strict checking by configuring the rule E3012 to be `strict` `false`.  This will do some conversion for you if needed.  Example.  If the type is number and you specify a string we will make sure the string is a number.  

The hard part with this one is that it should be long in the docs and the spec.

jarreds:
We use statically typed code generated off the spec, so this was a breaking change for us. Could this tool simply check if the string value is parseable as a long if it's not a legit long?

It seems like introducing this non-backward compatible change to the spec is unlikely, but that's pure speculation.

maulik887:
Hope this change will also get reflected to Documentation- https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-route53-recordset.html#cfn-route53-recordset-ttl

Issue #903, opened 2019-05-15
Description of cfn-lint development environment and build steps


matsev:
*cfn-lint version:*
cfn-lint 0.20.1

*Description of issue.*

In the [Contributing via Pull Requests](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/CONTRIBUTING.md#contributing-via-pull-requests) paragraph, it is stated that 

> 3. Ensure local tests pass.

However, neither in the [CONTRIBUTING.md](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/CONTRIBUTING.md) nor in the [README.md](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/README.md) document is it described how one can setup the cfn-lint development environment and execute these tests.

fatbasstard:
Yeah, you're completely right, the development setup is not really described.
We'll make sure this is (better) documented. 

Do you need any help or got a specific question right now @matsev ?

matsev:
@fatbasstard My intention was to setup the build environment and play with it to get a better understanding of the linter and CloudFormation, however I got stuck before I started. I have only some Python experience and last time I tried I spent a lot of time struggling with installing runtime versions, dependency management (pip, anaconda, virtualenv) and so on before I could do anything useful. I suggest that you add a few lines about required runtime, tools (if any, preferably with a link to each homepage respectively), environment config etc to your documentation.

PatMyron:
Biggest thing that helped me was realizing local repositories can be installed using:
`pip install -e .`

PatMyron:
**At least one of these properties must be specified:**
https://github.com/aws-cloudformation/cfn-python-lint/blob/b788cc9bd3d49ed20d5f2e58602755a0ef37f52c/src/cfnlint/data/AdditionalSpecs/AtLeastOne.json#L20-L25

**Only one of these properties may be specified:**
https://github.com/aws-cloudformation/cfn-python-lint/blob/b788cc9bd3d49ed20d5f2e58602755a0ef37f52c/src/cfnlint/data/AdditionalSpecs/OnlyOne.json#L79-L84

**If this property (`SnapshotIdentifier`) is specified, these properties must be excluded:**
https://github.com/aws-cloudformation/cfn-python-lint/blob/b788cc9bd3d49ed20d5f2e58602755a0ef37f52c/src/cfnlint/data/AdditionalSpecs/Exclusive.json#L102-L107

[...]

PR #1441, opened 2020-03-25
Make DB instance engine check non case sensitive

### LINKED ISSUE #1438: E3030 case sensitivity caused DB instance destruction!
*cfn-lint version: (`cfn-lint --version`)* 0.29.0

*Description of issue.*

E3030 told me that "Postgres" was not an allowed engine name, so I changed it to "postgres". This caused cloudformation to rebuild my database instance, aka complete data loss. Luckily only discovered this on staging environment.

I think since CloudFormation / RDS seems to allow it, this check should be made case insensitive :)

kddejong:
*Issue #, if available:*
Fix #1438
*Description of changes:*
- Move DB Instance Engine check from rule E3030 to new rule E3040


PatMyron:
We might be able to switch this rule back to [concise constraints](https://github.com/aws-cloudformation/cfn-python-lint/issues/903#issuecomment-595356853) with a case insensitive pattern:

https://stackoverflow.com/questions/500864/case-insensitive-regular-expression-without-re-compile

PR #1459, opened 2020-04-06
Dependency Management - downgrade the networkx to work with Python2

### LINKED ISSUE #1458: Python2.7 dependency issue
*cfn-lint version: 0.29.4*

Hi,

When installing a project that has cfn-lint as a dependency ([moto](https://github.com/spulec/moto)) using Python2.7, I'm getting this error:
```
Searching for networkx~=2.1
Reading https://pypi.org/simple/networkx/
...
Best match: networkx 2.4
...
NetworkX 2.3+ requires Python 3.5 or later (2.7 detected).
             
For Python 2.7, please install version 2.2 using:

$ pip install 'networkx==2.2'
```

It looks like this is due to incorrect usage of the ~= operator in setup.py:
```
'networkx~=2.4;python_version>="3.5"',
'networkx~=2.1;python_version<"3.5"'
```

This semantic has the same outcome for both python versions - use any version that's 2.x, but lower than 3.x.
IMO, this should say:
```
'networkx~=2.4;python_version>="3.5"',
'networkx<=2.2;python_version<"3.5"'
```
NetworkX 2.2 is the latest version that supports Python2: https://networkx.github.io/documentation/networkx-2.2/news.html

Happy to create a PR for this, if it's an acceptable change?

bblommers:
Fixes #1458 

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -73,7 +73,7 @@ that could potentially be improved'),
         'pathlib2>=2.3.0;python_version<="3.4"',
         'importlib_resources~=1.0.2;python_version<"3.7"',
         'networkx~=2.4;python_version>="3.5"',
-        'networkx~=2.1;python_version<"3.5"'
+        'networkx<=2.2;python_version<"3.5"'
     ],
     python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*',
     entry_points={


miparnisari:
@bblommers thanks for your bug report & contribution!

PR #1472, opened 2020-04-15
Added DMS EngineName kinesis

jtheuer:
Fixes:

```
E3030 You must specify a valid value for EngineName (kinesis).
Valid values are ["aurora-postgresql", "aurora", "azuredb", "db2", "dynamodb", "mariadb", "mongodb", "mysql", "oracle", "postgres", "redshift", "s3", "sqlserver", "sybase"]
```

in

```
  DMSEventStreamEndpoint:
    Type: AWS::DMS::Endpoint
    Properties:
      EndpointIdentifier: !Sub "${AWS::StackName}-dms"
      EndpointType: target
      EngineName: kinesis
      KinesisSettings:
        StreamArn: !StreamArn
        MessageFormat: JSON
        ServiceAccessRoleArn: !RoleArn
```

See also: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-dms-endpoint.html#cfn-dms-endpoint-enginename

PR #1473, opened 2020-04-15
AWS::DMS::Endpoint.EngineName AllowedValues

PatMyron:
documentdb, elasticsearch, and kafka

https://github.com/aws-cloudformation/cfn-python-lint/pull/1472

[documentation](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-dms-endpoint.html#cfn-dms-endpoint-enginename)


kddejong:
Good catch.

PR #1483, opened 2020-04-21
New rule to check HostedZoneName is a super domain for Name

### LINKED ISSUE #1400: Check whether RecordSet name is valid
*cfn-lint version: 0.28.2*

*Description of issue.*

cfn-lint could check whether the name is valid for a given hosted zone. Specifically:

```
+    "FooBarDelegation": {
+      "Type": "AWS::Route53::RecordSet",
+      "Properties": {
+        "HostedZoneName": "bar.example.com.",
+        "Name": "foo",
+        "TTL": "86400",
+        "Type": "NS",
+        "ResourceRecords": [
...
+        ]
+      }
```

could be attempted, which results in a not very clear error:

```
RRSet with DNS name foo. is not permitted in zone bar.example.com.
```

The solution is to have name `foo.bar.example.com` instead.
This came up in a highly upvoted SO question as well: https://stackoverflow.com/questions/11420063/rrset-with-dns-name-foo-is-not-permitted-in-zone-bar

It would be great if (when the values are available), cfn-lint raised an error "Name has to end with HostedZoneName"

kddejong:
*Issue #, if available:*
Fix #1400 
*Description of changes:*
- Add rule E3040 to check if HostedZoneName is a super domain for Name

### DIFF
--- a/src/cfnlint/rules/resources/route53/RecordSetName.py
+++ b/src/cfnlint/rules/resources/route53/RecordSetName.py
@@ -1,0 +1,60 @@
+"""
+Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+SPDX-License-Identifier: MIT-0
+"""
+import six
+from cfnlint.rules import CloudFormationLintRule
+from cfnlint.rules import RuleMatch
+
+
+class RecordSetName(CloudFormationLintRule):
+    """Check if a Route53 Resoruce Records Name is valid with a HostedZoneName"""
+    id = 'E3040'
+    shortdesc = 'RecordSet HostedZoneName is a superdomain of Name'
+    description = 'In a RecordSet, the HostedZoneName must be a superdomain of the Name being validated'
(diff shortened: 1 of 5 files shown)


test/unit/rules/resources/route53/test_recordset_name.py
```diff
@@ -0,0 +1,27 @@
+"""
+Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
PatMyron, line 2:
```suggestion
Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```

[can remove this now](https://github.com/aws-cloudformation/cfn-python-lint/pull/1528)

src/cfnlint/rules/resources/route53/RecordSetName.py
```diff
@@ -0,0 +1,59 @@
+"""
+Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
PatMyron, line 2:
```suggestion
Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```

[can remove dates now](https://github.com/aws-cloudformation/cfn-python-lint/pull/1528)

PR #1488, opened 2020-04-23
Update exception list for appsync needs

### LINKED ISSUE #1485: False Error
0.29.5

*Description of issue.*
I'm getting the following error in a [`RequestMappingTemplate`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-appsync-resolver.html#cfn-appsync-resolver-requestmappingtemplate) on [`AWS::AppSync::Resolver`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-appsync-resolver.html). I believe this should be an exception to the rule, as it's completely normal to have `${}` inside the mapping template.

>Found an embedded parameter "${context.arguments.clientId}" outside of an "Fn::Sub" at Resources/ListMemberAttributeDefinitionsResolver/Properties/RequestMappingTemplate

Example Resource:
```
  Resolver:
    Type: AWS::AppSync::Resolver
    Properties:
      ApiId: !ImportValue AppSyncApiId
      TypeName: Query
      Kind: UNIT
      FieldName: myFieldName
      DataSourceName: !GetAtt AppSyncDataSource.Name
      RequestMappingTemplate: |-
        {
            "version": "2018-05-29",
            "method": "GET",
            "params": {
                "headers": $utils.toJson($util.map.copyAndRemoveAllKeys($context.request.headers, ['host','connection','user-agent','expectation','transfer-encoding','content-length'])),
            },
            "resourcePath": "/some-path/${context.arguments.someId}"
        }
      ResponseMappingTemplate: |-
        ## return the body
        #if($context.result.statusCode == 200)
             $ctx.result.body
        #else
            ##if response is not 200, append the response to error block.
            $util.error($context.result.body, "$context.result.statusCode")
        #end
```

kddejong:
*Issue #, if available:*
fix #1485 
*Description of changes:*
- Update exceptions for rule E1029 to include RequestMappingTemplate

### DIFF
--- a/src/cfnlint/rules/functions/SubNeeded.py
+++ b/src/cfnlint/rules/functions/SubNeeded.py
@@ -18,8 +18,10 @@ class SubNeeded(CloudFormationLintRule):
 
     # Free-form text properties to exclude from this rule
     # content is part of AWS::CloudFormation::Init
+    # RequestMappingTemplate is because of issue #1485
     excludes = ['UserData', 'ZipFile', 'Condition', 'AWS::CloudFormation::Init',
-                'CloudWatchAlarmDefinition', 'TopicRulePayload', 'BuildSpec']
+                'CloudWatchAlarmDefinition', 'TopicRulePayload', 'BuildSpec',
+                'RequestMappingTemplate']
     api_excludes = ['Uri', 'Body']
 
     # IAM Policy has special variables that don't require !Sub, Check for these


Tro95:
Should this also include `ResponseMappingTemplate`?

PatMyron:
> Should this also include `ResponseMappingTemplate`?

covered in https://github.com/aws-cloudformation/cfn-python-lint/pull/1667

PR #1506, opened 2020-05-01
Add JUnit XML Report Format

### LINKED ISSUE #1368: --format JUnit XML reporting file format
**Feature request** 
Would be really handle if we can get some more formatting options for popular test frameworks, for example JUnit XML. 

These frameworks integrate into many build pipeline technologies for an integrated test reporting.

As it stands cfn-python-lint runs as a scripted step and will exit with errors if there are any structural issues with the templates. ideally i would like the output to be formatted into JUnit XML with results of each tested resources. this file is picked up by our build framework and the pipline will pass or fail based on the results of the test run, the same way we do for code tests.

Tro95:
*Issue #, if available:* #1368 

*Description of changes:* Added a `--format=junit` option, which produces an XML report in the style of JUnit, ideal for CI/CD systems. The main difference between this format and the existing ones is that this format includes successful and skilled tests. Closes #1368  


Tro95:
Screenshot 2020-05-01 at 10 14 04
Screenshot 2020-05-01 at 10 14 14

Example integration with CodeBuild's report groups.

src/cfnlint/formatters/__init__.py
```diff
@@ -40,6 +43,56 @@ def _format(self, match):
...
+                )
+                for match in matches:
+                    if match.rule.id == rule.id:
+                        test_case.add_failure_info(
+                            message=self._failure_format(match),
+                            failure_type=match.message
+                        )
+                test_cases.append(test_case)
+
+        test_suite = TestSuite('Cloudformation Lint', test_cases)
```
kddejong, line 91: Sorry, super picky here.  Should this be `CloudFormation`?

Tro95:
I tested that disabling specific rules in the template does not make the test show up as skipped in the report, it will simply not evaluate the specific lines against the rule and therefore no failures within the disabled section will be in the report.

Currently the XML produced will contain multiple `<failure>` tags for any rule that matches more than once. This is unlikely to cause breaking problems for consuming systems, however not every consuming system will leverage all the information, and will instead only surface the first `<failure>` tag. If this is problematic for people, we can look to either combine all failures for the same rule into a single `<failure>` tag (not preferred, due to the size and formating of this), or introduce a `--format=junit-single` option that permits people to select the former option.

In the future I think it would be nice to not have to require the rules to be passed into the formatter, but instead for the cfnlint runner to return a set of rules rather than matches with information on whether the rule was run, and any time it matched. This would allow timing information to be surfaced, which currently can't be done. Other formatters would obviously need to be adjusted to only report on rules with matching results.

setup.py
```diff
@@ -53,7 +53,8 @@ def get_version(filename):
         'importlib_resources~=1.0.2;python_version=="3.4"',
         'importlib_resources~=1.4;python_version<"3.7" and python_version!="3.4"',
         'networkx~=2.4;python_version>="3.5"',
-        'networkx<=2.2;python_version<"3.5"'
+        'networkx<=2.2;python_version<"3.5"',
+        'junit-xml~=1.9',
```
PatMyron, line 57: @miparnisari need to vet [this new dependency](https://pypi.org/project/junit-xml/)

kddejong: License info for `junit-xml`
- Freely Distributable
- OSI Approved :: MIT License

PR #1511, opened 2020-05-05
Output File CLI Option

Tro95:
*Issue #, if available:* None

*Description of changes:* Add `--output-file` argument, that writes the main output to the file rather than stdout, whilst allowing exceptions and other text to still be written to stdout/stderr. This is useful when producing test reports for consumption by CI/CD systems.


PatMyron:
@Tro95 is everything currently being written to stdout? Think it'd make sense to utilize stderr if so

Tro95:
@PatMyron I can only see two places where something is directly outputted to stdout, which is https://github.com/aws-cloudformation/cfn-python-lint/blob/c32ce78e0b5e87b67f3ea4ddb43d8d1954dfc4ef/src/cfnlint/core.py#L142 and https://github.com/aws-cloudformation/cfn-python-lint/blob/c32ce78e0b5e87b67f3ea4ddb43d8d1954dfc4ef/src/cfnlint/__main__.py#L33
Everything else is handled by the logger, which [uses stderr by default for all levels](https://docs.python.org/3/library/logging.handlers.html). You should be able to test this PR by trying `cfn-lint -t template.yml -D --output-file output.txt`, and seeing the debug information displayed in the console via `stderr` whilst the results of the linting are written to `output.txt`

kddejong:
The value of this is to help with CI/CD tooling. A lot of these tools will support "uploading" junit reports to more easily show the test results.  I'm in agreement that this should be an option.  @PatMyron any other concerns before I merge this one?

PR #1517, opened 2020-05-06
tagging rules

PatMyron:
(also removed newlines in some descriptions)

[`docs/rules.md`](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/docs/rules.md) changes generated with `cfn-lint --update-documentation`

sidenote: [`W1001`](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/rules/functions/RelationshipConditions.py) [`W3005`](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/rules/resources/DependsOnObsolete.py) [`E3008`](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/rules/resources/properties/ValueRefGetAtt.py) could add [`Sub`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-sub.html) support in the future


src/cfnlint/rules/resources/properties/Password.py
```diff
@@ -13,9 +13,9 @@ class Password(CloudFormationLintRule):
     """Check if Password Properties are properly configured"""
     id = 'W2501'
     shortdesc = 'Check if Password Properties are correctly configured'
-    description = 'Password properties should be strings and if parameter using NoEcho'
+    description = 'Password properties should not be strings and if parameter using NoEcho'
```
miparnisari, line 16: Nice find!

miparnisari:
> [`W3005`](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/rules/resources/DependsOnObsolete.py)  could add [`Sub`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-sub.html) support in the future

it could also be simplified using the graph 😄

PR #1480, opened 2020-04-17
CloudFormation Resource Specification 13.0.0

PatMyron:
```shell
pip3 install -e .
scripts/update_specs_from_pricing.py # requires Boto3 and Credentials
scripts/update_specs_services_from_ssm.py # requires Boto3 and Credentials
cfn-lint --update-specs
cfn-lint --update-iam-policies
```

as described in https://github.com/aws-cloudformation/cfn-python-lint/pull/1134#issuecomment-536184503

---

```shell
scripts/update_specs_services_from_ssm.py # requires Boto3 and Credentials
Traceback (most recent call last):
  File "scripts/update_specs_services_from_ssm.py", line 9, in <module>
    from cfnlint.helpers import get_url_content
  File "cfn-python-lint/src/cfnlint/__init__.py", line 17, in <module>
    from cfnlint.graph import Graph
  File "cfn-python-lint/src/cfnlint/graph.py", line 10, in <module>
    from networkx import networkx
ImportError: No module named networkx
```

PR #1524, opened 2020-05-10
Refactor: move all the spec-related code to one file

miparnisari:
*Issue #, if available:* N/A

*Description of changes:* Refactor: move all the spec-related code to one file. Ideally they would be in a new class too but.. baby steps :)


test/unit/module/specs/test_complete.py
```diff
@@ -42,9 +41,7 @@ def test_fail_run(self):
         filename = 'test/fixtures/templates/bad/override/complete.yaml'
         template = self.load_template(filename)
 
-        with open('test/fixtures/templates/override_spec/complete.json') as fp:
-            custom_spec = json.load(fp)
-        cfnlint.helpers.set_specs(custom_spec)
+        cfnlint.specs.override_specs('test/fixtures/templates/override_spec/complete.json')
```
PatMyron, line 44: nice catch with these 👍

PatMyron:
didn't think the [new region documentation](https://github.com/aws-cloudformation/cfn-python-lint/pull/1496) would get outdated this quickly 😜

src/cfnlint/maintenance.py
```diff
@@ -2,63 +2,19 @@
...
-from cfnlint.helpers import SPEC_REGIONS
+from cfnlint.specs import get_url_content
 import cfnlint.data.ExtendedSpecs
 import cfnlint.data.AdditionalSpecs
 
 
 LOGGER = logging.getLogger(__name__)
 
 
-def update_resource_specs():
```
PatMyron, line 22: thought it made sense and helped figuring out the [~weekly script](https://github.com/aws-cloudformation/cfn-python-lint/pull/1480#issuecomment-619308734) that they were all in this file or the [`scripts/`](https://github.com/aws-cloudformation/cfn-python-lint/tree/master/scripts) directory

miparnisari: Yeah I imagined that was why the code was placed here. But it's not very object-oriented this way. I think it could be better if we documented the process:

```bash
pip3 install -e .
scripts/update_specs_from_pricing.py # requires Boto3 and Credentials
scripts/update_specs_services_from_ssm.py # requires Boto3 and Credentials
cfn-lint --update-specs
cfn-lint --update-iam-policies
cfn-lint --update-documentation
```

BTW the last 2 commands are not explained in the readme's command line options, is that intended?

kddejong: Not really, there were at one point documented.  They ended up being considered to be more admin/maintainer style commands so they ended up being undocumented.  So many commands... `--update-all`?

miparnisari: > So many commands... --update-all?

But customers won't care about `--update-documentation`...

src/cfnlint/helpers.py
```diff
@@ -7,65 +7,26 @@
...
+
+
 if sys.version_info < (3,):
     import imp
 else:
     import importlib  # pylint: disable=ungrouped-imports
 
 LOGGER = logging.getLogger(__name__)
 
-SPEC_REGIONS = {
```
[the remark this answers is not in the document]
miparnisari: Not sure how to put `@refactored` around a variable 🤔

kddejong:
This works for Python 3 but Python 2 users are out of luck.

```
from cfnlint.specs import RESOURCE_SPECS as _deprecated_RESOURCE_SPECS

deprecated_names = {
    'RESOURCE_SPECS': 'is refactored and deprecated. Please use cfnlint.specs.RESOURCE_SPECS'
}


def __getattr__(name):
    if name in deprecated_names:
        warn("%s %s" % (name, deprecated_names[name]), FutureWarning)
        return globals()["_deprecated_%s" % name]
    raise AttributeError("module %s has no attribute %s" % (__name__, name))
```

kddejong:
@miparnisari should we just let my feedback go?  I don't mind as I thought it was easier to do that but you are correct for Parameters its a little confusing. 

Looks like we have a merge conflict that needs to get cleaned up and then we can get this in.

src/cfnlint/maintenance.py
miparnisari: Maybe we can have a Github action that runs on every PR? It would run `--update-documentation` and if it detected changes it would submit another PR

miparnisari:
@PatMyron thoughts on this?

PR #1533, opened 2020-05-15
New rule to check resources if ARNs use correctly placed Pseudo Parameters instead of hardcoded Partition, Region, and Account Number

dontirun:
*Description of changes:*
Adding Rule W3042 and related testcases. Checks Resources if ARNs use correctly placed Pseudo Parameters instead of hardcoded Partition, Region, and Account Number


dontirun:
The linting is failing on characters that need to be escaped for the regex expression to process properly

src/cfnlint/rules/resources/HardCodedArnProperties.py
```diff
@@ -0,0 +1,34 @@
...
+    id = 'W3042'
+    shortdesc = 'ARNs should use correctly placed Pseudo Parameters'
+    description = 'Checks Resources if ARNs use correctly placed Pseudo Parameters instead of hardcoded Partition, Region, and Account Number'
+    source_url = ''
+    tags = ['resources']
+
+    def match(self, cfn):
+        """Check CloudFormation Resources"""
+        matches = []
+        pattern = re.compile('arn:(\$\{[^:]*::[^:]*}|[^:]*):[^:]+:(\$\{[^:]*::[^:]*}|[^:]*):(\$\{[^:]*::[^:]*}|[^:]*)')
```
Tro95, line 21: Try using a raw string by adding `r` before the open quotes, i.e. `re.compile(r'arn:(\$\{[^:]*::[^:]*}|[^:]*):[^:]+:(\$\{[^:]*::[^:]*}|[^:]*):(\$\{[^:]*::[^:]*}|[^:]*)')`.

src/cfnlint/rules/resources/HardCodedArnProperties.py
```diff
@@ -0,0 +1,34 @@
...
+SPDX-License-Identifier: MIT-0
+"""
+import re
+from cfnlint.rules import CloudFormationLintRule
+from cfnlint.rules import RuleMatch
+
+
+class HardCodedArnProperties(CloudFormationLintRule):
+    """Checks Resources if ARNs use correctly placed Pseudo Parameters instead of hardcoded Partition, Region, and Account Number"""
+    id = 'W3042'
```
PatMyron, line 12: so I've been writing a warning-level rule for hardcoded partitions because partitions cannot communicate with each other but there are valid reasons for using other regions and account IDs

maybe this rule should be informational-level?

src/cfnlint/rules/resources/HardCodedArnProperties.py
```diff
@@ -0,0 +1,34 @@
+"""
+Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
PatMyron, line 2:
```suggestion
Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
```

[can remove dates now](https://github.com/aws-cloudformation/cfn-python-lint/pull/1528)

dontirun:
What are the next steps for this rule?

src/cfnlint/rules/resources/HardCodedArnProperties.py
PatMyron: > Given the same case it wouldn't make sense to throw a warning on a hardcoded Partition either

partitions cannot communicate with each other, so I don't think hardcoded partitions are ever necessary, whereas cross-account and cross-region communication is a valid use-case, so I think hardcoded accounts and regions should probably only be informational. [Separate from this rule specifically, I would like informational rules to start showing by default without affecting the exit code](https://github.com/aws-cloudformation/cfn-python-lint/issues/1586#issuecomment-642984414) and only affect the exit code if opted into `--include-checks I`

src/cfnlint/rules/resources/HardCodedArnProperties.py
```diff
@@ -0,0 +1,35 @@
...
+SPDX-License-Identifier: MIT-0
+"""
+import re
+from cfnlint.rules import CloudFormationLintRule
+from cfnlint.rules import RuleMatch
+
+
+class HardCodedArnProperties(CloudFormationLintRule):
+    """Checks Resources if ARNs use correctly placed Pseudo Parameters instead of hardcoded Partition, Region, and Account Number"""
+    id = 'W3042'
```
mvp-cloud, line 12: `   id = 'I3042'`

PR #1539, opened 2020-05-18
Add coc.nvim integration for (Neo)Vim to README

joenye:
*Description of changes:*

Adds an additional (Neo)Vim integration via [Coc](https://github.com/neoclide/coc.nvim) to the README: [coc-cfn-lint](https://github.com/joenye/coc-cfn-lint).


PatMyron:
can you elaborate on the difference between `ALE`, `Coc`, and `Syntastic`? not super familiar myself

joenye:
Sure! 

`Coc` gives (Neo)Vim full [Language Server](https://langserver.org/) support (think the same functionality as VSCode in terms of code completion, linting, auto-formatting, etc.), whereas `ALE` and `Syntastic` require extension authors to write `ALE`-specific or `Syntastic`-specific code every time.

Here's a nice visualisation of the benefits of No LSP (`ALE`, `Syntastic`) vs. LSP (`Coc`):

image.

What `ALE` offers over `Syntastic` is it uses (Neo)Vim's "async" API, which means it doesn't block while linting. Generally, `ALE` is regarded as superior/more modern than `Syntastic` for this reason.

`Coc` is the cool new kid on the block, and has quickly become very popular (it already has more stars than Syntastic and ALE). The reason for its popularity is it addresses the above "many-to-many" problem and provides full LSP client support. This means Neo(Vim) users benefit from a massive range of [extensions](https://github.com/neoclide/coc.nvim/wiki/Using-coc-extensions) - many of which are forked from VSCode extensions (which are also LSP clients). Being an LSP client, `Coc` extensions can provide far more than linting, so for Neo(Vim) users it's "one plugin to rule them all" in terms of code completion, linting, auto-formatting, etc.

In this case, the `coc-cfn-lint` extension only supports linting. If, for example, `cfn-lint` were to add auto-formatting in the future, then I could update the extension to allow `Coc` users to trigger auto-formatting in a familiar way.

PR #1546, opened 2020-05-20
Remove usage of jsonpointer

miparnisari:
*Issue #, if available:* N/A

*Description of changes:* Remove usage of [jsonpointer](https://pypi.org/project/jsonpointer/)

### DIFF
--- a/src/cfnlint/maintenance.py
+++ b/src/cfnlint/maintenance.py
@@ -7,7 +7,6 @@ import json
 import logging
 import multiprocessing
 import os
-import jsonpointer
 import jsonpatch
 import cfnlint
 from cfnlint.helpers import get_url_content, url_has_newer_version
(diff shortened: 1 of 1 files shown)


src/cfnlint/maintenance.py
```diff
@@ -155,7 +154,7 @@ def patch_spec(content, region):
                     jsonpatch.JsonPatch([all_patch]).apply(content, in_place=True)
                 except jsonpatch.JsonPatchConflict:
                     LOGGER.debug('Patch (%s) not applied in region %s', all_patch, region)
-                except jsonpointer.JsonPointerException:
+                except jsonpatch.JsonPointerException:
```
kddejong, line 157: huh, weird that the exception lists switched to `jsonpointer`.  Am I missing something here?  How did you catch this issue?

miparnisari: I just went through the list of dependencies in Scallion and noticed this package was there but in code it only appeared here, and it's not listed in setup.py

kddejong: Wish I would have written a test for this exception

miparnisari: It is covered by tests though. I am not sure why they passed 🤔

kddejong: I guess that is a good sign.  I was having troubles finding this definition in jsonpatch but it must be as it works.

PR #1554, opened 2020-05-29
--update-specs even if official CloudFormation Resource Specification hasn't changed

PatMyron:
https://github.com/aws-cloudformation/cfn-python-lint/pull/1383#issuecomment-629891506

[`cfn-lint --update-specs`](https://github.com/aws-cloudformation/cfn-python-lint/blob/cc6ac28ff7deba86cb82813733cceec4bdff68a2/src/cfnlint/maintenance.py#L142-L150) patches [`ExtendedSpecs`](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/data/ExtendedSpecs) and [`botocore` data](https://github.com/aws-cloudformation/cfn-python-lint/pull/1682) into [`CloudSpecs`](https://github.com/aws-cloudformation/cfn-python-lint/tree/master/src/cfnlint/data/CloudSpecs) in addition to newly downloaded [official CloudFormation Resource Specifications](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html)

[`ExtendedSpecs/$REGION/05_pricing_property_values.json`](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/data/ExtendedSpecs/all/05_pricing_property_values.json) written by [`scripts/update_specs_from_pricing.py`](https://github.com/aws-cloudformation/cfn-python-lint/blob/cc6ac28ff7deba86cb82813733cceec4bdff68a2/scripts/update_specs_from_pricing.py#L235)
[`ExtendedSpecs/$REGION/06_ssm_service_removal.json`](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/data/ExtendedSpecs/us-gov-east-1/06_ssm_service_removal.json) written by [`scripts/update_specs_services_from_ssm.py`](https://github.com/aws-cloudformation/cfn-python-lint/blob/cc6ac28ff7deba86cb82813733cceec4bdff68a2/scripts/update_specs_services_from_ssm.py#L165)
[`ExtendedSpecs/$REGION/07_ssm_service_addition.json`](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/data/ExtendedSpecs/us-gov-east-1/07_ssm_service_addition.json) written by [`scripts/update_specs_services_from_ssm.py`](https://github.com/aws-cloudformation/cfn-python-lint/blob/cc6ac28ff7deba86cb82813733cceec4bdff68a2/scripts/update_specs_services_from_ssm.py#L204)

the rest of [`ExtendedSpecs`](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/data/ExtendedSpecs) is hand maintained

### DIFF
--- a/src/cfnlint/maintenance.py
+++ b/src/cfnlint/maintenance.py
@@ -10,7 +10,7 @@ import os
 import jsonpointer
 import jsonpatch
 import cfnlint
-from cfnlint.helpers import get_url_content, url_has_newer_version
+from cfnlint.helpers import get_url_content
 from cfnlint.helpers import SPEC_REGIONS
 import cfnlint.data.ExtendedSpecs
 import cfnlint.data.AdditionalSpecs
@@ -44,8 +44,9 @@ def update_resource_spec(region, url):
     multiprocessing_logger.debug('Downloading template %s into %s', url, filename)
 
     # Check to see if we already have the latest version, and if so stop
-    if not url_has_newer_version(url):
-        return
+    # commented out for now due to https://github.com/aws-cloudformation/cfn-python-lint/pull/1383#issuecomment-629891506
+    # if not url_has_newer_version(url):
+    #     return
 
     spec_content = get_url_content(url, caching=True)
 


src/cfnlint/maintenance.py
```diff
@@ -44,8 +44,9 @@ def update_resource_spec(region, url):
     multiprocessing_logger.debug('Downloading template %s into %s', url, filename)
 
     # Check to see if we already have the latest version, and if so stop
-    if not url_has_newer_version(url):
-        return
+    # commented out for now due to https://github.com/aws-cloudformation/cfn-python-lint/pull/1383#issuecomment-629891506
+    # if not url_has_newer_version(url):
+    #     return
```
Tro95, line 49: It would be better to keep the `url_has_newer_version()` call, but instead of returning from the function continue on instead, just skipping the downloads. This would let you do the patching regardless of whether new external specs are available or not.

```python
    # Check to see if we already have the latest version, otherwise download it
    if url_has_newer_version(url):
        spec_content = get_url_content(url, caching=True)
        spec = json.loads(spec_content)
    else:
        with open(filename, 'r') as f:
            spec = json.load(f)
```

Tests will likely need to be adjusted.

PatMyron: > It would be better to keep the `url_has_newer_version()` call, but instead of returning from the function continue on instead, just skipping the downloads. This would let you do the patching regardless of whether new external specs are available or not.

Agreed, just using this in the meantime

PatMyron: @Tro95 maybe just an undocumented `--force` flag for now that skips this `url_has_newer_version()` check?
```bash
cfn-lint --update-specs --force
```

Tro95: That seems like a good short-term solution

PatMyron: [just added a bit more to `--update-specs` beyond the official Resource Specifications here as well](https://github.com/aws-cloudformation/cfn-python-lint/pull/1682)

kddejong:
Went ahead and created a force option.  Please take a look #2334

PR #1271, opened 2019-12-22
rules for approaching limits

PatMyron:
https://github.com/aws-cloudformation/cfn-python-lint/issues/1092

Didn't get around to `E3502`:
https://github.com/aws-cloudformation/cfn-python-lint/blob/cec8855bea20c04f866a1fd9aed152d9ce3d3232/src/cfnlint/rules/resources/properties/JsonSize.py#L17

@kddejong thoughts on Informational-level and 90% of the limit values?
Those are the parts I was least sure about

---

manually tested using modified versions of the limit templates here:
https://github.com/aws-cloudformation/cfn-python-lint/tree/master/test/fixtures/templates/bad

```
gsed -i 's/AndNowItIs//' limit_name.yaml
gsed -i 's/ Sed rhoncus eros eget purus accumsan cras amet.//' limit_parameter_value.yaml
gsed -i 's/4097/4095/' limit_parameter_value.yaml
cfn-lint limit_name.yaml limit_parameter_value.yaml approaching_limit_* -c I -i W1011 W1020 W2001
```

```
I2011 The length of parameter name (254) is approaching the limit (255)
limit_name.yaml:4:3
I7011 The length of mapping name (250) is approaching the limit (255)
limit_name.yaml:7:3
I3011 The length of resource name (252) is approaching the limit (255)
limit_name.yaml:12:3
I6011 The length of output name (248) is approaching the limit (255)
limit_name.yaml:18:3
I2012 The length of parameter default value (4055) is approaching the limit (4096)
limit_parameter_value.yaml:6:5
I2012 The MaxLength of parameter (4095) is approaching the limit (4096)
limit_parameter_value.yaml:21:5
I2012 The length of parameter allowed value (4055) is approaching the limit (4096)
limit_parameter_value.yaml:24:5
I2010 The number of parameters (60) is approaching the limit (60)
approaching_limit_numbers.yaml:4:1
I7010 The number of mappings (100) is approaching the limit (100)
approaching_limit_numbers.yaml:125:1
I7012 The amount of mapping attributes (64) is approaching the limit (64)
approaching_limit_numbers.yaml:129:5
I3010 The number of resources (200) is approaching the limit (200)
approaching_limit_numbers.yaml:689:1
I6010 The number of outputs (60) is approaching the limit (60)
approaching_limit_numbers.yaml:1890:1
I1002 The template file size (430425 bytes) is approaching the limit (460800 bytes)
approaching_limit_size.yaml:1:1
[...]

PR #1558, opened 2020-05-30
refactoring limit rules

PatMyron:
https://github.com/aws-cloudformation/cfn-python-lint/pull/1271

extracted 14+ of the 26 main limit rules into [`src/cfnlint/rules/limits.py`](https://github.com/aws-cloudformation/cfn-python-lint/pull/1558/files#diff-d7e9d56f75b110978061a5cbc7f742af)

---

```bash
# limit test templates
find test -name "*limit*" | grep -v '.py' | grep -v 'events'
test/fixtures/templates/bad/limit_name.yaml
test/fixtures/templates/bad/limit_parameter_value.yaml
test/fixtures/templates/bad/limit_numbers.yaml
test/fixtures/templates/bad/approaching_limit_numbers.yaml
test/fixtures/templates/bad/approaching_limit_size.yaml
test/fixtures/templates/bad/limit_size.yaml

# ignoring all their unrelated W1011, W1020, W2001 violations
find test -name "*limit*" | grep -v '.py' | grep -v 'events' | xargs cfn-lint -c I -t | grep -Eo '^\w\d+' | sort | uniq -c | sort -nr
1426 W1011
 936 W1020
   3 W2001
   3 E2012
   2 I7010
   2 I6010
   2 I2010
   2 E7010
   2 E6010
   2 E3010
   2 E2010
   1 I7012
   1 I6012
   1 I3010
   1 I1003
   1 I1002
   1 E7012
   1 E7011
   1 E6011
   1 E3011
   1 E2011
   1 E1003
   1 E1002

# same output when run with previous version and this version before the last casing commit
find test -name "*limit*" | grep -v '.py' | grep -v 'events' | xargs cfn-lint -i W1011 W1020 W2001 -c I -t

# testing informational level rules
gsed -i 's/AndNowItIs//' test/fixtures/templates/bad/limit_name.yaml
gsed -i 's/ Sed rhoncus eros eget purus accumsan cras amet.//' test/fixtures/templates/bad/limit_parameter_value.yaml
gsed -i 's/4097/4095/' test/fixtures/templates/bad/limit_parameter_value.yaml

# same output when run with previous version and this version before the last casing commit
find test -name "*limit*" | grep -v '.py' | grep -v 'events' | xargs cfn-lint -i W1011 W1020 W2001 -c I -t
```


src/cfnlint/rules/resources/ApproachingLimitNumber.py
```diff
@@ -17,13 +17,9 @@ class LimitNumber(CloudFormationLintRule):
 
     def match(self, cfn):
         """Check CloudFormation Resources"""
-
         matches = []
-
-        # Check number of resources against the defined limit
         resources = cfn.get_resources()
```
PatMyron, line 21: wary about switching these from [`get_resources()`](https://github.com/aws-cloudformation/cfn-python-lint/blob/b34913e1186edcf23fd4042d2d197b76c762d708/src/cfnlint/template.py#L63) to [`get('Resources', {})`](https://github.com/aws-cloudformation/cfn-python-lint/blob/b34913e1186edcf23fd4042d2d197b76c762d708/src/cfnlint/decode/node.py#L108) like the rest of these rules

src/cfnlint/rules/resources/LimitNumber.py
```diff
@@ -17,13 +17,9 @@ class LimitNumber(CloudFormationLintRule):
 
     def match(self, cfn):
         """Check CloudFormation Resources"""
-
         matches = []
-
-        # Check number of resources against the defined limit
         resources = cfn.get_resources()
```
PatMyron, line 21: wary about switching these from [`get_resources()`](https://github.com/aws-cloudformation/cfn-python-lint/blob/b34913e1186edcf23fd4042d2d197b76c762d708/src/cfnlint/template.py#L63) to [`get('Resources', {})`](https://github.com/aws-cloudformation/cfn-python-lint/blob/b34913e1186edcf23fd4042d2d197b76c762d708/src/cfnlint/decode/node.py#L108) like the rest of these rules

src/cfnlint/helpers.py
```diff
@@ -135,22 +135,22 @@
...
         'description': 1024  # in bytes
     },
-    'parameters': {
+    'Parameters': {
         'number': 60,
         'name': 255,  # in characters
         'value': 4096  # in bytes
     },
-    'resources': {
+    'Resources': {
```
PatMyron, line 153: matching [template section](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-anatomy.html) casing

src/cfnlint/rules/resources/LimitNumber.py
kddejong: I like `get_resources()` for these scenarios.  The idea being that we are validating that the basic construct of the resource is valid.  Its of type 'dict' for instance.  Ideally, that function could be built out even more.  Implement caching so we aren't checking that every time. Also check that Type is there no matter what. 

https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/template.py#L63-L81

kddejong:
I really like this.  The other idea is that there is also idea that you can keep embedding classes.

We did this for the Lambda EOL checks.  
https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/rules/resources/lmbd/DeprecatedRuntimeEnd.py#L10

Not sure if that would make life easier or just a different way to get to the same result or if I'm just way off.

src/cfnlint/rules/resources/LimitNumber.py
PatMyron: @kddejong [got these last couple rules refactored using `get_resources()` here](https://github.com/aws-cloudformation/cfn-python-lint/pull/1595)

PR #1560, opened 2020-06-02
disabling missing-docstring in pylint

PatMyron:
https://stackoverflow.com/questions/7877522/how-do-i-disable-missing-docstring-warnings-at-a-file-level-in-pylint
also switched all remaining error codes to error names:
[`superfluous-parens`](https://github.com/PyCQA/pylint/blob/67adbe09fe6dbe717261be837a6d276d2fb6bf7f/pylint/checkers/format.py#L166)
[`locally-disabled`](https://github.com/PyCQA/pylint/blob/f00144fb8faa81e1ce42516c2f58eb9fc797f7bd/pylint/lint/pylinter.py#L83)


kddejong:
Any reason we are doing this?  I sometimes find these useful so just curious.

miparnisari:
> Any reason we are doing this? I sometimes find these useful so just curious.

I think the key  is in the _sometimes_ :P

kddejong:
True. Sometimes was a purpose choice in my wording but I’m not sure that should prevent trying.

PatMyron:
Was thinking about removing most of the rule function docstrings since they seemed covered by the [directory structure](https://github.com/aws-cloudformation/cfn-python-lint/tree/master/src/cfnlint/rules/) already:
```bash
shopt -s globstar
awk '/def match\(self, cfn\):/{getline; print}' src/cfnlint/rules/**/*.py | sort | uniq -c | sort -nr
     13         """Check CloudFormation Parameters"""
     12         """Check CloudFormation Resources"""
     11         """Check CloudFormation Outputs"""
      7         """Check CloudFormation Mappings"""
      6         """Basic Matching"""
      5         """Check CloudFormation Mapping"""
      5         """Check CloudFormation Join"""
      4         """Check CloudFormation GetAtt"""
      4         """Check CloudFormation Conditions"""
      2         """Check EC2 VPC Resource Parameters"""
      2         """Check CloudFormation Ref"""
      2         """Check CloudFormation Properties"""
```

kddejong:
Yea I think in the rules they can get kind of tedious.  Especially when the description property is there.

I care more for the helper functions, core functions, or functions in the Template class, etc.  They aren't all perfect but some of them are pretty good.  In the end this change doesn't prevent us from doing good doc strings in those libraries.  It just doesn't catch it if we miss one.

PR #1571, opened 2020-06-04
Don't check SSM parameter values for CIDR Regex

### LINKED ISSUE #1570: E2004 issue with AWS::SSM::Parameter::Value<String> Parameters
0.33.0

*Description of issue.*
When using an SSM paramter store lookup to get a CIDR Range to use in a Security group, cfn-lint raised an E2004 error.

```yaml
Parameters:
  VpcID:
    Description: ID of VPC
    Type: "AWS::SSM::Parameter::Value<String>"
    Default: VpcId
    AllowedValues:
      - VpcId

  VpcCidrRange:
    Description: CIDR range of VPC
    Type: "AWS::SSM::Parameter::Value<String>"
    Default: VpcCIDRRange
    AllowedValues:
      - VpcCIDRRange

Resources:
  SecurityGroup:
    Type: AWS::EC2::SecurityGroup
    Properties:
      GroupDescription: Allow HTTP egress
      VpcId: !Ref VpcID
      SecurityGroupEgress:
        - IpProtocol: tcp
          FromPort: 80
          ToPort: 80
          CidrIp: !Ref VpcCidrRange
```

Error is `E2004 Cidr should be a Cidr Range based string for VpcCIDRRange product.template.yaml:9:3`

If I'm not misunderstanding, cfn-lint seems to be complaining that the String `VpcCIDRRange` is not a valid CIDR, where in fact during runtime CloudFormation will lookup the value of `VpcCIDRRange` in SSM Parameter Store and use that instead.

I guess a sensible approach here would be to disable E2004 for parameters that have a type of `AWS::SSM::Parameter::Value<String>` maybe?

kddejong:
*Issue #, if available:*
fix #1570
*Description of changes:*
- Update rule E2004 to not check `AllowedValues` when the `Type` is `AWS::SSM::Parameter::Value<String>`

### DIFF
--- a/src/cfnlint/rules/parameters/CidrAllowedValues.py
+++ b/src/cfnlint/rules/parameters/CidrAllowedValues.py
@@ -60,7 +60,7 @@ class CidrAllowedValues(CloudFormationLintRule):
             parameter = parameters.get(value, {})
             parameter_type = parameters.get(value, {}).get('Type', '')
             if parameter_type == 'AWS::SSM::Parameter::Value<String>':
-                # If we are reading from the Parameter store we can longer
+                # If we are reading from the Parameter store we can no longer
                 # check the value of the allowed values
                 return matches
             allowed_values = parameter.get('AllowedValues', None)
--- a/src/cfnlint/rules/parameters/CidrAllowedValues.py
+++ b/src/cfnlint/rules/parameters/CidrAllowedValues.py
@@ -58,6 +58,11 @@ class CidrAllowedValues(CloudFormationLintRule):
 
         if value in parameters:
             parameter = parameters.get(value, {})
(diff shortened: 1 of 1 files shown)


src/cfnlint/rules/parameters/CidrAllowedValues.py
```diff
@@ -58,6 +58,11 @@ def check_cidr_ref(self, value, path, parameters, resources):
 
         if value in parameters:
             parameter = parameters.get(value, {})
+            parameter_type = parameters.get(value, {}).get('Type', '')
+            if parameter_type == 'AWS::SSM::Parameter::Value<String>':
+                # If we are reading from the Parameter store we can longer
```
PatMyron, line 63:
```suggestion
                # If we are reading from the Parameter store we can no longer
```

PR #1574, opened 2020-06-06
refactoring condition rules

PatMyron:
https://github.com/aws-cloudformation/cfn-python-lint/pull/1558
https://github.com/aws-cloudformation/cfn-python-lint/pull/1569


src/cfnlint/rules/conditions/common.py
```diff
@@ -0,0 +1,57 @@
...
+def check_condition_list(cfn, function):
+    matches = []
+    # Build the list of functions
+    trees = cfn.search_deep_keys(function)
+    for tree in trees:
+        # Test when in Conditions
+        if tree[0] == 'Conditions':
+            value = tree[-1]
+            if not isinstance(value, list):
+                message = function + ' must be a condition list'
```
PatMyron, line 17: did condense these previously slightly more informative messages here:
`function + ' must be a list of exactly 1 condition'`
`function + ' must be a list of between 2 to 10 conditions'`

PR #1580, opened 2020-06-11
Allow ignoring of E0000 and E0001 errors

### LINKED ISSUE #1133: Ignore-checks doesn't work for E000*
*cfn-lint version: 0.22.3

*Ignore Checks isn't working*

Template: https://gist.github.com/jeffmacdonald/cbbdd8cc3a049f89dfb4a17991ca1afa

I am getting the following error:

```
E0000 Expecting value
templates/ssh-jump-host.yaml:1:1
```

When I run `pipenv run cfn-lint --ignore-checks=E0000 templates/ssh-jump-host.yaml`

The fun part is, I have I have errors in other templates (E2521) and I am able to ignore those just fine... it seems it's specifically E0000 that doesn't get ignored.

kddejong:
*Issue #, if available:*
fix #1133 
*Description of changes:*
- Ignore check support for E0000.

Ignore_bad_template was not properly handled in a lot of scenarios.  This puts a more universal view on that parameter and treats it the same as ignoring `E0000`


src/cfnlint/core.py
```diff
@@ -157,9 +157,19 @@ def get_args_filenames(cli_args):
 def get_template_rules(filename, args):
     """ Get Template Configuration items and set them as default values"""
 
-    (template, matches) = cfnlint.decode.decode(filename, args.ignore_bad_template)
+    ignore_bad_template = False
+    if args.ignore_bad_template:
+        ignore_bad_template = True
+    else:
+        for ignore_check in args.ignore_checks:
+            if 'E0000' == ignore_check or 'E0000'.startswith(ignore_check):
```
kddejong, line 165: Doesn't handle mandatory rules.  Need to find a better way to handle this.

src/cfnlint/decode/__init__.py
```diff
@@ -18,7 +18,7 @@
 LOGGER = logging.getLogger(__name__)
 
 
-def decode(filename, ignore_bad_template):
+def decode(filename):
```
kddejong, line 21: Only using ignore_bad_template once and so many other issues with a template didn't handle this parameter.  Decided to remove it from here and move that logic outside of the decode.

PatMyron:
@kddejong [should `E0001` and `E0002` also be dealt with the same way as `E0000`?](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/docs/rules.md#E0000)

kddejong:
Yes. Going to work on that next.

src/cfnlint/core.py
```diff
@@ -157,9 +157,19 @@ def get_args_filenames(cli_args):
 def get_template_rules(filename, args):
     """ Get Template Configuration items and set them as default values"""
 
-    (template, matches) = cfnlint.decode.decode(filename, args.ignore_bad_template)
+    ignore_bad_template = False
+    if args.ignore_bad_template:
+        ignore_bad_template = True
+    else:
+        for ignore_check in args.ignore_checks:
```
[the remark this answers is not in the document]
kddejong: Yea, I don't like the parameter at all at this point.  We have done a lot of work to make everything an EXXXX code.  I wish we removed this and just switched to `--ignore-checks`

src/cfnlint/core.py
```diff
@@ -157,9 +157,19 @@ def get_args_filenames(cli_args):
...
+            if ignore_check == 'E0000' or 'E0000'.startswith(ignore_check):
+                ignore_bad_template = True
+
+    (template, matches) = cfnlint.decode.decode(filename)
 
     if matches:
+        if len(matches) == 1 and ignore_bad_template and matches[0].rule.id == 'E0000':
+            return(template, [], [])
```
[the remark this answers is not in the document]
kddejong: `if matches` will only continue if `matches` is not `[]`, `None`, etc.  I'm not sure if `len(matches)` provides much value there thought.

src/cfnlint/__main__.py
```diff
@@ -27,7 +27,9 @@ def main():
         for filename in filenames:
             LOGGER.debug('Begin linting of file: %s', str(filename))
             (template, rules, template_matches) = cfnlint.core.get_template_rules(filename, args)
-            if not template_matches:
+            # template matches may be empty but the template is still None
+            # this happens when ignoring bad templates
+            if not template_matches and template:
```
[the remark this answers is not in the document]
kddejong: What am I missing here?  This validate that there were no errors in the template (`not template_matches`) and `template` is not empty (`{}`, `[]`, or `template`)

src/cfnlint/__main__.py
```diff
@@ -27,7 +27,9 @@ def main():
         for filename in filenames:
             LOGGER.debug('Begin linting of file: %s', str(filename))
             (template, rules, template_matches) = cfnlint.core.get_template_rules(filename, args)
```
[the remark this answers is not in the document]
kddejong: I like `fatal_errors` or `errors`.  `template_matches` is a little vague.

src/cfnlint/__main__.py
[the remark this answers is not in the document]
miparnisari: ahhh ok my bad.

src/cfnlint/__main__.py
```diff
@@ -26,14 +26,16 @@ def main():
         rules = None
         for filename in filenames:
             LOGGER.debug('Begin linting of file: %s', str(filename))
-            (template, rules, template_matches) = cfnlint.core.get_template_rules(filename, args)
-            if not template_matches:
+            (template, rules, errors) = cfnlint.core.get_template_rules(filename, args)
```
kddejong, line 29: @miparnisari this look better?

src/cfnlint/config.py
```diff
@@ -511,7 +511,8 @@ def ignore_checks(self):
     @property
     def include_checks(self):
         """ include_checks """
-        return self._get_argument_value('include_checks', True, True)
+        results = self._get_argument_value('include_checks', True, True)
+        return ['W', 'E'] + results
```
kddejong, line 515: Want to have the default values of W, E accounted for in the config.

src/cfnlint/core.py
```diff
@@ -157,10 +157,26 @@ def get_args_filenames(cli_args):
...
-
-    if matches:
-        return(template, [], matches)
+    ignore_bad_template = False
+    if args.ignore_bad_template:
+        ignore_bad_template = True
+    else:
+        # There is no collection at this point so we need to handle this
+        # check directly
+        if not ParseError().is_enabled(
```
kddejong, line 166: this should now handle determining if the ParseError is enabled just like a normal rule.

test/unit/module/core/test_run_cli.py
```diff
@@ -37,7 +37,7 @@ def test_template_not_found(self):
             ['--template', filename, '--ignore-bad-template'])
         (_, _, matches) = cfnlint.core.get_template_rules(filenames[0], args)
 
-        self.assertEqual(len(matches), 1)
+        self.assertEqual(len(matches), 0)
```
kddejong, line 40: This does change some output values.  Now ignore bad template will always return back 0 if the template fails in decode.

Birds:
Any particular reason behind the name change of matches to errors?

Wondering if that may get confusing seeing most rules and all still refer to errors as matches

src/cfnlint/rules/__init__.py
```diff
@@ -42,6 +42,37 @@ def verbose(self):
...
         """Initialize the rule"""
 
+    def is_enabled(self, include_experimental=False, ignore_rules=None, include_rules=None,
+                   mandatory_rules=None):
+        """ Is the rule enabled based on the configuration """
+        ignore_rules = ignore_rules or []
+        include_rules = include_rules or []
+        mandatory_rules = mandatory_rules or []
+
+        # Evaluate experimental rulestox
```
PatMyron, line 52:
```suggestion
        # Evaluate experimental rules
```

src/cfnlint/rules/__init__.py
```diff
@@ -165,41 +199,16 @@ def __len__(self):
...
-                return True
-
-        # Allowing ignoring of rules based on prefix to ignore checks
-        for ignore_rule in self.ignore_rules:
-            if rule_id.startswith(ignore_rule) and ignore_rule:
-                return False
-
-        return True
+        return rule.is_enabled(self.include_experimental, self.ignore_rules,
+                               self.include_rules, self.mandatory_rules)
```
PatMyron, line 211: I like the switch to only passing `rule` into `is_rule_enabled()`

the new extra `rule.is_enabled()` layer confuses me a bit though, seems potentially unnecessary?

kddejong: I had to move the logic if a rule is enabled to the rule itself.  This is because E000X rules aren't part of the rule collection.  So I needed the same logic for checking if a rule is enabled if the rule belonged to a collection or not.

src/cfnlint/core.py
```diff
@@ -185,17 +201,30 @@ def run_checks(filename, template, rules, regions, mandatory_rules=None):
...
-    matches = []
+    errors = []
+
+    runner = cfnlint.runner.Runner(rules, filename, template, regions,
+                                   mandatory_rules=mandatory_rules)
+
+    # Transform logic helps with handling serverless templates
+    ignore_transform_error = False
+    if not rules.is_rule_enabled(TransformError()):
+        ignore_transform_error = True
```
PatMyron, line 212:
```suggestion
    ignore_transform_error = not rules.is_rule_enabled(TransformError())
```

src/cfnlint/core.py
```diff
@@ -185,17 +201,30 @@ def run_checks(filename, template, rules, regions, mandatory_rules=None):
...
+
+    errors.extend(runner.transform())
+
+    if errors:
+        if ignore_transform_error:
```
PatMyron, line 217:
```suggestion
        if not rules.is_rule_enabled(TransformError())
```

PatMyron, line 217:
```suggestion
    errors.extend(runner.transform())

    if errors:
        if not rules.is_rule_enabled(TransformError()):
```

kddejong: yes we could.  Could put a deprecation notice on it.

Issue #1520, opened 2020-05-08
cfn-lint fails using Python 3.9


mikeabackhouse:
cfn-lint --version fails running  Python 3.9

```
PS C:\WINDOWS\system32> cfn-lint --version
Traceback (most recent call last):
  File "c:\python39\lib\runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "c:\python39\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Python39\Scripts\cfn-lint.exe\__main__.py", line 5, in <module>
  File "c:\python39\lib\site-packages\cfnlint\__init__.py", line 18, in <module>
    from cfnlint.transform import Transform
  File "c:\python39\lib\site-packages\cfnlint\transform.py", line 10, in <module>
    from samtranslator.translator.translator import Translator
  File "c:\python39\lib\site-packages\samtranslator\translator\translator.py", line 2, in <module>
    from samtranslator.model import ResourceTypeResolver, sam_resources
  File "c:\python39\lib\site-packages\samtranslator\model\sam_resources.py", line 5, in <module>
    import samtranslator.model.eventsources.pull
  File "c:\python39\lib\site-packages\samtranslator\model\eventsources\pull.py", line 5, in <module>
    from samtranslator.translator.arn_generator import ArnGenerator
  File "c:\python39\lib\site-packages\samtranslator\translator\arn_generator.py", line 1, in <module>
    import boto3
  File "c:\python39\lib\site-packages\boto3\__init__.py", line 16, in <module>
    from boto3.session import Session
  File "c:\python39\lib\site-packages\boto3\session.py", line 17, in <module>
    import botocore.session
  File "c:\python39\lib\site-packages\botocore\session.py", line 28, in <module>
    import botocore.configloader
  File "c:\python39\lib\site-packages\botocore\configloader.py", line 19, in <module>
    from botocore.compat import six
  File "c:\python39\lib\site-packages\botocore\compat.py", line 150, in <module>
    import xml.etree.cElementTree
ModuleNotFoundError: No module named 'xml.etree.cElementTree'
```

Issue is relates to / caused by: Boto package
https://github.com/boto/botocore/issues/2002

kddejong:
Python 3.9 was released so this is moving to the top of the list.

PatMyron:
@mikeabackhouse is this issue still outstanding? I see the `botocore` issue has been resolved

PR #1564, opened 2020-06-02
speeding up Travis

PatMyron:
https://docs.travis-ci.com/user/build-stages/

https://travis-ci.com/github/aws-cloudformation/cfn-python-lint/builds

[`Ran for 8 min 33 sec`](https://travis-ci.com/github/aws-cloudformation/cfn-python-lint/builds/169334233) -> [`Ran for 4 min 42 sec`](https://travis-ci.com/github/aws-cloudformation/cfn-python-lint/builds/169334803)

PR #1584, opened 2020-06-11
testing Python 3.9

PatMyron:
https://github.com/aws-cloudformation/cfn-python-lint/issues/1520

https://tox.readthedocs.io/en/latest/example/basic.html

to start being aware of anything that doesn't work in Python 3.9 in anticipation of its release

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,5 +1,5 @@
 [tox]
-envlist = py{27,34,35,36,37,38,39},pylint{27,36,37,38}
+envlist = pylint{27,36,37,38}
 
 [testenv]
 commands =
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py{27,34,35,36,37,38},pylint{27,36,37,38}
+envlist = pylint{27,38}
```
PatMyron, line 2: shouldn't need to specify any tox environments other than these non-default `pylint` environments that are still being used

miparnisari:
image

why is this running twice?

PatMyron:
> why is this running ~~twice~~ thrice?

[had to get rid of the Travis labels to speed up the build](https://github.com/aws-cloudformation/cfn-python-lint/pull/1564), so have to look at [`.travis.yml`](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/.travis.yml) itself now, but there's linting, unit tests, and integration tests in Python 3.8

PR #1559, opened 2020-06-01
Introduction of Custom Rules with initial EQUAL operator support

Birds:
*Description of changes:*
- New custom rules feature allowing simple one line rules to be defined using specified operators
- Implementation of first "EQUALS" operator
- Basic custom rules template with example

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1596, opened 2020-06-17
Five new operators, Configurable error levels and messages, New custom rule ID system

Birds:
*Description of changes:*
- Implementation of all four operators (EQUALS, NOT_EQUALS, INSET, NOT_INSET)
- Configurable error levels for custom rules (WARN, ERROR)
- Configurable custom error messages
- New custom rules ID numbers based on line number

To Do:
- Unit test creation
- Update default custom rules text file to reflect the new syntax to use the new features


src/cfnlint/custom_rules.py
```diff
@@ -4,37 +4,42 @@
...
+                    operator_result = Operator[rule.operator](template, rule, resource_properties)
+                    matches += operator_result
                 except KeyError:
-                    matches.append(cfnlint.rules.Match(line_number, '0', '0', '0', filename, CustomRule('E9999'),
-                                                       str(rule.operator) + ' not in supported operators: [EQUALS] at ' + str(line), None))
+                    matches.append(cfnlint.rules.Match(
+                        1, 1,
+                        1, 1,
+                        template.filename, cfnlint.customRules.Operators.CustomRule('E9999'),
+                        str(rule.operator) + ' not in supported operators: [EQUALS] at ' + str(line), None))
```
Birds, line 39: Outdated, marking to remind me to fix this.

Birds: Shouldn't be hard coded, instead get list of dict keys and print them here so it dynamically updates

src/cfnlint/custom_rules.py
```diff
@@ -4,37 +4,42 @@
...
 import cfnlint.customRules.Operators
 import cfnlint.customRules.Rule
 
 LOGGER = logging.getLogger(__name__)
-Operator = {'EQUALS': lambda x, y: cfnlint.customRules.Operators.equalsOp(x, y),
-            'PLACEHOLDER': lambda x, y: LOGGER.debug('Placeholder Op')}
+Operator = {'EQUALS': lambda x, y, z: cfnlint.customRules.Operators.equalsOp(x, y, z),
+            'NOT_EQUALS': lambda x, y, z: cfnlint.customRules.Operators.notEqualsOp(x, y, z),
+            'INSET': lambda x, y, z: cfnlint.customRules.Operators.InSetOp(x, y, z),
+            'NOT_INSET': lambda x, y, z: cfnlint.customRules.Operators.NotInSetOp(x, y, z)}
```
PatMyron, line 18: thoughts on 
```suggestion
            'IN': lambda x, y, z: cfnlint.customRules.Operators.InSetOp(x, y, z),
            'NOT_IN': lambda x, y, z: cfnlint.customRules.Operators.NotInSetOp(x, y, z)}
```

Birds: The current structure of the program compares the raw operator input from the text file with one of the dict keys - I would prefer to keep that portion as is, as it's one of the neatest ways I think that could be coded.

On the other hand, I do get what you mean by there being a lot of duplicate code with the operators. I think another approach I would look into is creating a general helper class instead, and having each operator run that if possible for the duplicate portions.

PatMyron: > The current structure of the program compares the raw operator input from the text file with one of the dict keys - I would prefer to keep that portion as is, as it's one of the neatest ways I think that could be coded.

Will edit to clarify. I agree with this, but was just wondering about your thoughts on alternative operator syntax for the set operators themselves

> On the other hand, I do get what you mean by there being a lot of duplicate code with the operators. I think another approach I would look into is creating a general helper class instead, and having each operator run that if possible for the duplicate portions.

even just a method called in the methods in `src/cfnlint/customRules/Operators.py` could help if another class seems like overkill

Birds: Ooooh, I misunderstood. That new syntax does sound a lot better!

Yeah, I'm thinking one common helper method within the operators itself would make the whole file a whole lot neater with so much duplicate code removed.

src/cfnlint/customRules/Operators.py
```diff
@@ -9,28 +9,65 @@
 def equalsOp(template, rule, propertyList):
     """ Process EQUALS operators """
     matches = []
+    for prop in propertyList:
+        actualValue = getProperty(prop, rule)
+        if actualValue.strip().lower() != str(rule.value).strip().lower():
```
PatMyron, line 14: Nice! New operators should be easier to implement now that they're sharing most of the code

---

Don't need to decide yet, but starting a thread here for thoughts on case sensitivity of these rules so we don't forget

Birds: At the current moment, operators are case sensitive but values are not case sensitive. I'm open to either option though, or even making that configurable!

miparnisari: Case sensitivity is a good point, it matters in some cases but not others. Maybe having two distinct operators, EQUALS and STRICT EQUALS?

src/cfnlint/customRules/Operators.py
```diff
@@ -9,28 +9,91 @@
...
+        actualValue = getProperty(prop, rule)
+        if actualValue.strip() in rule.ruleSet:
+            matches.append(addMatches(template, rule, actualValue, prop, 'Not in set check failed'))
+    return matches
+
+
+def addMatches(template, rule, actualValue, prop, defaultMessage):
+    path = prop['Path']
+    path = path + rule.prop.split('.')
+    linenumbers = template.get_location_yaml(template.template, path)
```
miparnisari, line 78: Nitpicking, you can move line 78 to just before line 84

src/cfnlint/customRules/Operators.py
```diff
@@ -9,28 +9,91 @@
...
     return matches
 
 
+def lessOp(template, rule, propertyList):
+    """ Process <= operators """
+    matches = []
+    for prop in propertyList:
+        actualValue = getProperty(prop, rule)
+        if actualValue.strip().isnumeric() and str(rule.value).strip().isnumeric():
+            if int(actualValue.strip()) > int(str(rule.value).strip()):
```
miparnisari, line 48: So for `lessOp`, you are comparing `actualValue` against `rule.value`. But for `InSetOp` you are comparing against `rule.ruleSet`... even though they are both binary operators mathematically speaking. Can we not have just one variable for both? `rule.operand`?

src/cfnlint/customRules/Rule.py
```diff
@@ -17,13 +17,44 @@ class Rule(object):
     prop = ''
     operator = ''
     value = ''
+    lineNumber = 0
+    error_level = 'W'
+    ruleSet = []
+    error_message = ''
```
miparnisari, line 23: These are all class variables, right? it's a bit confusing that there are also instance variables with the same name

Birds: Great point. Instance variables have been given a more descriptive name which is not shared with class variables.

src/cfnlint/customRules/Operators.py
Birds: Had a Java moment there. I can absolutely use one variable "value" to represent all cases including the sets (with potential lists) and strings.

src/cfnlint/customRules/Operators.py
Birds: That sounds like a great idea - The rule class ingests all custom rule input first. That class could easily support both these cases, so that we do not have to clutter the main "Operator" dict.

PatMyron: I think the case sensitivity decision affects most of the operators so far (`EQUALS`, `NOT_EQUALS`, `IN`, `NOT_IN`)

I agree that someone will eventually want the ability to choose either case sensitive or case insensitive enforcement

Regardless of the final decision, it should be consistent across all the applicable operators

Birds: I'm looking at implementation of a "STIRCT_" in front of ANY operator where we want case sensitivity to be enforced. Common across all operators.

This can be done in the rule class to keep the operator class as neat and friendly as possible so anyone can add in an operator with minimal effort and knowledge!

PatMyron:
think the [Serverless transform](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/transform-aws-serverless.html) isn't being run before processing these custom rules:

```yaml
Resources:
  Resource:
    Type: AWS::Serverless::SimpleTable
```
```bash
$ echo "AWS::DynamoDB::Table BillingMode EQUALS PROVISIONED" > custom_rules.txt
$ cfn-lint --info template.yaml 
2020-06-22 13:26:47,027 - cfnlint - INFO - SAM Translator: 1.24.0
2020-06-22 13:26:47,028 - cfnlint - INFO - Setting AWS_DEFAULT_REGION to us-east-1
2020-06-22 13:26:47,086 - cfnlint - INFO - Transformed template: 
{
  "Resources": {
    "Resource": {
      "Properties": {
        "AttributeDefinitions": [
          {
            "AttributeName": "id",
            "AttributeType": "S"
          }
        ],
        "BillingMode": "PAY_PER_REQUEST",
        "KeySchema": [
          {
            "AttributeName": "id",
            "KeyType": "HASH"
          }
        ],
        "TableName": "G"
      },
      "Type": "AWS::DynamoDB::Table"
    }
  }
}
2020-06-22 13:26:47,086 - cfnlint.runner - INFO - Run scan of template template.yaml
```

[still think it's worth considering extending `CloudFormationLintRule`](https://github.com/aws-cloudformation/cfn-python-lint/pull/1559#issuecomment-638539312)

these differences will likely continue to surface if these custom constraints are run separately from the rest of the rules

miparnisari:
> [still think it's worth considering extending `CloudFormationLintRule`](https://github.com/aws-cloudformation/cfn-python-lint/pull/1559#issuecomment-638539312)

@Birds were you able to try this approach? And if so, did you encounter any issues?

miparnisari:
> think the Serverless transform isn't being run before processing these custom rules:

This was a bug in the order of processing. Function `run_checks` was not _just_ running checks, it was also sneakily running the Transform, which means that we have to place the running of custom rules after.

src/cfnlint/core.py
```diff
@@ -191,13 +188,15 @@ def run_checks(filename, template, rules, regions, mandatory_rules=None):
             raise InvalidRegionException(msg, 32)
 
     matches = []
-
+    """Pre-Processing"""
```
miparnisari: This comment looks weird, why not `# Pre-processing`?

Birds: Can add on for next PR

PR #1599, opened 2020-06-22
Fix superdomain logic for E3041

tomislacker:
### Issue #, if available
No issue filed. I identified the problem within a project I was working and decided to fix it myself.

### Description of changes
The current logic works something like this:

1. If the Hosted Zone name (`hz_name`) ends with a `.`, strip it
1. If the Name (`name`) ends with a `.`, strip it
1. `If not name.endswith('.' + hz_name):`, then a match/error is raised

This is incorrect because it prevents the ability for setting the top-level name (possibly an apex). This PR provides more reasonable functionality and most specifically targets the ability to set that top-level domain record for whatever the hosted zone is.

- - -

### DIFF
--- a/src/cfnlint/rules/resources/route53/RecordSetName.py
+++ b/src/cfnlint/rules/resources/route53/RecordSetName.py
@@ -44,7 +44,8 @@ class RecordSetName(CloudFormationLintRule):
                     hz_name = hz_name[:-1]
                 if name[-1] == '.':
                     name = name[:-1]
-                if not name.endswith('.' + hz_name):
+
+                if hz_name not in [name, name[-len(hz_name):]]:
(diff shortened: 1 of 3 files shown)


kddejong:
Every once in a while we get two triggered Travis builds.   Something odd happened there.

Looks like a few examples from:
`test/fixtures/templates/bad/resources/route53/recordset_name.yaml`
should be moved to
`test/fixtures/templates/good/resources/route53/recordset_name.yaml`

tomislacker:
@kddejong Thank you much for the feedback. Give me a couple minutes to read into these tests so I grok them, and I'll push up a fix!

tomislacker:
@kddejong I think I'm finally ready for a real review here.

PR #1601, opened 2020-06-22
New rule to check resources if ARNs use correctly placed Pseudo Parameters instead of hardcoded Partition, Region, and Account Number

dontirun:
https://github.com/aws-cloudformation/cfn-python-lint/pull/1532, https://github.com/aws-cloudformation/cfn-python-lint/pull/1533, https://github.com/aws-cloudformation/cfn-python-lint/issues/19

Description of changes:
Adding Rule I3042 and related testcases. Checks Resources if ARNs use correctly placed Pseudo Parameters instead of hardcoded Partition, Region, and Account Number


src/cfnlint/rules/resources/HardCodedArnProperties.py
```diff
@@ -0,0 +1,34 @@
...
+    def match(self, cfn):
+        """Check CloudFormation Resources"""
+        matches = []
+        pattern = re.compile(r'arn:(\$\{[^:]*::[^:]*}|[^:]*):[^:]+:(\$\{[^:]*::[^:]*}|[^:]*):(\$\{[^:]*::[^:]*}|[^:]*)')
+        resources = cfn.template.get('Resources', {})
+        if resources:
+            for resourcename, val  in resources.items():
+                candidates = pattern.findall(str(val))
+                for candidate in candidates:
+                    if candidate[0] != '${AWS::Partition}' or candidate[1] not in ('${AWS::Region}', '') or candidate[2] not in ('${AWS::AccountId}', ''):
```
kddejong, line 27: We should probably check the length of candidate before we assume it has 3 items.

src/cfnlint/rules/resources/HardCodedArnProperties.py
```diff
@@ -0,0 +1,34 @@
...
+    tags = ['resources']
+
+    def match(self, cfn):
+        """Check CloudFormation Resources"""
+        matches = []
+        pattern = re.compile(r'arn:(\$\{[^:]*::[^:]*}|[^:]*):[^:]+:(\$\{[^:]*::[^:]*}|[^:]*):(\$\{[^:]*::[^:]*}|[^:]*)')
+        resources = cfn.template.get('Resources', {})
+        if resources:
+            for resourcename, val  in resources.items():
+                candidates = pattern.findall(str(val))
```
kddejong, line 25: I think the way this is written should prevent finding other things from triggering this check.  I'm thinking join `!Join`?

src/cfnlint/rules/resources/HardCodedArnProperties.py
dontirun: The regex string has 3 capturing groups, so candidate will always have 3 items

src/cfnlint/rules/resources/HardCodedArnProperties.py
dontirun: Agreed !Join wouldn't be caught by this. !Join is a difficult case to handle because there is a lot of variance how joins could be written. Examples I can quickly think of.

```
!Join
  - ''
  - - 'arn:'
    - !Ref AWS::Partition
    - ':s3:::elasticbeanstalk-*-'
    - !Ref 'AWS::AccountId'
``` 
```
!Join [ "", [ 'arn:' ,!Ref 'AWS::Partition',:s3:::bucket]]
```

```
!Join [ ":", [ 'arn' ,!Ref 'AWS::Partition',s3:::bucket]]
```

```
!Join [ ":", [ 'arn' ,!Ref 'AWS::Partition',s3:: , bucket]]
```

src/cfnlint/rules/resources/HardCodedArnProperties.py
dontirun: Those could be. Would it make more sense to have 3 separate rules?

1. Warning: Hardcoded Partition
2. Informational: Hardcoded Region
3. Informational: Hardcoded AccountId

src/cfnlint/rules/resources/HardCodedArnProperties.py
kddejong: I should have been more clear.  How it is written is probably ideal.  To your point it becomes tricky.  I was mostly making note for it for myself to know the code wouldn't fail on a join.

src/cfnlint/rules/resources/HardCodedArnProperties.py
kddejong: I think I could be okay with it being one rule.  I'll admit I've totally messed up the partition portion of this in my templates but for region and account ID I usually do a reasonable job. I have always liked the idea of calling out hard coded values that shouldn't be.  What are your thoughts on if those values were references to parameters?  Or to a FindInMap?

dontirun: That's a good point, I could see the case of trying to specify a cross region resource or a different account as a Parameter. A simple fix would be to take the explicit checks such as `candidate[0] != '${AWS::Region}'`  and check against a regex pattern.

Something similar to the following

```
if candidate[0] != '${AWS::Partition}' or not re.match(r'\$\{\w+}', candidate[1]) or not re.match(r'\$\{\w+}', candidate[2]):
```

dontirun: Just added an update to allow for Parameterized AccountId and Regions

```if candidate[0] != '${AWS::Partition}' or not re.match(r'^\$\{\w+}|\$\{AWS::Region}|$', candidate[1]) or not re.match(r'^\$\{\w+}|\$\{AWS::AccountId}|$', candidate[2]):```

kddejong:
Replaced with #1805

PR #1613, opened 2020-06-29
Check if Metadata references a NoEcho parameter

Birds:
*Description of changes:*
If a resource's _metadata_ section contains a reference to a [_NoEcho_ parameter](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html#parameters-section-structure-properties), CloudFormation will display the plaintext value of the parameter's description. This new rule will detect if any resource references a _NoEcho_ parameter, and raise a warning.


test/unit/rules/resources/test_noecho.py
```diff
@@ -0,0 +1,23 @@
...
+        super(TestNoEcho, self).setUp()
+        self.collection.register(NoEcho())
+
+    def test_file_positive(self):
+        """Test Positive"""
+        self.helper_file_positive()
+
+    def test_file_negative(self):
+        """Test failure"""
+        self.helper_file_negative('test/fixtures/templates/bad/noecho.yaml', 3)
```
PatMyron, line 23:
```suggestion
        self.helper_file_negative('test/fixtures/templates/bad/noecho.yaml', 1)
```

src/cfnlint/rules/resources/NoEcho.py
```diff
@@ -0,0 +1,44 @@
...
+from cfnlint.rules import CloudFormationLintRule
+from cfnlint.rules import RuleMatch
+
+
+class NoEcho(CloudFormationLintRule):
+    """Check maximum Resource limit"""
+    id = 'W4002'
+    shortdesc = 'Check for NoEcho References'
+    description = 'Check if there is a NoEcho enabled parameter referenced within a resource'
+    source_url = 'https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/dne.html'
```
PatMyron, line 15:
```suggestion
    source_url = 'https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html#parameters-section-structure-properties'
```

test/unit/rules/resources/test_noecho.py
```diff
@@ -0,0 +1,23 @@
+"""
+Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+SPDX-License-Identifier: MIT-0
+"""
+from test.unit.rules import BaseRuleTestCase
+from cfnlint.rules.resources.NoEcho import NoEcho  # pylint: disable=E0401
+
+
+class TestNoEcho(BaseRuleTestCase):
+    """Test parameters limit number"""
```
PatMyron, line 10:
```suggestion
```
can just remove these now:
https://github.com/aws-cloudformation/cfn-python-lint/pull/1560
https://github.com/aws-cloudformation/cfn-python-lint/pull/1568

src/cfnlint/rules/resources/NoEcho.py
```diff
@@ -0,0 +1,44 @@
...
+Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+SPDX-License-Identifier: MIT-0
+"""
+from ast import literal_eval
+from cfnlint.rules import CloudFormationLintRule
+from cfnlint.rules import RuleMatch
+
+
+class NoEcho(CloudFormationLintRule):
+    """Check maximum Resource limit"""
```
PatMyron, line 11:
```suggestion
```
can just remove these now:
https://github.com/aws-cloudformation/cfn-python-lint/pull/1560
https://github.com/aws-cloudformation/cfn-python-lint/pull/1568

test/fixtures/templates/bad/noecho.yaml
```diff
@@ -0,0 +1,14 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Parameters:
+  NoEchoParam:
+    Description: This shouldn't show
```
PatMyron, line 4:
```suggestion
```

Nitpick, but could be slightly confusing if this description is talking about the description itself instead of the value of the parameter, so maybe we can just drop it

Birds: I'm certainly open to removing the statement, but am wondering what your thoughts are on instead changing the value to be more specific with something like "this value should not show".

Namely, I just want to ensure that any developer in the future can clearly see the reason behind the NoEcho warning without the need for significant context.

PatMyron: also feels a bit misleading because the value is currently shown :)

src/cfnlint/rules/resources/NoEcho.py
```diff
@@ -0,0 +1,43 @@
...
+        for resource_name, resource_values in resource_properties.items():
+
+            resource = literal_eval(str(resource_values))
+            for key, value in resource.items():
+                if key == 'Metadata':
+                    metadata = literal_eval(str(value))
+                    for prop_name, prop_value in metadata.items():
+                        path = ['Resources', resource_name, 'Metadata', prop_name]
+                        matches.append(RuleMatch(path, 'As the resource "metadata" section contains reference to a '
+                                                       '"NoEcho" parameter (' + prop_value + '), CFN will display the '
```
PatMyron, line 39: [`exception while processing rule W4002: can only concatenate str (not "dict") to str`](https://travis-ci.com/github/aws-cloudformation/cfn-python-lint/jobs/355611684#L311)

https://github.com/aws-cloudformation/cfn-python-lint/blob/f209ddfef9bcc1a005adfebcfcc16220b18deddb/test/fixtures/templates/bad/core/directives.yaml#L7-L11

https://github.com/aws-cloudformation/cfn-python-lint/blob/f209ddfef9bcc1a005adfebcfcc16220b18deddb/test/fixtures/templates/bad/core/mandatory_checks.yaml#L7-L11

test/fixtures/templates/bad/noecho.yaml
PatMyron: error message got bit by this same misunderstanding:
https://github.com/aws-cloudformation/cfn-python-lint/blob/9c4e023050771075d472bf3a246a534f54104fe1/src/cfnlint/rules/resources/NoEcho.py#L45

src/cfnlint/rules/resources/NoEcho.py
```diff
@@ -0,0 +1,46 @@
...
+                    for prop_name, prop_value in metadata.items():
+                        properties = literal_eval(str(prop_value))
+                        for property_name, property_value in properties.items():
+                            for parameter in no_echo_params:
+                                if str(property_value.find(str(parameter))):
+                                    path = ['Resources', resource_name, 'Metadata', prop_name]
+                                    matches.append(
+                                        RuleMatch(path, 'As the resource "metadata" section contains reference to a '
+                                                        '"NoEcho" parameter (' + str(property_value) +
+                                                  '), CFN will display the parameters description in plaintext'))
```
PatMyron, line 45:
```suggestion
                                                  '), CloudFormation will display the parameter value in plaintext'))
```

src/cfnlint/rules/resources/NoEcho.py
Birds: Fixed in newer patches

src/cfnlint/rules/resources/NoEcho.py
Birds: Will implement in upcoming patch

src/cfnlint/rules/resources/NoEcho.py
```diff
@@ -0,0 +1,46 @@
...
+"""
+from ast import literal_eval
+from cfnlint.rules import CloudFormationLintRule
+from cfnlint.rules import RuleMatch
+
+
+class NoEcho(CloudFormationLintRule):
+    id = 'W4002'
+    shortdesc = 'Check for NoEcho References'
+    description = 'Check if there is a NoEcho enabled parameter referenced within a resource'
```
miparnisari, line 13:
```suggestion
    description = 'Check if there is a NoEcho enabled parameter referenced within a resource's Metadata section'
```

src/cfnlint/rules/resources/NoEcho.py
PatMyron: include the parameter name in the error message as well

PatMyron:
Do we know why this is flagging [`NoEcho` parameters](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html#parameters-section-structure-properties) in [`AWS::CloudFormation::Interface`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cloudformation-interface.html) templates?
```bash
cfn-lint test/fixtures/templates/quickstart/nist_application.yaml --ignore-checks E3012 W3005
cfn-lint test/fixtures/templates/quickstart/vpc-management.json --ignore-checks E3012 W2509 W3005
```

https://github.com/aws-cloudformation/cfn-python-lint/blob/f209ddfef9bcc1a005adfebcfcc16220b18deddb/test/fixtures/templates/quickstart/nist_application.yaml#L68
https://github.com/aws-cloudformation/cfn-python-lint/blob/f209ddfef9bcc1a005adfebcfcc16220b18deddb/test/fixtures/templates/quickstart/vpc-management.json#L5-L42

and should it?

PatMyron:
also crashing on this valid template:
```yaml
Parameters:
  Parameter:
    Type: String
    NoEcho: true
Resources:
  Resource:
    Type: AWS::SNS::Topic
    Metadata:
      meta: data
```
```bash
E0002 Unknown exception while processing rule W4002: malformed node or string: <_ast.Name object at 0x112c11e50>
```

miparnisari:
> also crashing on this valid template:
> 
> ```yaml
> Parameters:
>   Parameter:
>     Type: String
>     NoEcho: true
> Resources:
>   Resource:
>     Type: AWS::SNS::Topic
>     Metadata:
>       meta: data
> ```

Not sure where @PatMyron grabbed this template from but @Birds please ensure that it is a unit test :)

Birds:
> ```yaml
> Parameters:
>   Parameter:
>     Type: String
>     NoEcho: true
> Resources:
>   Resource:
>     Type: AWS::SNS::Topic
>     Metadata:
>       meta: data
> ```

Valid and invalid unit tests are included! There was a variety of syntactical errors and lack of error catching which has been implemented.

src/cfnlint/rules/resources/NoEcho.py
Birds: Parameter name has been added onto the message

src/cfnlint/rules/resources/NoEcho.py
```diff
@@ -0,0 +1,52 @@
...
+                    no_echo_params.append(parameter_name)
+
+        resource_properties = cfn.get_resources()
+        for resource_name, resource_values in resource_properties.items():
+
+            resource = get_dict(resource_values)
+            for key, value in resource.items():
+                if key == 'Metadata':
+                    metadata = get_dict(value)
+                    for prop_name, prop_value in metadata.items():
```
kddejong, line 35: I don't think `literal_eval(str(input_string))` handles the logic the value isn't a dict already.  If the value is `1` then this will fail (I believe) with `'int' object has no attribute 'items'`.

kddejong:
```
Metadata:
      Test: "test"
      NoEchoParamInMetadata: !Ref NoEchoParam
```
`E0002 Unknown exception while processing rule W4002: 'int' object has no attribute 'values'`

src/cfnlint/rules/resources/NoEcho.py
```diff
@@ -0,0 +1,52 @@
...
+    description = 'Check if there is a NoEcho enabled parameter referenced within a resources Metadata section'
+    source_url = 'https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html#parameters-section-structure-properties'
+    tags = ['resources', 'NoEcho']
+
+    def match(self, cfn):
+        matches = []
+        no_echo_params = []
+        parameters = cfn.get_parameters()
+
+        for parameter_name, parameter_value in parameters.items():
```
kddejong, line 22:
```suggestion
        for parameter_name, parameter_value in parameters.items():
            noecho = parameter_value.get('NoEcho', default=False)
            if bool_compare(noecho, True):
                no_echo_params.append(parameter_name)
```

src/cfnlint/rules/resources/NoEcho.py
```diff
@@ -0,0 +1,52 @@
...
+            parameter_properties = literal_eval(str(parameter_value))
+            for parameter_properties_name, parameter_properties_value in parameter_properties.items():
+                if parameter_properties_name == 'NoEcho' and parameter_properties_value is True:
+                    no_echo_params.append(parameter_name)
+
+        resource_properties = cfn.get_resources()
+        for resource_name, resource_values in resource_properties.items():
```
kddejong, line 29:
```suggestion
        for resource_name, resource_values in resource_properties.items():
            for metadata, _ in resource_values.get_safe('Metadata', type_t=dict):
                for prop_name, prop_value in metadata.items():
                    if isinstance(prop_value, dict):
                        for parameter, _ in prop_value.get_safe('Ref', type_t=six.string_types):
```

kddejong:
Made a few suggestions and I know multiline suggestions aren't great.  Sorry about that.

src/cfnlint/rules/resources/NoEcho.py
Birds: That's a great point there. I took another look into how the logic was done and found room for improvement, changed in the latest commit!

src/cfnlint/rules/resources/NoEcho.py
```diff
@@ -0,0 +1,52 @@
...
+    source_url = 'https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html#parameters-section-structure-properties'
+    tags = ['resources', 'NoEcho']
+
+    def match(self, cfn):
+        matches = []
+        no_echo_params = []
+        parameters = cfn.get_parameters()
+
+        for parameter_name, parameter_value in parameters.items():
+            parameter_properties = literal_eval(str(parameter_value))
```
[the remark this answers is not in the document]
Birds: Changed, is now checking that it is indeed an object. Adjusted to use list comprehension to filter out non-dicts, and get rid of the entire get_dict in it's entirety

src/cfnlint/rules/resources/NoEcho.py
Birds: Wow, that's super neat! Implemented

src/cfnlint/rules/resources/NoEcho.py
Birds: Adjusted with earlier commit

src/cfnlint/rules/resources/NoEcho.py
```diff
@@ -0,0 +1,51 @@
...
+            return no_echo_params
+
+        resource_properties = cfn.get_resources()
+        resource_dict = {key: resource_properties[key] for key in resource_properties if
+                         isinstance(resource_properties[key], dict)}
+        for resource_name, resource_values in resource_dict.items():
+            resource_values = {key: resource_values[key] for key in resource_values if
+                               isinstance(resource_values[key], dict)}
+            for key, metadata in resource_values.items():
+                if key == 'Metadata':
```
kddejong, line 36: Can we just remove the 3 previous lines and just do a `resource_values.get('Metadata', {})` ?  If I'm reading this correctly we are Iterating over all the keys and only caring about `Metadata`.

src/cfnlint/rules/resources/NoEcho.py
```diff
@@ -0,0 +1,51 @@
...
+                    metadata = {key: metadata[key] for key in metadata if
+                                isinstance(metadata[key], dict)}
+                    for prop_name, properties in metadata.items():
+                        if isinstance(properties, dict):
```
kddejong, line 40: Is this check necessary if you are already doing `if isinstance(metadata[key], dict)` on line 37,38

Birds: Would be a redundant check - Left as is since one of the checks was removed while implementing the above changes.

src/cfnlint/rules/resources/NoEcho.py
Birds: Great catch! I've refactored the code there per your suggestions.

src/cfnlint/rules/resources/NoEcho.py
```diff
@@ -0,0 +1,47 @@
...
+        matches = []
+        no_echo_params = []
+        parameters = cfn.get_parameters()
+        for parameter_name, parameter_value in parameters.items():
+            noecho = parameter_value.get('NoEcho', default=False)
+            if bool_compare(noecho, True):
+                no_echo_params.append(parameter_name)
+
+        if not no_echo_params:
+            return no_echo_params
```
PatMyron, line 27: @Birds are these lines necessary?
Wouldn't `[]` get returned either way?

Birds: You're correct, this can certainly work without those lines - At the end of the day it'll be returned with an empty array either way

PatMyron:
not sure how customers should remediate this warning since dynamic references don't always work there:
https://github.com/aws-cloudformation/aws-cloudformation-coverage-roadmap/issues/179
https://stackoverflow.com/questions/53589880/dynamic-references-to-specify-secret-manager-values-in-aws-cloudformation
and cfn-init relies on metadata:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-init.html#aws-resource-cloudformation-init-syntax

PR #1620, opened 2020-07-09
Update rule E3503 to not fail if ValidationDomain or DomainName aren't present

### LINKED ISSUE #1619: E3503 does not match CloudFormation - requires ValidationDomain when CF does not want it

cfn-lint 0.33.2

*Description of issue.*

I created an ACM certificate resource, and there were problems configuring the `DomainValidationOptions` block. If using DNS validation, the only properties needed are `DomainName` and `HostedZoneId`. However, cfn-lint was demanding a third property named `ValidationDomain`. When submitting the stack for deployment to CF, it triggered an immediate rollback because CF views `HostedZoneId` and `ValidationDomain` as mutually exclusive.

Adding an ignore rule to skip the E3503 error allowed me to proceed without issues. This rule should be adjusted to match what CF enforces.

**Sample:**

```yaml
Resources:
  Certificate:
    Type: AWS::CertificateManager::Certificate
    Metadata:
      cfn-lint:
        config:
          ignore_checks:
            - E3503
    Properties:
      DomainName: "*.aws.domain.com"
      ValidationMethod: DNS
      DomainValidationOptions:
        - DomainName: aws.domain.com
          HostedZoneId: !ImportValue SubdomainHostedZoneId
```

kddejong:
*Issue #, if available:*
Fix #1619
*Description of changes:*
- Update rule E3503 to use None instead of `''` when getting ValidationDomain and DomainName. There are valid scenarios where ValidationDomain may not be present and comparing to an empty string isn't the correct way to handle it.  We should only be doing this rule when the values are actually present.

### DIFF
--- a/test/fixtures/templates/bad/resources/certificatemanager/domain_validation_options.yaml
+++ b/test/fixtures/templates/bad/resources/certificatemanager/domain_validation_options.yaml
@@ -16,12 +16,3 @@ Resources:
       DomainValidationOptions:
         - DomainName: "prefixed-example.com"
           ValidationDomain: "example.com"
-  Certificate3:
-    Type: AWS::CertificateManager::Certificate
-    Properties:
-      DomainName: "*.aws.domain.com"
-      ValidationMethod: DNS
-      DomainValidationOptions:
-        # Don't fail when one of the values don't exist
-        - DomainName: aws.domain.com
-          HostedZoneId: !ImportValue SubdomainHostedZoneId
--- a/test/fixtures/templates/bad/resources/certificatemanager/domain_validation_options.yaml
+++ b/test/fixtures/templates/bad/resources/certificatemanager/domain_validation_options.yaml
@@ -16,3 +16,12 @@ Resources:
       DomainValidationOptions:
         - DomainName: "prefixed-example.com"
           ValidationDomain: "example.com"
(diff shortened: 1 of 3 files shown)


test/fixtures/templates/bad/resources/certificatemanager/domain_validation_options.yaml
```diff
@@ -16,3 +16,12 @@ Resources:
...
           ValidationDomain: "example.com"
+  Certificate3:
+    Type: AWS::CertificateManager::Certificate
+    Properties:
+      DomainName: "*.aws.domain.com"
+      ValidationMethod: DNS
+      DomainValidationOptions:
+        # Don't fail when one of the values don't exist
+        - DomainName: aws.domain.com
+          HostedZoneId: !ImportValue SubdomainHostedZoneId
```
PatMyron, line 27: nitpick, but should this resource go here if it's valid? [`test/fixtures/templates/good/resources/certificatemanager/domain_validation_options.yaml`](https://github.com/aws-cloudformation/cfn-python-lint/blob/d5f7dce74f9dc714144f775c76e51547f730773f/test/fixtures/templates/good/resources/certificatemanager/domain_validation_options.yaml)

PR #1627, opened 2020-07-16
Update rule W4002 to only look at Ref and Sub

### LINKED ISSUE #1625: False alarm from new W4002
*cfn-lint version: 0.34.0*

[Here](https://gist.github.com/schmiddy/44a779032a930995d22ee2722a18f163) is an example template which causes a false alarm like this:

```
$ cfn-lint /tmp/example.yml 
W4002 As the resource "metadata" section contains reference to a "NoEcho" parameter DBUser, CloudFormation will display the parameter value in plaintext
/tmp/example.yml:21:7

W4002 As the resource "metadata" section contains reference to a "NoEcho" parameter DBPass, CloudFormation will display the parameter value in plaintext
/tmp/example.yml:21:7
```

The problem seems to be that the rule is looking for any mention of the parameter name, even as a text description that is not actually referencing the parameter.

kddejong:
*Issue #, if available:*
Fix #1625
*Description of changes:*
- Update rule W4002 to look only at Ref and Sub

### DIFF
--- a/test/fixtures/templates/bad/noecho.yaml
+++ b/test/fixtures/templates/bad/noecho.yaml
@@ -1,13 +1,19 @@
-AWSTemplateFormatVersion: '2010-09-09'
+AWSTemplateFormatVersion: "2010-09-09"
 Parameters:
   NoEchoParam:
     Type: String
     Default: NoEchoWasEchoed
     NoEcho: true
 Resources:
-  SNSTopicWithSecretName:
+  SNSTopicWithSecretNameInRef:
     Type: AWS::SNS::Topic
     Properties:
       DisplayName: testname
     Metadata:
       NoEchoParamInMetadata: !Ref NoEchoParam
+  SNSTopicWithSecretNameInSub:
+    Type: AWS::SNS::Topic
+    Properties:
+      DisplayName: testname
+    Metadata:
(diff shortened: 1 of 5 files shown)


test/fixtures/templates/bad/noecho.yaml
```diff
@@ -1,13 +1,19 @@
...
     Properties:
       DisplayName: testname
     Metadata:
       NoEchoParamInMetadata: !Ref NoEchoParam
+  SNSTopicWithSecretNameInSub:
+    Type: AWS::SNS::Topic
+    Properties:
+      DisplayName: testname
+    Metadata:
+      NoEchoParamInMetadata: !Sub "${NoEchoParam}"
```
PatMyron, line 19: [Looks like it's catching both as expected.](https://travis-ci.com/github/aws-cloudformation/cfn-python-lint/jobs/361865901#L351) This should be bumped as well now:

https://github.com/aws-cloudformation/cfn-python-lint/blob/b8b6cf1cc4420b28c5cb757dc2377af63b4a1acf/test/unit/rules/resources/test_noecho.py#L22

EslamElHusseiny:
Is this PR going to be released any soon in the future?

kddejong:
Releasing right now.  It should be out there shortly.

PR #1640, opened 2020-08-04
Update rule E8003 to look for string based parameters

### LINKED ISSUE #1639: Cannot catch "Fn::Equals object requires a list of 2 string parameters"
```
cfn-lint 0.34.1
```

*aws-cli version: (`aws --version`)*
```
aws-cli/1.18.97 Python/3.8.3 Linux/4.15.0-111-generic botocore/1.16.26
```

*Description of issue.*
```
$ cat template.yml
Parameters:
  Environments:
    Type: CommaDelimitedList
    Default: dev,test
Conditions:
  TagEnvironments: !Not
  - !Equals
    - !Ref Environments  # missed !Join
    - ''
Resources:
  VPC:
    Type: AWS::EC2::VPC
    Properties:
      CidrBlock: 172.16.0.0/24
      Tags: !If
      - TagEnvironments
      - - Key: Environments
          Value: !Join
          - ','
          - !Ref Environments
      - []
$ aws cloudformation deploy --stack-name test --template-file template.yml

An error occurred (ValidationError) when calling the CreateChangeSet operation: Template error: every Fn::Equals object requires a list of 2 string parameters.
$ aws cloudformation validate-template --template-body file://template.yml

An error occurred (ValidationError) when calling the ValidateTemplate operation: Template error: every Fn::Equals object requires a list of 2 string parameters.
$ cfn-lint template.yml
$ echo $?
0
```

kddejong:
*Issue #, if available:*
fix #1639 

*Description of changes:*
- Update rule E8003 to look for string based parameters

### DIFF
--- a/src/cfnlint/helpers.py
+++ b/src/cfnlint/helpers.py
@@ -171,6 +171,64 @@ valid_snapshot_types = [
     'AWS::Redshift::Cluster'
 ]
 
+VALID_PARAMETER_TYPES_SINGLE = [
+    'AWS::EC2::AvailabilityZone::Name',
+    'AWS::EC2::Image::Id',
+    'AWS::EC2::Instance::Id',
+    'AWS::EC2::KeyPair::KeyName',
+    'AWS::EC2::SecurityGroup::GroupName',
+    'AWS::EC2::SecurityGroup::Id',
+    'AWS::EC2::Subnet::Id',
+    'AWS::EC2::VPC::Id',
+    'AWS::EC2::Volume::Id',
+    'AWS::Route53::HostedZone::Id',
+    'AWS::SSM::Parameter::Name',
+    'Number',
+    'String',
+    'AWS::SSM::Parameter::Value<AWS::EC2::AvailabilityZone::Name>',
+    'AWS::SSM::Parameter::Value<AWS::EC2::Image::Id>',
+    'AWS::SSM::Parameter::Value<AWS::EC2::Instance::Id>',
+    'AWS::SSM::Parameter::Value<AWS::EC2::KeyPair::KeyName>',
+    'AWS::SSM::Parameter::Value<AWS::EC2::SecurityGroup::GroupName>',
+    'AWS::SSM::Parameter::Value<AWS::EC2::SecurityGroup::Id>',
+    'AWS::SSM::Parameter::Value<AWS::EC2::Subnet::Id>',
+    'AWS::SSM::Parameter::Value<AWS::EC2::VPC::Id>',
+    'AWS::SSM::Parameter::Value<AWS::EC2::Volume::Id>',
+    'AWS::SSM::Parameter::Value<AWS::Route53::HostedZone::Id>',
+    'AWS::SSM::Parameter::Value<AWS::SSM::Parameter::Name>',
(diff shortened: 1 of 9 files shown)


PatMyron:
Could look into pulling [`02_parameter_types.json`](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/data/ExtendedSpecs/all/02_parameter_types.json) out of the patching process with the `VALID_PARAMETER_TYPES` helper as well:

https://github.com/aws-cloudformation/cfn-python-lint/blob/bf0bb6bbcdaba3cc72a7efdf7a1274543e1a046e/src/cfnlint/rules/resources/properties/ValueRefGetAtt.py#L67

src/cfnlint/helpers.py
```diff
@@ -171,6 +171,39 @@
...
+    'List<AWS::EC2::Image::Id>',
+    'List<AWS::EC2::Instance::Id>',
+    'List<AWS::EC2::SecurityGroup::GroupName>',
+    'List<AWS::EC2::SecurityGroup::Id>',
+    'List<AWS::EC2::Subnet::Id>',
+    'List<AWS::EC2::VPC::Id>',
+    'List<AWS::EC2::Volume::Id>',
+    'List<AWS::Route53::HostedZone::Id>',
+    'List<Number>',
+    'List<String>',
```
PatMyron, line 202: actually not seeing `List<String>` [documented](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html#parameters-section-structure-properties), so just wanna double check that's valid

src/cfnlint/helpers.py
```diff
@@ -171,6 +171,39 @@
...
+    'AWS::EC2::Image::Id',
+    'AWS::EC2::Instance::Id',
+    'AWS::EC2::KeyPair::KeyName',
+    'AWS::EC2::SecurityGroup::GroupName',
+    'AWS::EC2::SecurityGroup::Id',
+    'AWS::EC2::Subnet::Id',
+    'AWS::EC2::VPC::Id',
+    'AWS::EC2::Volume::Id',
+    'AWS::Route53::HostedZone::Id',
+    'AWS::SSM::Parameter::Name',
```
PatMyron, line 185: what about the other documented [`AWS::SSM::Parameter` parameter types](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html#aws-ssm-parameter-types)?

src/cfnlint/helpers.py
kddejong: it is valid (I think).  I thought I had tested this before but left it off because it isn't documented any where.

src/cfnlint/helpers.py
kddejong: good call.

kddejong:
We didn't include the return types for those parameters.  We probably should have.  We could fix that easily enough though.  Thoughts?

kddejong:
Moving this forward.  I think @PatMyron is correct.  We need to figure out how to better handle the data around Parameters between the spec patching and what is in the helpers.  I'm going to leave that as a different request at this point because that data in the spec is being used when determining valid Refs to Parameters.

PR #1647, opened 2020-08-12
Update regex to check MetricValue for number or string that starts with $

### LINKED ISSUE #1525: AWS::Logs::MetricFilter MetricValues permits invalid strings
`cfn-lint 0.30.1`

AWS::Logs::MetricFilter.Properties.MetricTransformations[*].MetricValue allows a bare string not starting with '$' which it appears is never actually valid, ie 
"MetricValue: length" vs "MetricValue: $length"

Assuming I'm reading the documentation correctly MetricValue must always either be a number OR start with a '$' character.

The following fragment lints, but is rejected by CloudFormation at runtime without a $ at the start of the named MetricValue field
```
    QueueLengthMetricFilter:
        Type: AWS::Logs::MetricFilter
        Properties:
            LogGroupName: !Ref LogGroup
            FilterPattern: '[date, time, tag="rh-sched*", x01=throttling, x02="jobs.", ..., x10=Len, x11=of, x12=job, x13="queue*",  length]'
            MetricTransformations:
              - MetricValue: length
                MetricNamespace: !Sub '${EnvironmentName}'
                MetricName: 'JobsQueued'
```

Note: I believe that this is also missed by the AWS ValidateTemplate API

kddejong:
*Issue #, if available:*
Fix #1525 

*Description of changes:*
- Include a regex pattern to check MetricValue is either a number or starts with $
- Update rule E3031 to convert int or float to string before doing an allowed pattern match on it


src/cfnlint/rules/resources/properties/AllowedPattern.py
```diff
@@ -34,18 +35,22 @@ def check_value(self, value, path, property_name, **kwargs):
         # the RegEx itself is used.
         value_pattern = kwargs.get('value_specs', {}).get('AllowedPattern', value_pattern_regex)
 
-        if value_pattern_regex:
-            regex = re.compile(value_pattern_regex)
+        if isinstance(value, (int, float)):
+            value = str(value)
```
PatMyron, line 39: [seems okay to convert from looking through the existing `AllowedPatternRegex`](https://github.com/aws-cloudformation/cfn-python-lint/search?l=JSON&q=AllowedPatternRegex)

kddejong: Yea ran into an issue when this regex was looking at something was just a number.  Realized we needed to convert an int to a string to make sure the regex pattern will work.

PR #1648, opened 2020-08-12
Update rule E1029 to skip TemplateBody properties

### LINKED ISSUE #1645: Found an embedded parameter "${AWS::Partition}" outside of an "Fn::Sub" at Resources
cfn-lint 0.29.5

*Description of issue.*
I have a CloudFormation Template embedded into a AWS Systems Manager Automation Document. 

* Template linting issues:

`E1029 Found an embedded parameter "${AWS::Partition}" outside of an "Fn::Sub" at Resources/DNSForwarderSetup/Properties/Content/mainSteps/0/inputs/TemplateBody
ad-3.template:433:13`

This is actually okay, I am using !Sub correctly in an Embedded CloudFormation.... 

https://github.com/aws-quickstart/quickstart-microsoft-activedirectory/blob/b071c3def8d9a30fd055571c9603cad10a753f40/templates/ad-3.template#L433

  * Validate if the issue still exists with the latest version of `cfn-lint` and/or the latest Spec files
* Feature request:
  * Please provide argumentation about the missing feature. Context is key!
This errors if I use !Sub or Fn::Sub: when it actually correct.. 

Cfn-lint uses the [CloudFormation Resource Specifications](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html) as the base to do validation. These files are included as part of the application version. Please update to the latest version of `cfn-lint` or update the spec files manually (`cfn-lint -u`)

kddejong:
*Issue #, if available:*
fix #1645 

*Description of changes:*
- Update rule E1029 to skip over TemplateBody parameters.   Which are used to have embedded templates in SSM


PatMyron:
@kddejong I was actually confused about this one since it seems like that template did use [`Sub`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-sub.html) there:

https://github.com/aws-quickstart/quickstart-microsoft-activedirectory/blob/b071c3def8d9a30fd055571c9603cad10a753f40/templates/ad-3.template#L479-L481

kddejong:
TemplateBody is a string who's value is a nested template.

kddejong:
@PatMyron We don't parse strings looking for the Sub.  Since this is a string https://github.com/aws-quickstart/quickstart-microsoft-activedirectory/blob/b071c3def8d9a30fd055571c9603cad10a753f40/templates/ad-3.template#L434-L508 that is as far as we go.  This rule (the way its written today) would have had to have a !Sub on ~434.  That would have (or could have) lead to some unexpected results.

PR #1659, opened 2020-08-18
update E1019 and E1010 to handle attributes of a resource having type Map

### LINKED ISSUE #1658: AWS::ServiceCatalog::CloudFormationProvisionedProduct.Outputs
In July 2020, support for retrieving the outputs of provisioned Service Catalog products [was added to CloudFormation](https://aws.amazon.com/blogs/mt/how-to-launch-secure-and-governed-aws-resources-with-aws-cloudformation-and-aws-service-catalog/).

Given the following example template:

```yaml

Resources:
  ProvisionedProduct:
    Type: AWS::ServiceCatalog::CloudFormationProvisionedProduct
    Properties:
      ProductName: example
      ProvisioningArtifactName: v1

Outputs:
  OutputValue:
    Value: !GetAtt ProvisionedProduct.Outputs.Example

  OutputSub:
    Value: !Sub '${ProvisionedProduct.Outputs.Example}-example'
```

The following linting errors are output when using cfn-lint 0.35.0:

```
E1010 Invalid GetAtt ProvisionedProduct.Outputs.Example for resource OutputValue
example.yml:12:5

E1019 Parameter ProvisionedProduct.Outputs.Example for Fn::Sub not found at Outputs/OutputSub/Value/Fn::Sub
example.yml:15:5
```

Access to these outputs using `!GetAtt` is documented in the announcement of this functionality (e.g. `!GetAtt ApacheALB.Outputs.ALBTarget`). Additionally the example template given above works correctly when deployed.

`Output` is documented in the [AWS::ServiceCatalog::CloudFormationProvisionedProduct return values](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-servicecatalog-cloudformationprovisionedproduct.html#aws-resource-servicecatalog-cloudformationprovisionedproduct-return-values) and appears to already be defined within cfn-lint CloudSpecs as a `Map` type.

I believe that this a false positive and am not aware of any alternative syntax to access these outputs (e.g. `!FindInMap` doesn't appear to be usable in this instance).

kddejong:
*Issue #, if available:*
fix #1658

*Description of changes:*
- update E1019 and E1010 to handle attributes of a resource having type Map

### DIFF
--- a/src/cfnlint/rules/functions/GetAtt.py
+++ b/src/cfnlint/rules/functions/GetAtt.py
@@ -80,6 +80,15 @@ class GetAtt(CloudFormationLintRule):
             if resname:
                 if resname in valid_getatts:
                     if restype is not None:
+                        # Check for maps
+                        restypeparts = restype.split('.')
+                        if restypeparts[0] in valid_getatts[resname] and len(restypeparts) >= 2:
+                            if valid_getatts[resname][restypeparts[0]].get('Type') != 'Map':
+                                message = 'Invalid GetAtt {0}.{1} for resource {2}'
+                                matches.append(RuleMatch(
+                                    getatt[:-1], message.format(resname, restype, getatt[1])))
+                            else:
+                                continue
                         if restype not in valid_getatts[resname] and '*' not in valid_getatts[resname]:
                             message = 'Invalid GetAtt {0}.{1} for resource {2}'
                             matches.append(RuleMatch(
--- a/src/cfnlint/rules/functions/Sub.py
+++ b/src/cfnlint/rules/functions/Sub.py
@@ -115,19 +115,19 @@ class Sub(CloudFormationLintRule):
                         tree, message.format(parameter, '/'.join(map(str, tree)))))
             for resource, attributes in get_atts.items():
                 for attribute_name, attribute_values in attributes.items():
-                    if resource == parameter.split('.')[0] and attribute_name == '*':
-                        if attribute_values.get('Type') == 'List':
-                            message = 'Fn::Sub cannot use list {0} at {1}'
-                            matches.append(RuleMatch(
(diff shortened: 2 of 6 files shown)


src/cfnlint/rules/functions/GetAtt.py
```diff
@@ -80,6 +80,15 @@ def match(self, cfn):
             if resname:
                 if resname in valid_getatts:
                     if restype is not None:
+                        # Check for maps
+                        restypeparts = restype.split('.')
+                        if restypeparts[0] in valid_getatts[resname] and len(restypeparts) >= 2:
+                            if valid_getatts[resname][restypeparts[0]].get('Type') != 'Map':
```
PatMyron, line 86: @kddejong [`AWS::ServiceCatalog::CloudFormationProvisionedProduct`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-servicecatalog-cloudformationprovisionedproduct.html#aws-resource-servicecatalog-cloudformationprovisionedproduct-return-values)'s the only `Map` so far:

```bash
curl -s --compressed https://d1uauaxba7bl26.cloudfront.net/latest/gzip/CloudFormationResourceSpecification.json | jq '.ResourceTypes' | jq 'with_entries(.value |= .Attributes)' | grep -v ': null' | grep -Eo 'Type": ".*"' | sort | uniq -c | sort -nr
 504 Type": "String"
  26 Type": "List"
   6 Type": "Integer"
   4 Type": "Boolean"
   1 Type": "Map"
   1 Type": "Json"
   1 Type": "FindingsFilterListItem"
```

so I was considering hardcoding that exception, but this seems even better 👍

kddejong: Yea, I'm not a huge fan of this solution.  I think there are better long term solutions if this continues to grow.

Issue #1683, opened 2020-09-05
Return W3002 instead of E3002 for additional resources supported by package command


asjohnston-asf:
cfn-lint version: `cfn-lint 0.35.1`

Some resources supported by `aws cloudformation package` return `E3002 Expecting an object...`; they should instead return `W3002 This code may only work with package cli command...` when the resource value is a string.

Example:
```
$ cfn-lint --version
cfn-lint 0.35.1

$ cat myTemplate.yml 
Resources:

  myStateMachine:
    Type: AWS::StepFunctions::StateMachine
    Properties:
      RoleArn: myRoleArn
      DefinitionS3Location: myLocalDefinitionFile.json

$ cfn-lint myTemplate.yml 
E3002 Expecting an object at Resources/myStateMachine/Properties/DefinitionS3Location
myTemplate.yml:7:7
```

My team is specifically interested in support for [`AWS::StepFunctions:StateMachine DefinitionS3Location`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-stepfunctions-statemachine.html#cfn-stepfunctions-statemachine-definitions3location).  We would prefer to use `package` for our step function definitions so we can perform static analysis on the definition separate from the cloudformation template.  That would currently require an `--ignore-checks E3002` flag when validating the template, but we consider that check too critical to ignore.

Resources currently supported by `package` are documented at:
https://docs.aws.amazon.com/cli/latest/reference/cloudformation/package.html

Exceptions for some of these resources are implemented in `cfn-lint` at:
https://github.com/aws-cloudformation/cfn-python-lint/blob/fe0af928a4ab26a6cedfd76564a26c9a4120b14f/src/cfnlint/rules/resources/properties/Properties.py#L179-L184
https://github.com/aws-cloudformation/cfn-python-lint/blob/fe0af928a4ab26a6cedfd76564a26c9a4120b14f/src/cfnlint/rules/resources/properties/PropertiesTemplated.py#L42-L47

asjohnston-asf:
Adding an exception for DefinitionS3Location ito resolve the issue with my example is simple enough: https://github.com/aws-cloudformation/cfn-python-lint/compare/master...asjohnston-asf:master

[...]

Issue #608, opened 2019-01-25
cfn-lint throwing E3002 instead of W3002 on Lambda::LayerVersion


RadicalDev:
*cfn-lint version: `0.12.1`*

*Description of issue.*

cfn-lint is throwing `E3002 Expecting an object at Resources/LambdaLayerVersion/Properties/Content` when linting a template where a path is passed for a `AWS::Lambda::LayerVersion`'s `Content` similarly to how we would do that for a `AWS::Lambda::Function`'s `Code`. 

This is intended to work with the `aws cloudformation package` command which now supports `AWS::Lambda::LayerVersion`. 

Please provide as much information as possible:
* Template linting issues:
  * Please provide a CloudFormation sample that generated the issue.
```yaml
AWSTemplateFormatVersion: 2010-09-09

Resources:

  LambdaLayerVersion:
    Type: AWS::Lambda::LayerVersion
    Properties:
      Description: A layer version which will contain static content
      CompatibleRuntimes:
        - python3.6
        - python3.7
      Content: ../static-content
```
  * If present, please add links to the (official) documentation for clarification.

    The documentation doesn't mention `aws cloudformation package` support for `AWS::Lambda::LayerVersion` yet, but support was added in aws-cli version [1.16.93](https://github.com/aws/aws-cli/blob/1.16.93/awscli/customizations/cloudformation/artifact_exporter.py#L373)

* Feature request:
  * Please provide argumentation about the missing feature. Context is key!

We're packaging static files in lambda layers and would like for cfn-lint to provide the expected behavior of throwing a warning like `W3002 This code may only work with `package` cli command as the property (Resources/LambdaLayerVersion/Properties/Content) is a string`

PR #1684, opened 2020-09-05
expanding templated_exceptions property types that require package command (E3002/W3002)

PatMyron:
fixes https://github.com/aws-cloudformation/cfn-python-lint/issues/1683

According to https://github.com/aws-cloudformation/cfn-python-lint/issues/608, this [documentation seems incomplete](https://docs.aws.amazon.com/cli/latest/reference/cloudformation/package.html), so I used [this as the source instead](https://github.com/aws/aws-cli/blob/master/awscli/customizations/cloudformation/artifact_exporter.py)

---

these other property types were already `String`:

```python
            'AWS::AppSync::FunctionConfiguration': ['RequestMappingTemplateS3Location', 'ResponseMappingTemplateS3Location'],
            'AWS::AppSync::GraphQLSchema': ['DefinitionS3Location'],
            'AWS::AppSync::Resolver': ['RequestMappingTemplateS3Location', 'ResponseMappingTemplateS3Location'],
            'AWS::CloudFormation::Stack': ['TemplateURL'],
            'AWS::Glue::Job': ['Command.ScriptLocation'],  # not sure about this syntax
```

---

Test template:

```yaml
Resources:
  StateMachine:
    Type: AWS::StepFunctions::StateMachine
    Properties:
      RoleArn: RoleArn
      DefinitionS3Location: file
  LayerVersion:
    Type: AWS::Lambda::LayerVersion
    Properties:
      Content: file
  RestApi:
    Type: AWS::ApiGateway::RestApi
    Properties:
      BodyS3Location: file
```

---

@kddejong any way this section can use `templated_exceptions` keys as well?

https://github.com/aws-cloudformation/cfn-python-lint/blob/df4b7bc463d301b638f0e0274ed0e41d37239092/src/cfnlint/rules/resources/properties/PropertiesTemplated.py#L34-L38

### DIFF
--- a/src/cfnlint/rules/resources/properties/Properties.py
+++ b/src/cfnlint/rules/resources/properties/Properties.py
@@ -177,11 +177,11 @@ class Properties(CloudFormationLintRule):
             - Start with handling exceptions for templated code.
         """
         templated_exceptions = {
-            'AWS::ApiGateway::RestApi': ['BodyS3Location'],
+            'AWS::ApiGateway::RestApi': ['S3Location'],
             'AWS::Lambda::Function': ['Code'],
             'AWS::Lambda::LayerVersion': ['Content'],
             'AWS::ElasticBeanstalk::ApplicationVersion': ['SourceBundle'],
-            'AWS::StepFunctions::StateMachine': ['DefinitionS3Location'],
+            'AWS::StepFunctions::StateMachine': ['S3Location'],
         }
 
         exceptions = templated_exceptions.get(parenttype, [])
--- a/src/cfnlint/rules/resources/properties/Properties.py
+++ b/src/cfnlint/rules/resources/properties/Properties.py
@@ -181,6 +181,7 @@ class Properties(CloudFormationLintRule):
             'AWS::Lambda::Function': ['Code'],
             'AWS::Lambda::LayerVersion': ['Content'],
             'AWS::ElasticBeanstalk::ApplicationVersion': ['SourceBundle'],
+            'AWS::StepFunctions::StateMachine': ['DefinitionS3Location'],
         }
 
         exceptions = templated_exceptions.get(parenttype, [])
--- a/src/cfnlint/rules/resources/properties/PropertiesTemplated.py
+++ b/src/cfnlint/rules/resources/properties/PropertiesTemplated.py
@@ -22,7 +22,9 @@ class PropertiesTemplated(CloudFormationLintRule):
         self.resource_property_types.extend([
             'AWS::ApiGateway::RestApi',
             'AWS::Lambda::Function',
+            'AWS::Lambda::LayerVersion',
             'AWS::ElasticBeanstalk::ApplicationVersion',
+            'AWS::StepFunctions::StateMachine',
         ])
 
     def check_value(self, value, path):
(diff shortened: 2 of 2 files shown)


src/cfnlint/rules/resources/properties/Properties.py
```diff
@@ -177,10 +177,11 @@ def check_exceptions(self, parenttype, proptype, text):
             - Start with handling exceptions for templated code.
         """
         templated_exceptions = {
-            'AWS::ApiGateway::RestApi': ['BodyS3Location'],
+            'AWS::ApiGateway::RestApi': ['S3Location'],
```
kddejong, line 180: Isn't this [BodyS3Location](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-restapi.html#cfn-apigateway-restapi-bodys3location) or [BodyS3Location](https://github.com/aws/aws-cli/blob/master/awscli/customizations/cloudformation/artifact_exporter.py#L378)

src/cfnlint/rules/resources/properties/Properties.py
```diff
@@ -177,10 +177,11 @@ def check_exceptions(self, parenttype, proptype, text):
...
             'AWS::Lambda::Function': ['Code'],
             'AWS::Lambda::LayerVersion': ['Content'],
             'AWS::ElasticBeanstalk::ApplicationVersion': ['SourceBundle'],
+            'AWS::StepFunctions::StateMachine': ['S3Location'],
```
kddejong, line 184: And this [DefinitionS3Location](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-stepfunctions-statemachine.html#cfn-stepfunctions-statemachine-definitions3location) or [DefinitionS3Location](https://github.com/aws/aws-cli/blob/master/awscli/customizations/cloudformation/artifact_exporter.py#L422)

kddejong (commented):
I think I'm missing something here @PatMyron

src/cfnlint/rules/resources/properties/Properties.py
PatMyron: Screen Shot 2020-09-08 at 10 54 22 AM

src/cfnlint/rules/resources/properties/Properties.py
PatMyron: Screen Shot 2020-09-08 at 10 54 01 AM

PatMyron: As @asjohnston-asf caught here: ["I don't appreciate any nuance of the property type being S3Location vs DefinitionS3Location"](https://github.com/aws-cloudformation/cfn-python-lint/issues/1683#issuecomment-687662932), the name of the property in the template is actually different than the name of the property type in the [Resource Specification](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html)

From testing, these rules seems to be parsing the name of the property differently because of this, hence the divergence between `templated_exceptions` in the two rules

kddejong (approved):
I got it now.  We are using the resource spec name not the property name.  That was an unfortunate chose but I get why its the way you did this.

ozbillwang:
Could we improve this PR? Looks, it did half job.

The code has been changed from E3002 to W3002, 

```
W3002 This code may only work with `package` cli command as the property
```

it is ok to get warning as this information (start with W means warning, not error, right?), but don't make it as failed with exist status code 4

```
$ echo $?
4
```

PR #1685, opened 2020-09-08
CloudFormation Resource Specification 18.2.0

kddejong:
*Issue #, if available:*

*Description of changes:*
```
scripts/update_specs_from_pricing.py # requires Boto3 and Credentials
scripts/update_specs_services_from_ssm.py # requires Boto3 and Credentials
cfn-lint --update-specs # https://github.com/aws-cloudformation/cfn-python-lint/pull/1383\#issuecomment-629891506
cfn-lint --update-iam-policies
cfn-lint --update-documentation
```

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1697, opened 2020-09-14
Update resource spec to 18.3.0

kddejong:
*Issue #, if available:*

*Description of changes:*
- Update CloudFormation Specs to 18.3.0


PatMyron:
looks like same failure reasons as https://github.com/aws-cloudformation/cfn-python-lint/pull/1685

src/cfnlint/rules/resources/ApproachingLimitName.py
```diff
@@ -8,7 +8,7 @@
 
 class LimitName(CloudFormationLintRule):
     """Check maximum Resource name size limit"""
-    id = 'I3011'
+    id = 'I3012'
```
PatMyron, line 11: noticed this but wasn't sure how changing existing rule IDs would work

luckily, this should be one of the rarer rules, so hopefully there aren't many existing `--ignore-checks` for it

kddejong: I figured this one was less used than the other `I3011` rule

src/cfnlint/data/ExtendedSpecs/all/01_spec_patch.json
```diff
@@ -160,5 +174,150 @@
...
+          "Documentation": "http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-route53-healthcheck-healthcheckconfig.html#cfn-route53-healthcheck-healthcheckconfig-type",
+          "PrimitiveType": "String",
+          "Required": true,
+          "UpdateType": "Immutable",
+          "Value": {
+            "ValueType": "Route53HealthCheckConfigType"
+          }
+        }
+      }
+    }
```
PatMyron, line 321: We should work to get these fixed upstream so we don't have to maintain patches this big forever

kddejong: Absolutely agreed there. I don't like doing this patching.  Since things are kind of broken with a feature we provide I feel the need to patch it for now.

PatMyron: Also not sure if `ValueType` should be included in these patches since [that's for Linter-specific specific constraints](https://github.com/aws-cloudformation/cfn-python-lint/issues/903#issuecomment-595356853)

kddejong: Agreed.  That probably should be separate.

Issue #1043, opened 2019-07-20
Better checking when using Ref AWS::NoValue


kddejong:
When doing a [`!Ref AWS::NoValue`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/pseudo-parameter-reference.html#cfn-pseudo-param-novalue) without a condition we need to validate these remaining rules still work and handle things properly.

- [x] [E3003](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/rules/resources/properties/Required.py) - Required
- [x] [E2520](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/rules/resources/properties/Exclusive.py) - Exclusive
- [x] [E2521](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/rules/resources/properties/Inclusive.py) - Inclusive
- [x] [E2522](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/rules/resources/properties/AtLeastOne.py) - At Least One
- [x] [E2523](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/rules/resources/properties/OnlyOne.py) - Only One

kddejong:
This should be covered now.

PR #1717, opened 2020-09-28
Update E2521 to better handle Ref AWS::NoValue

kddejong:
*Issue #, if available:*
#1043

*Description of changes:*
- Update rule E2521 to do better handling of Ref [`AWS::NoValue`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/pseudo-parameter-reference.html#cfn-pseudo-param-novalue)

### DIFF
--- a/src/cfnlint/data/AdditionalSpecs/Inclusive.json
+++ b/src/cfnlint/data/AdditionalSpecs/Inclusive.json
@@ -12,22 +12,6 @@
       "IamCertificateId": [
         "SslSupportMethod"
       ]
-    },
-    "AWS::EC2::SecurityGroup.Ingress": {
-      "FromPort": [
-        "ToPort"
-      ],
-      "ToPort": [
-        "FromPort"
-      ]
-    },
-    "AWS::EC2::SecurityGroup.Egress": {
-      "FromPort": [
-        "ToPort"
-      ],
-      "ToPort": [
-        "FromPort"
-      ]
     }
   },
   "ResourceTypes": {
@@ -65,30 +49,11 @@
       "VpcId": [
         "DefaultSubnetId"
       ]
+    },
+    "AWS::EC2::SecurityGroup": {
+      "SecurityGroupEgress": [
+        "VpcId"
+      ]
     }
-  },
-  "AWS::EC2::SecurityGroup": {
-    "SecurityGroupEgress": [
-      "VpcId"
-    ],
-    "VpcId": [
-      "SecurityGroupEgress"
-    ]
-  },
-  "AWS::EC2::SecurityGroupIngress": {
-    "FromPort": [
-      "ToPort"
-    ],
-    "ToPort": [
-      "FromPort"
-    ]
-  },
-  "AWS::EC2::SecurityGroupEgress": {
-    "FromPort": [
-      "ToPort"
-    ],
-    "ToPort": [
-      "FromPort"
-    ]
   }
 }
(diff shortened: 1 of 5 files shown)


src/cfnlint/data/AdditionalSpecs/Inclusive.json
```diff
@@ -65,30 +49,11 @@
...
         "DefaultSubnetId"
       ]
+    },
+    "AWS::EC2::SecurityGroup": {
+      "SecurityGroupEgress": [
+        "VpcId"
+      ]
     }
-  },
-  "AWS::EC2::SecurityGroup": {
```
kddejong, line 70: These aren't in the right section.

src/cfnlint/data/AdditionalSpecs/Inclusive.json
```diff
@@ -12,22 +12,6 @@
       "IamCertificateId": [
         "SslSupportMethod"
       ]
-    },
-    "AWS::EC2::SecurityGroup.Ingress": {
-      "FromPort": [
```
kddejong, line 17: Same example goes for the rest of these.
If you do `-1` the for protocol you can specify just `FromPort`. So this `FromPort` `ToPort` requirement is only a problem when the Protocol is TCP or UDP

src/cfnlint/rules/resources/properties/Inclusive.py
```diff
@@ -35,10 +35,11 @@ def check(self, properties, inclusions, path, cfn):
 
         property_sets = cfn.get_object_without_conditions(properties)
         for property_set in property_sets:
-            for prop in property_set['Object']:
+            obj = property_set['Object'].clean()
```
kddejong, line 38: This will remove any leftover `Ref: AWS::NoValue`s

src/cfnlint/data/AdditionalSpecs/Inclusive.json
PatMyron: Not sure I follow exactly what you mean. Is there another section they should be in? Or just that they shouldn't be here?

kddejong: Some how it was out of resource types and property types.

test/fixtures/templates/bad/resources/properties/inclusive.yaml
```diff
@@ -20,3 +20,11 @@ Resources:
...
         MasterUserPassword: test
         Engine: test
+  LoadBalancer:
+    Type: AWS::ApplicationAutoScaling::ScalingPolicy
+    Properties:
+      PolicyName: Test
+      ScalableDimension: Test
+      ResourceId: Id
+      PolicyType: StepScaling
+      ServiceNamespace: !Ref 'AWS::NoValue'
```
kddejong, line 30: From Inclusive docs.
```
"ScalableDimension": [
        "ResourceId",
        "ServiceNamespace"
      ],
```

test/unit/rules/resources/properties/test_inclusive.py
```diff
@@ -21,4 +24,4 @@ def test_file_positive(self):
     def test_file_negative(self):
         """Test failure"""
         self.helper_file_negative(
-            'test/fixtures/templates/bad/resources/properties/inclusive.yaml', 2)
+            'test/fixtures/templates/bad/resources/properties/inclusive.yaml', 4)
```
kddejong, line 27: Errors are up by two because the definition of inclusive is for 3 attributes.

Issue #882, opened 2019-05-09
Results summary


phene:
*cfn-lint version: 0.20.0*

It would be great it cfn-lint supported a summary of results about the current run, including:

* How many files were checked
* How many rules failed.

Require flag to enable it.

luistapia6-11:
+1 It would great to get more information at the end of a test. How many rules were checked, how many rules passed/failed. Similar to running an Ansible playbook.

fatbasstard:
Actually like this one. I've been hacking some counters in the past to find the amount of files processed.... Making it a feature is 🎉

jakrob:
+1 It would great to get summary table showing stats for each error/warning types

PatMyron:
> +1 It would great to get summary table showing stats for each error/warning types

@jakrob I've been piping the output into something like this to get the count for rules flagged:

```shell
$ cfn-lint template.yaml | grep -Eo '^\w\d+' | sort | uniq -c | sort -nr
  20 W3011
  16 E3008
  16 E3002
   8 E2531
   8 E2001
   4 W2511
   4 W2501
   1 E0000
```

PR #1742, opened 2020-10-20
Colored Output and Pretty Formatting

kddejong:
*Issue #, if available:* https://github.com/aws-cloudformation/cfn-python-lint/issues/971

*Description of changes:*
- Add pretty formatter that groups outputs by file name and adds color for terminals that support ansi

### DIFF
--- a/src/cfnlint/formatters/__init__.py
+++ b/src/cfnlint/formatters/__init__.py
@@ -17,16 +17,17 @@ class color(object):
     unknown = '\033[37m'
     green = '\033[32m'
     reset = '\033[0m'
-    bold_green = '\033[1;32m'
     bold_reset = '\033[1:0m'
+    underline_reset = '\033[4m'
 
 
 def colored(s, c):
     """ Takes in string s and outputs it with color """
     if sys.stdout.isatty():
         return '{}{}{}'.format(c, s, color.reset)
-    else:
-        return s
+
+    return s
+
 
 class BaseFormatter(object):
     """Base Formatter class"""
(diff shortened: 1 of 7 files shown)


PatMyron:
Wondering about sorting violations by severity instead of just line number:

Screen Shot 2020-10-21 at 1 13 50 PM

src/cfnlint/formatters/__init__.py
```diff
@@ -2,11 +2,29 @@
...
 
 
+class color(object):
+    error = '\033[31m'
+    warning = '\033[33m'
+    informational = '\033[34m'
+    unknown = '\033[37m'
+    green = '\033[32m'
+    reset = '\033[0m'
+    bold_green = '\033[1;32m'
```
PatMyron, line 20: Couldn't find much shell color documentation more authoritative than https://gist.github.com/vratiu/9780109

Any other resource you used?

kddejong: Not really.  https://github.com/tartley/colorama/blob/master/colorama/ansi.py#L49-L58 looked at other color packages too.

kddejong:
Interesting.  Absolutely we could sort by severity.

PatMyron:
Also wondering if [summaries](https://github.com/aws-cloudformation/cfn-python-lint/issues/882) should be dealt with in the same format as [color](https://github.com/aws-cloudformation/cfn-python-lint/issues/971)

kddejong:
Good call.  We could add that to this one.

kddejong:
I got it down to something like this.
`Cfn-lint scanned 11 templates against 127 rules and found 7 errors, 3 warnings, and 0 informational violations`

thoughts?

kddejong:
Current output view.

image

kddejong:
not sure the line and column numbers need to be green either.

kddejong:
Line and column numbers are now white

PatMyron:
Not a big deal, but heads up that tab spacing can lead to cliffs for different number of digits in line numbers:

Screen Shot 2020-10-28 at 12 47 50 PM

kddejong:
@PatMyron fixed.

src/cfnlint/formatters/__init__.py
```diff
@@ -172,3 +188,60 @@ def _format(self, match):
...
+
+class PrettyFormatter(BaseFormatter):
+    """Generic Formatter"""
+
+    def _format(self, match):
+        """Format output"""
+        formatstr = '{0}{1}{2}'
+        pos = '{0}:{1}:'.format(match.linenumber, match.columnnumber)
+        return formatstr.format(
+            colored('{:20}'.format(pos), color.reset),
```
kddejong, line 201: `{:20}` will keep the width without tabs

Issue #1744, opened 2020-10-21
Feature Request: Conditionally require StartingPosition for AWS::Lambda::EventSourceMapping


iamed2:
[`StartingPosition`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-eventsourcemapping.html#cfn-lambda-eventsourcemapping-startingposition) is required for certain source types (Amazon Kinesis, Amazon DynamoDB, and Amazon MSK Streams): 

Deploying without specifying StartingPosition results in an error like:
```
validation error detected: Value null at 'startingPosition' failed to satisfy constraint: Member must not be null. (Service: Lambda, Status Code: 400, Request ID: 7dfa2c0c-8ea6-4d87-a8ef-e9e3d6a50e23, Extended Request ID: null)
```

It would be helpful to get a better error earlier, from cfn-lint.

PatMyron:
[don't think this fits into any of our existing concise constraint formats](https://github.com/aws-cloudformation/cfn-python-lint/issues/903#issuecomment-595356853), so it would require a new rule
seems like [`EventSourceArn`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-eventsourcemapping.html#cfn-lambda-eventsourcemapping-eventsourcearn) would have to be parsed

kddejong:
Yea, agreed.  I think this has come up enough that it would be good to have a generic rule to do this.  I think the 3 ways I've seen this.
1. A property is required based on a different properties value (within the same property layer)
2. A value of a property has to be a certain value based on a different properties value (within the same property layer)
3. A property in another resource has values that may depend on another resources property value. 

Things get a littler harder and taxing when we have to cross resource or property layers

iamed2:
Thanks so much!

PR #1712, opened 2020-09-23
AWS::Config::ConfigRule.Description length

jbpratt:
fixes #1710

*Description of changes:*
Adding validation to ConfigRule.Description length
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-config-configrule.html#cfn-config-configrule-description

```
Minimum: 0
Maximum: 256
```

Maybe this should be set to 0, but I saw `StringMin: 1` and assumed this
was fine.

example template:
```yaml
AWSTemplateFormatVersion: "2010-09-09"

Resources:
  AccessKeysRotatedConfigRule:
    Type: "AWS::Config::ConfigRule"
    Properties:
      ConfigRuleName: InfoSecAccessKeysRotated
      Description: |
        This is a test on longer description This is a test on longer description This is a test on longer description This is a test on longer description This is a test on longer description This is a test on longer description This is a test on longer description This is a test on longer description This is a test on longer description This is a test on longer description This is a test on longer description This is a test on longer description This is a test on longer description This is a test on longer description This is a test on longer description This is a test on longer description This is a test on longer description.
      InputParameters:
        maxAccessKeyAge: 90
      MaximumExecutionFrequency: One_Hour
      Source:
        Owner: AWS
        SourceIdentifier: ACCESS_KEYS_ROTATED
```

@PatMyron note, this is still not currently working. I may have incorrectly defined the path for the config rule. This can be marked as a draft if desired.

Thanks for working with me on this.

Issue #1439, opened 2020-03-25
New rule for when ZipFile can be used based on Runtime


kddejong:
New rule for Lambda Function.

`ZipFile` can only be used when `Runtime` is set to either of nodejs8.10, nodejs10.x, nodejs12.x, python2.7, python3.6, python3.7.

PatMyron:
[I think this one should be fixed in CloudFormation itself](https://github.com/aws-cloudformation/aws-cloudformation-coverage-roadmap/issues/80#issuecomment-604138808)

kddejong:
I think that is okay.  Do you think that would ever cover all languages?  I'm not sure ZipFile would  work with Go for example.

kddejong:
Another issue `Invalid request provided: Using the ZipFile attribute requires a runtime to be specified`

kddejong:
Fixed with #2676

Issue #1059, opened 2019-07-29
AWS::CloudWatch::Alarm.Period >= 60 for metrics in the AWS/ namespace


irvingpop:
*cfn-lint version: (`cfn-lint --version`)*
```
cfn-lint 0.22.3
```

*Description of issue.*

AWS won't let you set [`AWS::CloudWatch::Alarm.Period`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cw-alarm.html#cfn-cloudwatch-alarms-period) less than `60`, it would be neat if cfn-lint flagged that.

Please provide as much information as possible:

The following resource should get flagged because the Period property is invalid, but doesn't:
```
  ELB5XXExceeded:
    Type: AWS::CloudWatch::Alarm
    Properties:
      AlarmName: !Sub AppELB_5XX_Exceeded-${AWS::StackName}
      AlarmDescription: Alarms when an 5xx requests exceed a specified threshold.
      AlarmActions: [!Ref AlertNotificationTopic]
      MetricName: HTTPCode_ELB_5XX_Count
      Namespace: AWS/ApplicationELB
      ComparisonOperator: GreaterThanOrEqualToThreshold
      EvaluationPeriods: 2
      Period: 30
      Statistic: Sum
      Threshold: 10
      Dimensions:
        - Name: LoadBalancer
          Value: !GetAtt ChefALB.LoadBalancerFullName
```

Cfn-lint uses the [CloudFormation Resource Specifications](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html) as the base to do validation. These files are included as part of the application version. Please update to the latest version of `cfn-lint` or update the spec files manually (`cfn-lint -u`)

PatMyron:
Good catch, reproduced:

`Period (30) not supported for namespaces with the following prefix: AWS/ (Service: AmazonCloudWatch; Status Code: 400; Error Code: ValidationError; Request ID: )`

```yaml
Resources:
  Alarm:
    Type: AWS::CloudWatch::Alarm
    Properties:
      ComparisonOperator: GreaterThanOrEqualToThreshold
      EvaluationPeriods: 2
      Period: 30
      MetricName: HTTPCode_ELB_5XX_Count
      Namespace: AWS/ApplicationELB
```

Wondering if that limitation is specific to the `AWS/` prefix from that error message

PR #1746, opened 2020-10-22
add rule E3017 to check when a prop is required based on a value

kddejong:
*Issue #, if available:*
#1744 
*Description of changes:*
- Add rule E3017 to validate when a property is required based on another property having a certain value


src/cfnlint/data/AdditionalSpecs/RequiredBasedOnValue.json
```diff
@@ -0,0 +1,22 @@
...
+                        "AWS::MSK::Cluster"
+                    ],
+                    "RequiredProperties": [
+                        "StartingPosition"
+                    ]
+                }
+            ]
+        }
+    }
+}
```
PatMyron, line 22: Like reading directly from [`AdditionalSpecs`](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/data/AdditionalSpecs) and avoiding the patching process of [`ExtendedSpecs`](https://github.com/aws-cloudformation/cfn-python-lint/blame/master/src/cfnlint/data/ExtendedSpecs/) into [`CloudSpecs`](https://github.com/aws-cloudformation/cfn-python-lint/tree/master/src/cfnlint/data/CloudSpecs) and the ability to provide patterns in addition to literal strings 👍

Dislike JSON, separate `ResourceTypes`/`PropertyTypes` sections (should be inferable from type name alone and has caused confusion in the past: https://github.com/aws-cloudformation/cfn-python-lint/pull/1608/, https://github.com/aws-cloudformation/cfn-python-lint/pull/1654, https://github.com/aws-cloudformation/cfn-python-lint/pull/1712), and mapping [`Ref`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-ref.html)/[`GetAtt`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-getatt.html) per constraint

We should optimize for [concise constraints](https://github.com/aws-cloudformation/cfn-python-lint/issues/903#issuecomment-595356853) at the expense of potentially more one-time parsing code

```bash
# just an example possibility for discussion
StartingPosition if AWS::Lambda::EventSourceMapping.EventSourceArn == arn:aws*:(kinesis|kafka|dynamodb):*
```

src/cfnlint/data/AdditionalSpecs/RequiredBasedOnValue.json
```diff
@@ -0,0 +1,22 @@
+{
+    "PropertyTypes": {},
+    "ResourceTypes": {
+        "AWS::Lambda::EventSourceMapping": {
+            "EventSourceArn": [
+                {
+                    "Regex": "arn:aws:(kinesis|kafka|dynamodb):*",
```
PatMyron, line 7: [missing support for other partitions](https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html#arns-syntax)
```suggestion
                    "Regex": "arn:aws*:(kinesis|kafka|dynamodb):*",
```

PatMyron:
Think investigating similar rules/requests will be helpful:
[CloudWatch](https://github.com/aws-cloudformation/cfn-python-lint/issues/1059), [DynamoDB](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/rules/resources/dynamodb/BillingMode.py), [EC2](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/rules/resources/ectwo/Ebs.py), [Lambda](https://github.com/aws-cloudformation/cfn-python-lint/issues/1439), [RDS](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/rules/resources/rds/AuroraDBInstanceProperties.py), [RDS](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/rules/resources/rds/AuroraScalingConfiguration.py), [Route53](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/rules/resources/route53/HealthCheck.py), [Route53](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/rules/resources/route53/RecordSet.py), [Route53](https://github.com/aws-cloudformation/cfn-python-lint/issues/94), [StepFunctions](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/rules/resources/stepfunctions/StateMachine.py)

Something to note is that quite a few of them are the opposite: a certain property _cannot_ be specified based on the value of another property

src/cfnlint/data/AdditionalSpecs/RequiredBasedOnValue.json
kddejong: I like the readability of what you provided but yea we will need to think about that and how to parse it.  Let me try a few things to see if I can make that work well.

PatMyron: Or maybe even just [schematize](https://json-schema.org/) some of these constraint files so we automatically catch mistakes

kddejong: I agree.  I removed the resource types and property types.  Let me schematize it and add testing for that file.

kddejong:
For tracking after this PR is merged I will evaluate if these rules can be deleted and moved to this rule.
https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/rules/resources/rds/AuroraScalingConfiguration.py
https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/rules/resources/rds/AuroraDBInstanceProperties.py
https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/rules/resources/route53/HealthCheck.py

src/cfnlint/data/AdditionalSpecs/RequiredBasedOnValue.json
kddejong: Created a JSON Schema for the file and added testing.

kddejong:
@PatMyron I'm going to look at another rule for doing the opposite check because I'm finding use cases where it could be valuable.

test/unit/specs/test_required_based_on_value.py
```diff
@@ -0,0 +1,25 @@
...
+LOGGER = logging.getLogger('cfnlint.maintenance')
+LOGGER.addHandler(logging.NullHandler())
+
+
+class TestRequiredBasedOnValueSpec(BaseTestCase):
+    """Used for Testing Rules"""
+
+
+    def test_success_sbd_domain_removed(self):
+        """Success removal of SBD Domain form unsupported regions"""
```
PatMyron, line 21: Some outdated copy-pasted code here
```suggestion


    def validate_additional_specs_schema(self):
```

test/unit/rules/resources/properties/test_required_based_on_value.py
```diff
@@ -0,0 +1,27 @@
+"""
+Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+SPDX-License-Identifier: MIT-0
+"""
+from test.unit.rules import BaseRuleTestCase
+from cfnlint.rules.resources.properties.RequiredBasedOnValue import RequiredBasedOnValue  # pylint: disable=E0401
+
+
+class TestRequiredBasedOnValue(BaseRuleTestCase):
+    """Test OnlyOne Property Configuration"""
```
PatMyron, line 10:
```suggestion
```

src/cfnlint/data/AdditionalSpecs/RequiredBasedOnValue.json
PatMyron: @kddejong sorry, got confused by that trailing wildcard. Should be something like this instead:

```diff
+                    "Regex": "arn:aws.*:(kinesis|kafka|dynamodb):",
```

kddejong: Oops.  Resolved it here #1759

PR #1759, opened 2020-10-29
Add new rule E3018 to flag unwanted properties when a certain value exists

kddejong:
*Issue #, if available:* https://github.com/aws-cloudformation/cfn-python-lint/pull/1746#issuecomment-714670093

*Description of changes:*
- New rule E3018 to alert when a property isn't needed when another property has a certain value


kddejong:
For tracking after this PR is merged I will evaluate if these rules can be deleted and moved to this rule.  This is a copy from #1746 comment but these rules more closely align with this new rule.

https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/rules/resources/rds/AuroraScalingConfiguration.py
https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/rules/resources/rds/AuroraDBInstanceProperties.py
https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/rules/resources/dynamodb/BillingMode.py

Issue #546, opened 2019-01-02
Does not lint deeply nested template paths


phene:
*cfn-lint version: 0.9.2*

Using `**` path globs does not behave as expected. Assume the following directory structure

```yaml
project:
  templates:
    - a:
      - b:
        - c:
          - foo.yaml
          - bar.yaml
        - d:
          - baz.yaml
          - quux.yaml
      - e:
        - ipsum.yaml
        - lorem.yaml
    - f:
      - zyx.yaml
      - quuz.yaml
    - main.yaml
```

If I use the glob format in `.cfnlintrc`:

```yaml
templates:
- project/templates/**/*.yaml
```

The only files that are checked are `project/templates/f/zyx.yaml` and `project/templates/f/quux.yaml`. All others are ignored. I would expect `project/templates/**/*yaml` to match all deeply nested templates as well as `project/templates/main.yaml`.

kddejong:
The easy fix for Python 3.5 or higher is to add `, recursive=True` to https://github.com/awslabs/cfn-python-lint/blob/master/src/cfnlint/config.py#L439 

I'm debating punting on the backwards compatibility of this and just saying we support what glob can do based on the version of Python available.

fatbasstard:
@phene What OS and what Python version are you running?

cmmeyer:
I like @kddejong's solution.  Based on my research, getting recursive globbing working on 3.4 would require moving to something like pathlib.  Adding `recursive=True` gets us compatibility back to 3.5 (which is pretty good) without the need for any significant code changes. I'm going to go ahead and approve it.

phene:
This is still not behaving appropriately:

```yaml
a: 
  - b:
    - c.yaml
  - d.yaml
```

I expect `a/**/*yaml` to match both `a/b/c.yaml` and `a/d.yaml`

PatMyron:
@phene could you try [`shopt -s globstar`](https://www.gnu.org/software/bash/manual/html_node/The-Shopt-Builtin.html#:~:text=globstar) beforehand if using bash? [Solved a similar issue](https://github.com/aws-cloudformation/cfn-python-lint/issues/1676#issuecomment-685820851)

phene:
This shouldn't require bash to function. It _should_ just work on systems that just have `/bin/sh` or use zsh.

PR #1775, opened 2020-11-09
globbing doc

Eriner:
*Description of changes:*

Python globing does not seem to work as expected. `**` was not evaluated when passed as a direct argument to `cfn-lint`, as in `os.Command("/usr/bin/cfn-lint", "**/*.template");` using Python 3.8. 

The current documentation stumped me for longer than I would care to admit, making me believe there was an issue in my invocation of the script or argument passing.

As mentioned in this issue, it seems that `cfn-python-lint` relies on shell globing: https://github.com/aws-cloudformation/cfn-python-lint/issues/546

This change makes that clear in the documentation.

If the Python globing is indeed working without shell globing, I would be happy to amend this PR to include the previous Python note. However, as this was not working as expected, I removed the note entirely to prevent future user confusion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,8 @@ Lint all `yaml` files in `path` and all subdirectories (recursive):
 
 - `cfn-lint path/**/*.yaml`
 
-*Note*: Glob in Python 3.5 supports recursive searching `**/*.yaml`.  If you are using an earlier version of Python you will have to handle this manually (`folder1/*.yaml`, `folder2/*.yaml`, etc).
+*Note*: If using sh/bash/zsh, you must enable globbing.
+(`setopt -s globstar` for sh/bash, `setopt extended_glob` for zsh).
 
 ##### Specifying the template as an input stream
 


README.md
```diff
@@ -91,7 +91,8 @@ Lint all `yaml` files in `path` and all subdirectories (recursive):
 
 - `cfn-lint path/**/*.yaml`
 
-*Note*: Glob in Python 3.5 supports recursive searching `**/*.yaml`.  If you are using an earlier version of Python you will have to handle this manually (`folder1/*.yaml`, `folder2/*.yaml`, etc).
+*Note*: If using sh/bash/zsh, you must enable globbing.
+(`setopt -s globstar` for sh/bash, `setopt extended_glob` for zsh).
```
PatMyron, line 95: not sure about `sh`, but think it's `shopt` instead of `setopt` for `bash`:
https://www.gnu.org/software/bash/manual/html_node/The-Shopt-Builtin.html

PatMyron: @kddejong Python 3.5 note is still relevant, right?
https://github.com/aws-cloudformation/cfn-python-lint/pull/549

Eriner: ah yes, this was an error in transcription. `shopt` for bash, `setopt` for zsh.

Eriner: @PatMyron unless there is a test case where this works *without* bash or zsh expansion, I would leave off the Python note. The Python note only conveys meaning if the expansion actually works in Python. Currently, I do not think it does (hence this PR).

kddejong: my `**` seems to work on my setup without that option set using zsh.  I can repeat this using bash so I could see the value of this change.

Issue #1705, opened 2020-09-21
tests fail after 2020-12-31


bmwiedemann:
*cfn-lint version: (`cfn-lint --version`)* 0.36.0

*Description of issue.*

While working on [reproducible builds](https://reproducible-builds.org/) for [openSUSE](https://en.opensuse.org/openSUSE:Reproducible_Builds), I found that our `python-cfn-lint` package fails its tests after 2020-12-31

To reproduce:
```bash
osc checkout openSUSE:Factory/python-cfn-lint && cd $_
osc build --noservice --vm-type=kvm --build-opt=--vm-custom-opt="-rtc base=2020-12-31T00:00:00" standard
```

maybe from [`src/cfnlint/data/AdditionalSpecs/LmbdRuntimeLifecycle.json`](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/data/AdditionalSpecs/LmbdRuntimeLifecycle.json)
```json
    "python2.7": {
        "eol": "2020-12-31",
        "deprecated": "2020-12-31",
        "successor": "python3.8"
    }
```

error log:
```
 ======================================================================
 FAIL: test_templates (integration.test_quickstart_templates.TestQuickStartTemplates)
 Test same templates using integration approach
 ----------------------------------------------------------------------
 Traceback (most recent call last):
   File "/home/abuild/rpmbuild/BUILD/cfn-python-lint-0.30.1/test/integration/test_quickstart_templates.py", line 79, in test_templates
     self.run_module_integration_scenarios(rules)
   File "/home/abuild/rpmbuild/BUILD/cfn-python-lint-0.30.1/test/integration/__init__.py", line 90, in run_module_integration_scenarios
     self.assertEqual(len(expected_results), len(matches), 'Expected {} failures, got {} on {}'.format(
 AssertionError: 40 != 41 : Expected 40 failures, got 41 on test/fixtures/templates/quickstart/openshift.yaml
 
 ======================================================================
 FAIL: test_module_integration (integration.test_quickstart_templates_non_strict.TestQuickStartTemplates)
 Test same templates using integration approach
 ----------------------------------------------------------------------
 Traceback (most recent call last):
   File "/home/abuild/rpmbuild/BUILD/cfn-python-lint-0.30.1/test/integration/test_quickstart_templates_non_strict.py", line 53, in test_module_integration
     self.run_module_integration_scenarios(rules)
[...]

PR #1777, opened 2020-11-12
python2.7 Lambda runtime support extended

PatMyron:
closes https://github.com/aws-cloudformation/cfn-python-lint/issues/1705
[Runtime support policy](https://docs.aws.amazon.com/lambda/latest/dg/runtime-support-policy.html)
[`UPDATE – Oct 20, 2020 – We’re extending the support of Python 2.7 in AWS Lambda until at least June 1, 2021.`](https://aws.amazon.com/blogs/compute/continued-support-for-python-2-7-on-aws-lambda/)


kddejong (approved):

kddejong:
We should probably update our python 2.7 support agreement too.
https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/__main__.py#L19-L21

PR #1780, opened 2020-11-13
Fix bug with Custom:: resources

tdekoning93:
*Description of changes:*
We are having an issue since 38.0 with the extension of our own linter getting the below error;
`E0002 Unknown exception while processing rule E3003: 'NoneType' object has no attribute 'get'`

On the following piece of YAML;
```
Resources:
  CustomResource:
    Type: "Custom::Resource"
    Properties:
      test: "test"
```

At the condition you check if the resource does not exists in the self.resourcetypes. After that it tries to get the custom resource it already saw it was not there and it fails. The result will be None and internally in this function there a get via this None object. I would guess that on your if condition, you should not have the "not" on "resource_type not in self.resourcetypes". This PR at least solves it on our end. 

The change in the PR solves it. If its not intended to be like this or I should instead raise an issue so that you can solve it in a better way, please close the PR. 

@kddejong

### DIFF
--- a/src/cfnlint/rules/resources/properties/Required.py
+++ b/src/cfnlint/rules/resources/properties/Required.py
@@ -96,7 +96,7 @@ class Required(CloudFormationLintRule):
         for resourcename, resourcevalue in cfn.get_resources().items():
             if 'Properties' in resourcevalue and 'Type' in resourcevalue:
                 resource_type = resourcevalue['Type']
-                if resource_type.startswith('Custom::') and resource_type not in self.resourcetypes:
+                if resource_type.startswith('Custom::') and resource_type in self.resourcetypes:
                     resource_type = 'AWS::CloudFormation::CustomResource'
                     matches.extend(
                         self.check_obj(


kddejong:
I think I did mean `not in` in this case.  I'm still looking into what the issue is here and trying to replicate it.  Can you give me a little more explanation on `After that it tries to get the custom resource it already saw it was not there and it fails`.  From what I can tell we reconfigure the Type to `AWS::CloudFormation::CustomResource` and then we get the required attributes for that type.  I can't get the failure with the your example template so I think I'm missing something.

tdekoning93:
You are correct, I'll look further into the issue we are having and raise another PR or a bug report if its related to cfn-pythonlint.

rrezino:
@kddejong in line 99, the statement guarantees that `resource_type` is not on `self.resourcetypes`

Then, at line 105, `self.resourcetypes.get(resource_type))`, we try to get the object again. Which is proven to not exists because of the if condition. So we get back `None`.

With this `None` object, we call, `_get_required_attrs_specs`, which internally invokes `.get` over a None object.

Here we have the error, as a `None` object does not have a function called `.get`

tdekoning93:
Yeah so basically `self._get_required_attrs_specs(self.resourcetypes.get(resource_type))` doesn't exist when the resource is `Custom::`. Trying to print that will give;
E0002 Unknown exception while processing rule E3003: 'NoneType' object has no attribute 'get'

kddejong:
https://github.com/aws-cloudformation/cfn-python-lint/pull/1780/files#diff-6129aac875eed2b84e62d4aa2f37cd1348dbff3f1813ad7553472caa0f657128L100 sets the resource type to `AWS::CloudFormation::CustomResource`.  Basically we are checking to see if a resource starts with `Custom::` so we can validate it like `AWS::CloudFormation::CustomResource`.  I'm trying to replicate the issue.  Is the template you provided cause an issue on your side?

rrezino:
Interesting, even setting `AWS::CloudFormation::CustomResource` our result is `None`, isn't it the case for you?

Screen Shot 2020-11-17 at 8 55 45 AM

rrezino:
I don't see CloudForrmation on my set list. 

Screen Shot 2020-11-17 at 8 58 11 AM

kddejong:
okay looking a little deeper.  What version of Python are you using? 

Validated this is in every region spec so it should be there.
```json
"AWS::CloudFormation::CustomResource": {
      "AdditionalProperties": true,
      "Documentation": "http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cfn-customresource.html",
      "Properties": {
        "ServiceToken": {
          "Documentation": "http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cfn-customresource.html#cfn-customresource-servicetoken",
          "PrimitiveType": "String",
          "Required": true,
          "UpdateType": "Immutable"
        }
      }
    }
```

kddejong:
also are you using the default region or changing it?

rrezino:
Kevin, thanks for the feedback and help.
We figured out that a dev created a custom list of the resources + rules. that's why we didn't have it out of the box. 
Though is nice to have the protection you sent on the last link

kddejong:
Thanks @rrezino let me know if something didn't happen as expected so we can get that cleared up.

PR #1782, opened 2020-11-13
Fix typo

miparnisari:


src/cfnlint/rules/functions/FindInMap.py
```diff
@@ -55,7 +55,7 @@ def map_name(self, map_name, mappings, tree):
                 matches.extend(self.check_dict(map_name, tree[:] + [0]))
             else:
                 if map_name not in mappings:
-                    message = 'Map Name {0} doesnt exist for {0}'
+                    message = 'Map Name {0} doesn\'t exist for {0}'
```
PatMyron, line 58: thoughts on this?
```suggestion
                    message = 'Map Name {0} does not exist for {0}'
```

kddejong (approved):

PR #1785, opened 2020-11-16
Fix parsing issue with !GetAtt

jameskoll-wf:
*Issue #, if available:*
N/A

*Description of changes:*

When a yaml file has !GetAtt with an attribute with a dot in it, it is not parsed correctly.

For example, consider
!GetAtt ProvisionedProduct.Outputs.Example

Before this fix it is parsed as
{"Fn::GetAtt": ["ProvisionedProduct", "Outputs", "Example"]}
Instead of the correct representation
{"Fn::GetAtt": ["ProvisionedProduct", "Outputs.Example"]}

This can be confirmed by running the following:
```
python3 -c "from src.cfnlint.decode import decode; data = decode('test/fixtures/templates/good/functions/getatt.yaml'); print(data[0]['Outputs']['OutputValue']['Value'])"
```


kddejong:
I think this makes sense.  Been playing this one back and forth in my mind and make sure I don't break anything downstream. In general this is how it is considered based on standard getatt structure (2 elements... the resource name and the attribute).  In cases like CloudFormation stacks where the Outputs add dots I like to parse it to as many elements as there are dots.  However consistency here is key.

kddejong (approved):

PR #1646, opened 2020-08-12
excluding AWS::EC2::FlowLog.LogFormat from Sub required rule (E1029)

PatMyron:
fixes https://github.com/aws-cloudformation/cfn-python-lint/issues/1644
[`AWS::EC2::FlowLog.LogFormat`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-flowlog.html#cfn-ec2-flowlog-logformat)

Lot of issues/churn for `E1029`, wonder if we need to re-think it at all?
https://github.com/aws-cloudformation/cfn-python-lint/issues?q=E1029
https://github.com/aws-cloudformation/cfn-python-lint/commits/master/src/cfnlint/rules/functions/SubNeeded.py

PR #1788, opened 2020-11-18
Refactoring E1029 to take exclusion criteria from a JSON spec file

### LINKED ISSUE #1662: E1029 - False positive with multiline YAML string
0.35.0

*Description of issue.*
Using ${iot:Connection.Thing.ThingName} inside an AWS::IoT::Policy PolicyDocument is flagged as an embedded parameter outside of Fn::Sub.

Freshly installed cfn-lint this morning. I'm using Python 2.7 and VS Code

* Sample Template:
```

Resources:
  TestPolicy:
    Type: AWS::IoT::Policy
    Properties:
      PolicyDocument: >
        {
          "Version": "2012-10-17",
          "Statement": [
            {
              "Effect": "Allow",
              "Action": "iot:Connect",
              "Resource": "arn:aws:iot:eu-central-1:123456789012:client/${iot:Connection.Thing.ThingName}"
            }
          ]
        }
      PolicyName: ConnectWithThingName
```

In the second example in their developer guide on AWS IoT they use the exact same policy document. https://docs.aws.amazon.com/iot/latest/developerguide/connect-policy.html

* Feature request:
Like with IAM policies, IoT policies shouldn't have variables flagged as embedded parameters.

Tro95:
*Issue #, if available:* None

*Description of changes:* I moved all of the hard-coded excludes into a new `AdditionalSpecs/SubNeededExcludes.json` spec file, and refactored rule E1029 to use it. Whilst this does not address the need for periodic updates to keep the rule in sync with discovered false-positives, it does mean any future updates can be done solely in the new json file, which in turn opens up opportunities to improve how this file is populated.

See https://github.com/aws-cloudformation/cfn-python-lint/pull/1646 for some background discussion.

Edit: The most recent commit now fixes #1662 also.

### DIFF
--- a/src/cfnlint/data/AdditionalSpecs/SubNeededExcludes.json
+++ b/src/cfnlint/data/AdditionalSpecs/SubNeededExcludes.json
@@ -1,0 +1,181 @@
+{
+    "ResourceTypes": {
+        "AWS::ApiGateway::Method": {
+            "ExcludeRegex": "stageVariables\\..*"
+        },
+        "AWS::AppSync::FunctionConfiguration": {
+            "Properties": {
+                "RequestMappingTemplate": {
+                    "ExcludeAll": true
+                },
+                "ResponseMappingTemplate": {
+                    "ExcludeAll": true
+                }
+            }
+        },
+        "AWS::AppSync::Resolver": {
+            "Properties": {
+                "RequestMappingTemplate": {
+                    "ExcludeAll": true
+                },
+                "ResponseMappingTemplate": {
+                    "ExcludeAll": true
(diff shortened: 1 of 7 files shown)


Tro95:
With the most recent change, this closes #1662

kddejong (approved):

PatMyron:
Properties like `PolicyDocument`, `UserData`, `RequestMappingTemplate`, `ResponseMappingTemplate` seem the same across resource types. Wondering if we could just treat them the same regardless of resource type rather than needing to specify every resource type with these properties

PR #1791, opened 2020-11-19
Make sure types are strings before assuming they are

### LINKED ISSUE #1521: Need to validate that resource type is a string
*cfn-lint version: (`cfn-lint --version`)* 0.30.1

*Description of issue.*

image

image

kddejong:
*Issue #, if available:*
fix #1521
*Description of changes:*
- Fix an issues with get valid getatts that was an issue if the type wasn't a string
- fix an issue with rule E3038 to make sure types are strings before assuming they are


kddejong:
The reason I don't like this fix is we can run into it again. When resources register for the resource type the likely hood of this is less.  Also doing the fix in get_resources should allow many rules to benefit from it.

PR #1535, opened 2020-05-17
adding Los Angeles, ElastiCache CacheNodeType, Elasticsearch InstanceType, and fixing Neptune InstanceClass pricing information

PatMyron:
**[Offer files for reference:](https://docs.aws.amazon.com/awsaccountbilling/latest/aboutv2/using-ppslong.html#download-the-offer-index)**
https://pricing.us-east-1.amazonaws.com/offers/v1.0/aws/index.json
[AmazonEC2](https://pricing.us-east-1.amazonaws.com/offers/v1.0/aws/AmazonEC2/current/index.json)
[AmazonRedshift](https://pricing.us-east-1.amazonaws.com/offers/v1.0/aws/AmazonRedshift/current/index.json)
[AmazonDAX](https://pricing.us-east-1.amazonaws.com/offers/v1.0/aws/AmazonDAX/current/index.json)
[AmazonMQ](https://pricing.us-east-1.amazonaws.com/offers/v1.0/aws/AmazonMQ/current/index.json)
[AmazonRDS](https://pricing.us-east-1.amazonaws.com/offers/v1.0/aws/AmazonRDS/current/index.json)
[AmazonNeptune](https://pricing.us-east-1.amazonaws.com/offers/v1.0/aws/AmazonNeptune/current/index.json)
[AmazonDocDB](https://pricing.us-east-1.amazonaws.com/offers/v1.0/aws/AmazonDocDB/current/index.json)
[AmazonElastiCache](https://pricing.us-east-1.amazonaws.com/offers/v1.0/aws/AmazonElastiCache/current/index.json)
[AmazonES](https://pricing.us-east-1.amazonaws.com/offers/v1.0/aws/AmazonES/current/index.json)

---

[unrelated to this change, `m6g` instance types launched this week](https://aws.amazon.com/blogs/aws/new-m6g-ec2-instances-powered-by-arm-based-aws-graviton2/)

---

moved [`US West (Los Angeles)` from `region_exceptions` to `region_map` for `us-west-2`](https://aws.amazon.com/blogs/aws/aws-now-available-from-a-local-zone-in-los-angeles/)

which adds `i3en.metal-2tb` EC2 instance type to `src/cfnlint/data/ExtendedSpecs/us-west-2/05_pricing_property_values.json` and 37 lines to `src/cfnlint/data/AdditionalSpecs/RdsProperties.json`

---

[...]

PR #1792, opened 2020-11-20
automating maintenance with Github actions

PatMyron:
[`Encrypted secrets`](https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets) for IAM actions currently required:
https://github.com/aws-cloudformation/cfn-python-lint/blob/1ad5de337c3a388d949c30b738e48d15fae2092b/scripts/update_specs_services_from_ssm.py#L131
https://github.com/aws-cloudformation/cfn-python-lint/blob/1ad5de337c3a388d949c30b738e48d15fae2092b/scripts/update_specs_from_pricing.py#L90

---

[`GitHub Actions usage is free for public repositories`](https://docs.github.com/en/free-pro-team@latest/github/setting-up-and-managing-billing-and-payments-on-github/about-billing-for-github-actions)

[`Shortest interval you can run scheduled workflows is once every 5 minutes`](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#onschedule) ([`takes while before schedules actions run at all in a new repo`](https://github.community/t/schedule-workflows-missing/17653/4))

[`peter-evans/create-pull-request`](https://github.com/peter-evans/create-pull-request#usage)

tested in my fork since I wasn't sure how to test in this repo without pushing


.github/workflows/maintenance.yaml
```diff
@@ -0,0 +1,28 @@
+on:
+  schedule:
+    - cron: '*/15 * * * *'
```
kddejong, line 3: This seems aggressive for how frequently these things change.

PatMyron: It won't change anything if nothing has changed, and it'll just update the same PR if anything has changed, so I considered running frequently and we could just merge it right before releases. Every 15 minutes seems to still be getting `ThrottlingException when calling the GetProducts operation` so trying every 20 minutes now

PatMyron: We might be able to use [pricing files](https://github.com/aws-cloudformation/cfn-python-lint/pull/1535) instead of the pricing API?

jotompki (approved):

kddejong (approved):

.github/workflows/maintenance.yaml
PatMyron: [can also temporarily disable Github actions](https://github.com/aws-cloudformation/cfn-python-lint/actions?query=workflow%3A.github%2Fworkflows%2Fmaintenance.yaml) if we need to patch [Resource Specification](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html) bugs any week

PatMyron: @miparnisari / anyone else receiving `Run failed: .github/workflows/maintenance.yaml` emails in their fork because they don't have `secrets.AWS_ACCESS_KEY_ID`/`secrets.AWS_SECRET_ACCESS_KEY` set: [GitHub Actions notifications can be disabled](https://github.com/settings/notifications) (doesn't look more configurable yet), individual workflows can be disabled, or [encrypted](https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets) AWS credentials set

kddejong:
```json
[
  "ssm:GetParametersByPath",
  "iam:ListPolicies",
  "pricing:GetProducts"
]
```

PR #1795, opened 2020-11-20
Get the rest of the user data items

kddejong:
*Issue #, if available:* https://github.com/aws-cloudformation/cfn-python-lint/pull/1788

*Description of changes:*
- Update rule E1029 spec file to handle  all UserData items. Fix typos in `AWS::EMR::InstanceGroupConfig`

### DIFF
--- a/src/cfnlint/data/AdditionalSpecs/SubNeededExcludes.json
+++ b/src/cfnlint/data/AdditionalSpecs/SubNeededExcludes.json
@@ -63,23 +63,54 @@
                 }
             }
         },
+        "AWS::EC2::SpotFleet": {
+            "Metadata": {
+                "AWS::CloudFormation::Init": {
+                    "ExcludeAll": true
+                }
+            },
+            "Properties": {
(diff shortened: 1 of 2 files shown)


src/cfnlint/data/AdditionalSpecs/SubNeededExcludes.json
```diff
@@ -63,23 +63,54 @@
...
+                    }
+                }
+            }
+        },
         "AWS::EMR::InstanceGroupConfig": {
             "Properties": {
                 "AutoScalingPolicy": {
                     "Properties": {
-                        "ScalingRule": {
+                        "Rules": {
```
kddejong, line 101: We need to follow the actual Json/Yaml path and not the spec path because those items don't always match.

Tro95: Good catch! Apologies, it wasn't straight forward to reverse some of the hard-coded excludes originally in the rule, and clearly I missed the property name not aligning with the sub property type.

PR #1796, opened 2020-11-20
Global Excludes for E1029

Tro95:
*Issue #, if available:* Addressing @PatMyron 's comments in #1788.

*Description of changes:* Added a `GlobalPropertyTypes` field in the specs file, allowing commonly-occuring properties such as `UserData` and `PolicyDocument` to be matched globally, rather than on a per-resource type basis.


kddejong:
@Tro95 does this replace #1795 ?

Tro95:
> @Tro95 does this replace #1795 ?

Yes, it does now. The `GlobalPropetyTypes` succeeds all the UserData additions you added, along with the ones that were originally there, and I've since merged in the fixes you did to the specs file and added the test you did.

src/cfnlint/data/AdditionalSpecs/SubNeededExcludes.json
```diff
@@ -56,89 +29,24 @@
...
                     "Properties": {
-                        "ScalingRule": {
+                        "Rules": {
                             "Properties": {
-                                "ScalingTrigger": {
+                                "Trigger": {
                                     "Properties": {
-                                       "CloudWatchAlarmDefinition": {
+                                        "CloudWatchAlarmDefinition": {
                                             "ExcludeAll": true
```
PatMyron, line 43: [`AWS::EMR::Cluster.ScalingTrigger.CloudWatchAlarmDefinition`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-elasticmapreduce-cluster-scalingtrigger.html#cfn-elasticmapreduce-cluster-scalingtrigger-cloudwatchalarmdefinition)

https://github.com/aws-cloudformation/cfn-python-lint/pull/1788#issuecomment-731345937 was just meant to be a few examples rather than an exhaustive list

I'm not sure I understand the benefit over the previous format:

https://github.com/aws-cloudformation/cfn-python-lint/blob/52c02962e90354b265fec839e71e1d038e899fd4/src/cfnlint/rules/functions/SubNeeded.py#L21-L23

since it's now hundreds of line instead with decreased coverage and increased maintenance

Tro95: Prior to #1788, there was a lot of custom logic to handle things such as IAM and IoT policies, where there's a specific list of permitted substitutions the rule should not match, and API gateway substitutions, where there's a regex to match permitted substitutions. I decided to make the rule more tolerant to such use-cases, and that moving the exclusions into a separate configuration file was a sensible addition. Should the resource provider schema ever include such information itself, then the rule is in a much better position to take advantage of it.

I fully recognise the specs file is a lot bigger than the original exclusion variables, however I think it's also a lot more descriptive, and I disagree around the decreased coverage and increased maintenance. With this PR, the specs file allows you to exclude a property regardless of what resource or path within a resource it is (using the `GlobalPropertyTypes` key), so I just need to move the `CloudWatchAlarmDefinition` block to there to cover both the `AWS::EMR::Cluster` and `AWS::EMR::InstanceGroupConfig`.

Maybe it makes more sense for everything to be under `GlobalPropertyTypes` rather than `ResourceTypes`, however I was keen to avoid false-negatives if two resources have a property with the same name, but with different behaviours. Moving all the current blocks to under `GlobalPropertyTypes` will emulate the original `excludes` variable's behaviour.

kddejong (approved):

src/cfnlint/data/AdditionalSpecs/SubNeededExcludes.json
PatMyron: > Should the resource provider schema ever include such information itself

This is unlikely btw

> With this PR, the specs file allows you to exclude a property regardless of what resource or path within a resource it is

believe this was already the case

> however I was keen to avoid false-negatives if two resources have a property with the same name, but with different behaviours

Agree linters should strive for false positives over false negatives, but I think this is actually more likely to cause false negatives than simply excluding all properties with the same name

PatMyron: same issue with [`AWS::CodeBuild::Project.Source.BuildSpec`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-codebuild-project-source.html#cfn-codebuild-project-source-buildspec) not working anymore

PatMyron:
@kddejong had an idea I like that I think we should consider:
"instead of looking for `${anything}`, we look for things we know need [`Sub`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-sub.html), like a resource name or parameter name"

src/cfnlint/data/AdditionalSpecs/SubNeededExcludes.json
```diff
@@ -56,90 +29,6 @@
...
-                        "transfer:HomeBucket",
-                        "transfer:HomeDirectory",
-                        "transfer:HomeFolder",
-                        "transfer:UserName",
-                        "redshift:DbUser",
-                        "cognito-identity.amazonaws.com:aud",
-                        "cognito-identity.amazonaws.com:sub",
-                        "cognito-identity.amazonaws.com:amr"
-                    ]
-                }
```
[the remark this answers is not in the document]
kddejong: Move `TemplateBody` to global exceptions.  It is used across 4 different resource types today.

kddejong (changes requested):

src/cfnlint/data/AdditionalSpecs/SubNeededExcludes.json
```diff
@@ -169,5 +58,45 @@
...
+        },
+        "RequestMappingTemplate": {
+            "ExcludeAll": true
+        },
+        "ResponseMappingTemplate": {
+            "ExcludeAll": true
+        },
+        "CloudWatchAlarmDefinition": {
+            "ExcludeAll": true
+        }
```
kddejong, line 100: IAM Policy conditions are meant to be here too.

```suggestion
        },
        "Condition": {
            "ExcludeAll": true
        }
```

src/cfnlint/data/AdditionalSpecs/SubNeededExcludes.json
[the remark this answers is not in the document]
PatMyron: `Amplify` has a couple more `BuildSpec` properties if theirs should be ignored as well. Would rather not keep track of these individually

kddejong:
I was getting kind of close to this code over the weekend.  I started by looking into some better logic around Ref/GetAtt.  We already have functions that build the valid REFs and GetAtts (GetAtts have wild cards because of custom or stack resources).  I do think this would allow us to remove most of the exceptions we have today (we still may need one for TemplateBody?).  We would lose a Ref or GetAtt typo but I don't think we should worry about that.  Basically we would be looking for a double error (missing Sub and a resource Ref/GetAtt typo).

src/cfnlint/data/AdditionalSpecs/SubNeededExcludes.json
Tro95: Are IAM policy conditions not already matched by the `PolicyDocument` block on Line 63?

Tro95:
I've updated the specs file to address some of your comments, but it seems there's a broader concern of regressions here, as well as this PR (+ #1788) not really improving anything.

I see a few options:

1. Once you're satisfied, we can merge this PR. I am hopeful that there are no regressions, but without 100% test coverage of the entire CloudFormation resource space we cannot guarantee this. I have added to the test coverage along the way, but there's still every chance of missing something.
2. I can move all the existing exclusion critera in the `ResourceTypes` block into the global blocks, which will be equivalent to the existing behaviour prior to #1788. This will mean there are no regressions, at the cost of a small chance for false-negatives. Should false-negatives be discovered, we can easily modify the offending exclusion criteria back to the `ResourceTypes` block to handle it appropriately. This is perhaps the best option if we wish to merge this PR.
3. We unmerge #1788 and explore a completely different way of doing this rule, such as being smarter about what a property's value should be. It sounds like both of you are leaning more towards this option, although I'm not clear on how the rule should work.

Do you think this is a fair assessment of the options, and do you see any additional options?

On a personal level, I'm happy to work towards Option 1 and 2, and if we can get alignment on a clear technical direction then also Option 3.

Let me know what your thoughts are.

kddejong:
I think that summarizes it well.  I am a little bit concerned about some regression but going through this I'm starting to feel comfortable that there shouldn't be any with these changes merged in. I think the other concern is will this make it easier to manage going forward.

That being said this rule has been one of our more noisy rules so I'm leaning more towards option 3. I don't think that should prevent us from merging this and moving forward until option 3 is ready. I'm open to feedback if option 3 would even make sense.

src/cfnlint/data/AdditionalSpecs/SubNeededExcludes.json
kddejong: Good question.  `Condition` was a really early add so I wonder if it still even applies.

kddejong:
@tro95 apologize here.  Based on some the flux I reverted the original PR.  I have it backed up so we can resubmit it but we are working to get a release out the door and got a little nervous on this one.  Lets circle back later today or tomorrow.  #1802

Tro95:
> @Tro95 apologize here. Based on some the flux I reverted the original PR. I have it backed up so we can resubmit it but we are working to get a release out the door and got a little nervous on this one. Lets circle back later today or tomorrow. #1802

That's fine, I fully understand and would do the same in your place. This PR contains all the commits in #1788 as well, so once I've fixed the conflicts we can just work on this. I wanted @PatMyron to review #1788 before merging, but only people with write permissions on the repository can request specific people for review, so in future I'll ask in the PR description if I think it needs a deeper review.

src/cfnlint/data/AdditionalSpecs/SubNeededExcludes.json
Tro95: It used to be `Resource`, `NotResource`, and `Condition`, however #1662 shows that this wasn't sufficient as policies can be embedded as a JSON string in the `PolicyDocument` property. I merged the three and put it at the `PolicyDocument` level, which will also match all sub-properties of a `PolicyDocument` if expressed as a JSON or YAML document rather than an embedded string.

PatMyron:
replaced by https://github.com/aws-cloudformation/cfn-lint/pull/2031

PR #1800, opened 2020-11-24
support modules

miparnisari:
*Issue #, if available:* N/A

*Description of changes:* add support for newly-launched modules


kddejong (approved):

src/cfnlint/rules/resources/Configuration.py
```diff
@@ -62,7 +62,6 @@ def _check_resource(self, cfn, resource_name, resource_values):
             ))
 
         resource_type = resource_values.get('Type', '')
-        check_attributes = []
```
PatMyron, line 65: @miparnisari why was this removed?

miparnisari: PyCharm reported that it was unused

PR #1808, opened 2020-11-29
Loosen version sepcifier of importlib_resources

ashb:
Similar to #1479, but there is now a version 2 and version 3, neither of the
"breaking" changes affects the use in this project (of just loading text files) so
is okay for us to use.


kddejong (approved):

setup.py
```diff
@@ -51,7 +51,7 @@ def get_version(filename):
         'jsonschema~=3.0',
         'pathlib2>=2.3.0;python_version<="3.4"',
         'importlib_resources~=1.0.2;python_version=="3.4"',
-        'importlib_resources~=1.4;python_version<"3.7" and python_version!="3.4"',
+        'importlib_resources>=1.4,<4;python_version<"3.7" and python_version!="3.4"',
```
PatMyron, line 54: Does version 4 have breaking changes?

ashb: Version 4 hasn't been released yet, 3.x is the current latest -- this is just being safe as we don't know what the breaking change will be when it comes.

PR #1810, opened 2020-11-30
AWS::DocDB::DBCluster.EngineVersion AllowedValues expansion (4.0)

troyfactor4:
*Issue #, if available: n/a

*Description of changes: Adds DocumentDB 4.0 to the allowed Engine types

### DIFF
--- a/src/cfnlint/data/CloudSpecs/ap-east-1.json
+++ b/src/cfnlint/data/CloudSpecs/ap-east-1.json
@@ -44854,7 +44854,8 @@
     },
     "DocumentDBEngineVersion": {
       "AllowedValues": [
-        "3.6.0"
+        "3.6.0",
+        "4.0"
       ]
     },
     "DocumentDBInstanceClass": {
--- a/src/cfnlint/data/CloudSpecs/ap-east-1.json
+++ b/src/cfnlint/data/CloudSpecs/ap-east-1.json
@@ -44854,8 +44854,7 @@
     },
     "DocumentDBEngineVersion": {
       "AllowedValues": [
-        "3.6.0",
-        "4.0"
+        "3.6.0"
       ]
     },
     "DocumentDBInstanceClass": {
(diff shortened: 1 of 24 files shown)


kddejong:
@troyfactor4 if you make the change here https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/data/ExtendedSpecs/all/03_value_types.json#L166-L170 we can do the rest of the changes you made.  We use JSON patching to patch in the values to the specs as we do the update.

PatMyron:
https://github.com/aws-cloudformation/cfn-python-lint/blob/620d14ede12e340497580b47124c06d856b06104/src/cfnlint/data/ExtendedSpecs/all/03_value_types.json#L166-L169

 Also from looking at the documentation, do you know if it should be `4.0` instead of [`4.0.0`](https://docs.aws.amazon.com/documentdb/latest/developerguide/db-cluster-create.html#db-cluster-create-con)?

troyfactor4:
@kddejong - gotcha. Figured there had to be a better way. There it is.

@PatMyron - yeah I've been using "4.0" in a few scripts over the last few days and it works. Documented here: "https://docs.aws.amazon.com/documentdb/latest/developerguide/compatibility.html#compatibility-getstarted"

troyfactor4:
@PatMyron - just checked out the page you linked on 4.0.0. Interesting, I wonder if both values work?

PatMyron:
> @PatMyron - just checked out the page you linked on 4.0.0. Interesting, I wonder if both values work?

Yeah both of these worked, so maybe add both:
```yaml
Resources:
  DocDBDBCluster:
    Type: AWS::DocDB::DBCluster
    Properties:
      MasterUsername: MasterUsername
      MasterUserPassword: MasterUserPassword
      EngineVersion: 4.0.0
```
```yaml
Resources:
  DocDBDBCluster:
    Type: AWS::DocDB::DBCluster
    Properties:
      MasterUsername: MasterUsername
      MasterUserPassword: MasterUserPassword
      EngineVersion: 4.0
```

troyfactor4:
Done. I've reverted all of those initial changes and added both "4.0" and 4.0.0" to the json patching file.

PatMyron (approved):

PR #1812, opened 2020-12-01
Iam sid errors

jnewbigin:
*Issue #, if available:*

*Description of changes:*
Validate IAM Sid against the specified regex and check for uniqueness in the policy document.

I based it on the IAM Version validator. I can't say I really understand how that works so happy to make changes.


src/cfnlint/rules/resources/iam/Sid.py
```diff
@@ -0,0 +1,95 @@
...
+            'AWS::IAM::ManagedPolicy': 'PolicyDocument',
+            'AWS::IAM::Policy': 'PolicyDocument',
+            'AWS::IAM::Role': 'Policies',
+            'AWS::IAM::User': 'Policies',
+        }
+        for resource_type in self.resources_and_keys:
+            self.resource_property_types.append(resource_type)
+        for resource_type in self.idp_and_keys:
+            self.resource_property_types.append(resource_type)
+        self.sid_regex = re.compile('^[A-Za-z0-9]*$')
```
benbridts, line 40: SIDs for ManagedPolicies have a different regex than SIDs in an inline policy, IIRC. I believe spaces are allowed in an inline policy SID

PatMyron: Similar discussion here: https://github.com/aws-cloudformation/cfn-python-lint/issues/708

jnewbigin: I started working off the cloudformation error I revieved on my stack.
Then I turned to the AWS docs to try and find more details.
The docs are pretty vague - particularly for resource policies
If I get a chance I'll set up some test templates and try them out. Seems to be where #708 got to

jnewbigin: I can confirm that ECR and KMS policies do not limit to that regex or require uniqueness.
I have not tested the other types but it seems that the rules are definitely different between resource types.
Given that, I think for now these rules should only apply to IAM resources.
I will update my PR to reflect that

jnewbigin: Still to confirm the correctness for inline policies

lobinator:
any progress planned on this?

Risae:
I believe this has already been implemented with https://github.com/aws-cloudformation/cfn-lint/pull/3982?

Issue #1813, opened 2020-12-01
False alarm -  E8003 Every Fn::Equals object requires a list of 2 string parameters


Maycon-Santos-sm:
*cfn-lint version: 0.42.0*

*Description of issue.*

Please provide as much information as possible:

cfn-lint is showing a false alarm for the check E8003 when you use conditional with Equals and Ref function and the **previous parameter** listed in the template is a List type.

I've checked and it seems to be related to the file https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/rules/conditions/Equals.py#L39 on line 39 which uses `valid_refs[element_value].get('Type')` method to get the parameter type, but for some reason, this is returning the type of the previous parameter. When switching to `valid_refs.get(element_value).get('Type')`  the validation works as expected.

You can use the following template to test:

```yaml
---

AWSTemplateFormatVersion: "2010-09-09"
Parameters:

  CreatePolicy:
    Type: String
    Default: 'true'
    Description: "Create IAM policy"
    AllowedValues:
    - 'false'
    - 'true'

  IAMArn:
    Type: "CommaDelimitedList"
    Description: The Cross account IAM having access on the repository
    Default: "None"

  IAMArnAllowWrite:
    Type: String
    Default: 'false'
    Description: "Allow IAM writer policy"
    AllowedValues:
    - 'false'
    - 'true'

Conditions: 
  ShouldCreatePolicy: !Equals [!Ref CreatePolicy, 'true']
  ShouldAllowWritePolicy: !Equals [!Ref IAMArnAllowWrite, 'true']

Resources:

  ECRRepositry:
    Type: AWS::ECR::Repository
    Properties:
      RepositoryPolicyText:
        Fn::If:
        - ShouldCreatePolicy
        - 
          Version: "2012-10-17"
          Statement:
            -
              Sid: AllowPushPull
              Effect: Allow
              Principal:
                AWS: !Ref IAMArn
              Action:
                Fn::If:
                - ShouldAllowWritePolicy
                - - "ecr:GetDownloadUrlForLayer"
                  - "ecr:BatchGetImage"
                  - "ecr:PutImage"
                  - "ecr:BatchCheckLayerAvailability"
                  - "ecr:InitiateLayerUpload"
                  - "ecr:UploadLayerPart"
                  - "ecr:CompleteLayerUpload"
                - - "ecr:GetDownloadUrlForLayer"
                  - "ecr:BatchGetImage"
[...]

PR #1814, opened 2020-12-01
Avoid getting type from previous cf parameter

Maycon-Santos-sm:
This fixes the issue reported in the *Issue #1813*

*Description of changes:*

The validation of the parameter type is getting the value from the previous parameter when using dict[key] method to retrieve value of in the dictionary.

### DIFF
--- a/src/cfnlint/rules/conditions/Equals.py
+++ b/src/cfnlint/rules/conditions/Equals.py
@@ -35,8 +35,8 @@ class Equals(CloudFormationLintRule):
                 elif element_key == 'Ref':
                     valid_refs = cfn.get_valid_refs()
                     if element_value in valid_refs:
-                        if valid_refs[element_value].get('From') == 'Parameters':
-                            if valid_refs[element_value].get('Type') in VALID_PARAMETER_TYPES_LIST:
+                        if valid_refs.get(element_value).get('From') == 'Parameters':
+                            if valid_refs.get(element_value).get('Type') in VALID_PARAMETER_TYPES_LIST:
                                 message = 'Every Fn::Equals object requires a list of 2 string parameters'
                                 matches.append(RuleMatch(
                                     path, message))


src/cfnlint/rules/conditions/Equals.py
```diff
@@ -35,8 +35,8 @@ def _check_equal_values(self, cfn, element, path):
                 elif element_key == 'Ref':
                     valid_refs = cfn.get_valid_refs()
                     if element_value in valid_refs:
-                        if valid_refs[element_value].get('From') == 'Parameters':
-                            if valid_refs[element_value].get('Type') in VALID_PARAMETER_TYPES_LIST:
+                        if valid_refs.get(element_value).get('From') == 'Parameters':
```
kddejong, line 38: Good catch.

kddejong (approved):

src/cfnlint/rules/conditions/Equals.py
benbridts: Are we sure this works as expected?

If the key exist, they should behave the same. If it doesn't you get an exception for doing None.get()

Or am I misreading this?

kddejong: yea agreed.  I like `get` but we should probably have a default value to avoid any possible failures.  I actually think the fix you put in solved this issue.  Testing some more

Maycon-Santos-sm: It seems like that, by using .get(key) instead of dic[key] I was avoiding the __getitem__ method in the RegexDict class, that's why it "worked" but probably not the way you would expect.

benbridts: Should we implement that method and either:
- throw a NotImplementedError
- actually implement it (catch KeyError, return default)?

kddejong: added this to #1816 
```
def get(self, key, default=None):
        try:
            return self[key]
        except KeyError:
            return default
```

Issue #1822, opened 2020-12-04
Newline in rule output for W2031 breaks parseable format


direvus:
https://github.com/aws-cloudformation/cfn-python-lint/blob/e6b5e4f8863ba64cf6fbe2a3e36c113e65c200a6/src/cfnlint/rules/parameters/AllowedPattern.py#L64 and again on L72 embeds a newline character in the message from rule W2031.

This breaks the output in `-f parseable` mode, which is line based.

I think we should make 2 changes:

- remove newlines from all messages produced by rules, and
- modify the behaviour of `-f parseable` to replace newlines in rule messages with spaces.

PR to follow.

PatMyron:
[recommend against the `parseable` format btw](https://github.com/aws-cloudformation/cfn-python-lint/issues/1467#issuecomment-687864814)

direvus:
> [recommend against the `parseable` format btw](https://github.com/aws-cloudformation/cfn-python-lint/issues/1467#issuecomment-687864814)

Yeah, the issue I've raised is certainly not the only problem with "parseable".  It's pretty depressing that it doesn't bother to escape or quote colon characters within the fields.  But, unless you want to formally deprecate it, "parseable" is still one of the supported formats, so it should work, to some extent.  I feel that my PR is an easy win that moves it a little closer towards useful.

PR #1823, opened 2020-12-04
Bugfix/1822

direvus:
*Issue #, if available:* #1822 

*Description of changes:* Remove newlines in rule messages to avoid breaking --parseable output.  Also replace any newlines that do make it through to ParseableFormatter with space.


kddejong (changes requested):
Looks good just need to get the tests cleared up.  Thats something I can help with if you would like.

direvus:
> Looks good just need to get the tests cleared up. Thats something I can help with if you would like.

Yes please.  I was a bit surprised that the CI build failed, I did go to have a look at the results but I didn't see anything that seemed even vaguely related to my changes.

kddejong:
Replaced with #1854

PR #1842, opened 2020-12-22
Update AWS::Elasticsearch::Domain DomainEndpointOptions

lkinley-rythmos:
* Add CustomEndpoint, CustomEndpointCertificateArn,
CustomEndpointEnabled properties

*Issue #, if available:*

*Description of changes:*
Add Custom Endpoint properties that were added to CFN on Nov 5.  This change appears to have slipped in with only a general announcement here: https://aws.amazon.com/about-aws/whats-new/2020/11/amazon-elasticsearch-service-now-supports-defining-a-custom-name-for-your-domain-endpoint/
There is no announcement on the CFN release history page here https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/ReleaseHistory.html
However, they are documented in the CFN docs here: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticsearch-domain.html


PatMyron:
These should be autogenerated by https://github.com/aws-cloudformation/cfn-python-lint/pull/1831
We just need to fix some of the other unrelated issues before being able to merge that in

PR #1861, opened 2021-01-08
spec patch cleanup

kddejong:
*Issue #, if available:*

*Description of changes:*
- Remove any manual patches being applied that no longer need to exist.


PatMyron:
Should cleanup the duplicate mentioned in https://github.com/aws-cloudformation/cfn-python-lint/issues/1865 here as well

PR #1732, opened 2020-10-08
CloudFormation Registry resource schema support (--registry-schemas)

PatMyron:
https://github.com/aws-cloudformation/cfn-python-lint/issues/1277

[CloudFormation Registry announcement](https://aws.amazon.com/blogs/aws/cloudformation-update-cli-third-party-resource-support-registry/)
[Resource Schema repo](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/)

---

- [x] jsonschema.validate -> [`RuleMatch`](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/docs/getting_started/rules.md#rule-matches) format (**only returns one error per resource + exact context isn't great**)
  - [ ] investigate [iterating all ValidationErrors](https://python-jsonschema.readthedocs.io/en/latest/validate/#jsonschema.IValidator.iter_errors) per resource
- [x] generalize beyond 1 resource type
- [ ] automate updating [PUBLIC resource schemas](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resource-type-schemas.html)
  - [ ] with caching
- [x] handle CloudFormation template syntax
  - [x] https://github.com/aws-cloudformation/cfn-python-lint/blob/338d5fb29a2f47069f0b1091fc420f6dd9e6013d/src/cfnlint/helpers.py#L77
  - [x] https://github.com/aws-cloudformation/cfn-python-lint/blob/338d5fb29a2f47069f0b1091fc420f6dd9e6013d/src/cfnlint/helpers.py#L126
  - [x] https://github.com/aws-cloudformation/cfn-python-lint/blob/338d5fb29a2f47069f0b1091fc420f6dd9e6013d/src/cfnlint/helpers.py#L66
  - [x] https://github.com/aws-cloudformation/cfn-python-lint/blob/1a6657bafb5330ff8e68ae99538ff99f42bcd0bc/src/cfnlint/helpers.py#L65
  - [ ] [JSON schema type checking stricter than CloudFormation's](https://github.com/aws-cloudformation/cfn-python-lint/issues/547)
- [x] allow non-PUBLIC resource schemas to be passed in
  - [x] Update [`E3001`](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/docs/rules.md#E3001) to include these resource types:https://github.com/aws-cloudformation/cfn-python-lint/blob/93659fba5cf86ee34177c55b28df0aa9140b7dc9/src/cfnlint/rules/resources/Configuration.py#L105
[...]

PR #1682, opened 2020-09-05
AllowedValues from botocore during --update-specs

PatMyron:
most commonly maintained [concise constraint format](https://github.com/aws-cloudformation/cfn-python-lint/issues/903#issuecomment-595356853) last time I checked:
```
src/cfnlint/data/ExtendedSpecs/all $ git grep 'AllowedValues":' | wc -l
AllowedValues: 187 # only 42 hardcoded AllowedValues remaining after this PR
NumberMin/NumberMax: 46
AllowedPatternRegex: 12
StringMin/StringMax: 8
ListMin/ListMax: 4
JsonMax: 2
```

https://github.com/aws-cloudformation/cfn-python-lint/issues/50#issuecomment-668265454

---

[`CloudSpecs`](https://github.com/aws-cloudformation/cfn-python-lint/tree/master/src/cfnlint/data/CloudSpecs) is autogenerated by [`cfn-lint --update-specs`](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/maintenance.py#L21)

---

`AllowedValues` were hardcoded in [`ExtendedSpecs/all`](https://github.com/aws-cloudformation/cfn-python-lint/tree/master/src/cfnlint/data/ExtendedSpecs/all) and patched into all the [Resource Specifications](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html) in [`CloudSpecs`](https://github.com/aws-cloudformation/cfn-python-lint/tree/master/src/cfnlint/data/CloudSpecs) when `cfn-lint --update-specs` was run. Now they're references to botocore types in [`ExtendedSpecs/all`](https://github.com/aws-cloudformation/cfn-python-lint/tree/master/src/cfnlint/data/ExtendedSpecs/all) and `cfn-lint --update-specs` pulls from botocore directly to write that part of [`CloudSpecs`](https://github.com/aws-cloudformation/cfn-python-lint/tree/master/src/cfnlint/data/CloudSpecs)

---

[`AWS::WorkSpaces::Workspace.ComputeTypeName`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-workspaces-workspace-workspaceproperties.html#cfn-workspaces-workspace-workspaceproperties-computetypename
[...]

Issue #1277, opened 2019-12-30
Support registry resource schemas


softprops:
*cfn-lint version: (`cfn-lint --version`)*

any/all

*Description of issue.*

My company uses this tool to ensure correctness of our CloudFormation templates and would like to explore CloudFormation's new registry for third party resources.  In particular, we'd are trialing [those provided by datadog](https://github.com/DataDog/datadog-cloudformation-resources/). 

Understandably cf-lint is not currently aware of the schema's these declare for custom resources but I was wondering if adding support for them would possible given the current schema definition formation cf-lint relies on? The problem for us is that cf-lint will always fail for templates containing these resources. The current work around is to put them into separate files and use a naming convention that cf-lint avoids but I don't think that's an ideal approach for the way CloudFormation registry resources are intended to be used.

Strawman idea. Add a new multi value parameter the s3 schema uris these resources are required to publish to cf-lint and have it incorporate those in its validation. This allows for organizations to change resource schema versions of these resources and to update s3 uris provided to cfn lint in sync

kddejong:
@softprops how would you feel about a cfn-lint command that would update a patch file from that registry?   I'm not sure I would like to do it with `--update-specs` at this point because we have tried to make that command to not require authentication to AWS.  @PatMyron any thoughts here?

softprops:
I don't know enough implementation details to know what that does. Rather than focus on implementation I'll try to make clearer the goal. I would like to have a way to continue to use cfn-lint in ci and not fail when it encounters a cfn template for a custom resource. 

A common property of all third party resources is that they all require hosting of a schema file describing that resource. What would be great if there was a way for cfn to leverage that type to schema mapping to adapt aws third party resource schema model to what ever schema model cfn lint uses internally.

How does the patch file system work?

PatMyron:
There are now publicly accessible [CloudFormation Registry](https://aws.amazon.com/blogs/aws/cloudformation-update-cli-third-party-resource-support-registry/) [resource provider schemas](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/) for all [of the PUBLIC resource types in Amazon's namespaces](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resource-type-schemas.html)

The Linter has a lot of Amazon resource type information that is not available in these schemas that we don't want to lose:

[...]

Issue #51, opened 2018-05-08
Better error messaging for invalid type or property errors


kddejong:
Currently you can have an invalid type or invalid property based on it actually being invalid, not supported in the region being tested, or by the override-spec parameter.   

Ideally the error message would more appropriately relate to the error.  Example: This resource type is invalid based on the override spec or this resource type isn't supported in this region. 

The goal would be to make sure its understood what the CloudFormation Spec failed and what was a failure of that particular run (with parameter or regions) so we don't confuse the users.

fatbasstard:
I've also been thinking about this also when building the whitelist/blacklist feature because the handling of errors is/was exactly the same for incorrect resource types and resources types not suppoerted in a specific region....

For the override-spec it could be farely easy by adding a property to the spec file, like (just an idea):

```
  myS3Bucket:
    Supported: false 
    Type: AWS::S3::Bucket
    Properties:
      BucketName: "my-bucket-name"
```
Which you could implement in the actual test and raise a corresponding error message.
But that means actually "extending" the Spec files with new features which feels not really correct.
And besides that, the "unsupported in region" is then also not handled.

There's no "complete" list of AWS/CloudFormation Resources. If that would be there then there would be workarounds...

If an organisation/user implements an override-spec, you know that there's an exclusion mechanism, so the message being "unclear" is an assumption I believe...  (Based on experience with letting dozens of developers write CloudFormation which is linted with this feature for months)

PatMyron:
As far as a "complete" list of CloudFormation Resources, could either the `us-east-1` specification or a superset of all of the regions' specifications be used?

The first option would be easier, and I think it could be accurate enough. If a resource or property is invalid in `ap-northeast-3`, it could be informative to know if it is valid in `us-east-1` because that is likely the difference between being truly invalid versus being simply unsupported in a region. 

<img width="934" alt="services-per-region" src="https://user-images.githubusercontent.com/7014355/62002835-2f48fd00-b0c1-11e9-9d2e-f5a436d72f8c.png">
<img width="937" alt="resources-per-region" src="https://user-images.githubusercontent.com/7014355/62002836-31ab5700-b0c1-11e9-8530-6d3f3865f92f.png">

PR #1867, opened 2021-01-12
Get value constraints from AWS registry types

kddejong:
*Issue #, if available:* https://github.com/aws-cloudformation/cfn-python-lint/issues/1277

*Description of changes:*
- Use the Registry to get AWS property constraints
(in addition to already existing AWS constraints [manually maintained](https://github.com/aws-cloudformation/cfn-python-lint/issues/903#issuecomment-595356853) and [pulled from `boto`](https://github.com/aws-cloudformation/cfn-python-lint/pull/1682/) and [3rd party constraints from Registry](https://github.com/aws-cloudformation/cfn-python-lint/pull/1732))


scripts/update_from_registry.py
```diff
@@ -0,0 +1,180 @@
...
+    schema = json.loads(details.get('Schema'))
+    results = get_object_details(name, schema.get('properties'), schema)
+
+    return results
+
+def main():
+    """ main function """
+    configure_logging()
+
+    types = client.list_types(
```
PatMyron, line 121: comment to remember pagination (+1 to https://github.com/boto/botocore/pull/2177)

PatMyron: These hosted [resource provider schemas](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resource-type-schemas.html) might be even better to avoid credential requirements and pagination

scripts/update_from_registry.py
```diff
@@ -0,0 +1,180 @@
+#!/usr/bin/env python
```
PatMyron, line 1: let's make sure this is a part of https://github.com/aws-cloudformation/cfn-python-lint/pull/1792 as well

PatMyron: thoughts on just putting this in `cfn-lint --update-specs` like https://github.com/aws-cloudformation/cfn-python-lint/pull/1682 if we use the hosted [resource provider schemas](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resource-type-schemas.html) to avoid new credential requirements?

src/cfnlint/data/ExtendedSpecs/all/08_registry_value_types.json
```diff
@@ -0,0 +1,1883 @@
+[
```
PatMyron, line 1: thoughts on ditching these files + writing directly to [`CloudSpecs`](https://github.com/aws-cloudformation/cfn-python-lint/blame/master/src/cfnlint/data/CloudSpecs) if we make this part of `cfn-lint --update-specs`?

think it's confusing that [`ExtendedSpecs/all`](https://github.com/aws-cloudformation/cfn-python-lint/tree/master/src/cfnlint/data/ExtendedSpecs/all) is partially manually maintained and partially autogenerated

might be useful to include some metadata that the information is coming from the Registry though

PatMyron:
Noting this appears to overwrite any already existing constraints since it's applied last (seems okay, just noteworthy)

PatMyron:
Do you know how this handles [combined JSON schemas](https://json-schema.org/understanding-json-schema/reference/combining.html)? Fine to not handle them yet since [there are so few resource provider schemas using them](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resource-type-schemas.html), just want to make sure they're at least not handled incorrectly:

```bash
grep -l -E -i '(oneOf|anyOf|allOf)' *
aws-applicationinsights-application.json
aws-cloudformation-moduledefaultversion.json
aws-cloudformation-stackset.json
aws-cloudwatch-metricstream.json
aws-databrew-dataset.json
aws-databrew-recipe.json
aws-gamelift-alias.json
aws-iotwireless-wirelessdevice.json
aws-kendra-datasource.json
aws-rds-globalcluster.json
aws-s3-storagelens.json
aws-sagemaker-pipeline.json
aws-synthetics-canary.json
```

scripts/update_from_registry.py
```diff
@@ -0,0 +1,180 @@
...
+            if propdetails.get('properties'):
+                results.update(get_object_details(name + '.' + subname, propdetails.get('properties'), schema))
+            else:
+                return results
+                LOGGER.info("Type %s object for %s has no properties", name, propname)
+        elif t not in ['string', 'integer', 'number', 'boolean']:
+            if t.get('$ref'):
+                results.update(get_object_details(name + '.' + propname, schema.get('definitions').get(t.get('$ref').split("/")[-1]), schema))
+            else:
+                print('fix later')
```
PatMyron, line 78: Can you expand on what's happening here?

scripts/update_from_registry.py
kddejong: yea good question.  Any thoughts on how this would scale and the speed of it?  We could download the zips and read them instead of using the APIs (honestly didn't know that existed).  Plus it would help with my rate limiting issues.

scripts/update_from_registry.py
```diff
@@ -0,0 +1,210 @@
...
+                results.update(get_object_details(name + '.' + propname, schema.get('definitions').get(t.get('$ref').split("/")[-1]), schema))
+            else:
+                LOGGER.info("Unable to handle %s object for %s property", name, propname)
+        elif t == 'string':
+            if not results.get(name + '.' + propname):
+                if propdetails.get('pattern') or (propdetails.get('minLength') and propdetails.get('maxLength')) or propdetails.get('enum'):
+                    results[name + '.' + propname] = {}
+            if propdetails.get('pattern'):
+                results[name + '.' + propname].update({
+                    'AllowedPattern': propdetails.get('pattern')
```
PatMyron, line 97:
```suggestion
                    'AllowedPatternRegex': propdetails.get('pattern')
```

PatMyron:
This is awesome btw, this alone is pulling in significantly more constraints:
```
AllowedValues: 315
NumberMin/NumberMax: 49
AllowedPatternRegex: 773
StringMin/StringMax: 662
```
[than we had before](https://github.com/aws-cloudformation/cfn-python-lint/pull/1682)

scripts/update_from_registry.py
kddejong: Good catch

src/cfnlint/data/ExtendedSpecs/all/08_registry_value_types.json
```diff
@@ -0,0 +1,12809 @@
...
+      "AllowedPattern": "^.?P.*",
+      "StringMax": 15,
+      "StringMin": 1
+    }
+  },
+  {
+    "op": "add",
+    "path": "/ValueTypes/AWS::SageMaker::ModelBiasJobDefinition.EndpointInput.EndTimeOffset",
+    "value": {
+      "Allowed
```
PatMyron, line 12394: Similar to https://github.com/aws-cloudformation/cloudformation-cli/issues/459, we need to think about how to verify regular expressions are valid in Java and Python
(this one doesn't seem to be valid in either though 🤔)

src/cfnlint/data/ExtendedSpecs/all/08_registry_value_types.json
kddejong: agreed.  The automation portions can get confusing.  Let me see what I can do.

src/cfnlint/data/ExtendedSpecs/all/08_registry_value_types.json
PatMyron: [`ERROR: Failed building wheel for regex`](https://travis-ci.com/github/aws-cloudformation/cfn-python-lint/jobs/471402886)

src/cfnlint/rules/resources/ectwo/Ebs.py
```diff
@@ -2,7 +2,7 @@
 Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 SPDX-License-Identifier: MIT-0
 """
-import re
+import re
```
PatMyron, line 5:
```suggestion
import re
```

src/cfnlint/data/ExtendedSpecs/all/08_registry_value_types.json
kddejong: Current values that would need to get corrected in `AWS::DataBrew::Job.DatasetName`
```
DatasetName" : {
      "description" : "Dataset name",
      "type" : "string",
      "pattern" : "[ -퟿-�𐀀-\t]*",
      "minLength" : 1,
      "maxLength" : 255
    },
```

PatMyron: found all [non-ASCII characters](https://stackoverflow.com/questions/3001177/how-do-i-grep-for-all-non-ascii-characters) in [resource schemas](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resource-type-schemas.html):
```bash
$ brew install coreutils
$ ggrep --color='auto' -P -n "[^\x00-\x7F]" *
aws-appflow-connectorprofile.json:158:          "description" : "A unique alphanumeric identiﬁer used to authenticate a user, developer, or calling program to your API.",
aws-appflow-connectorprofile.json:171:          "description" : "A unique alphanumeric identiﬁer used to authenticate a user, developer, or calling program to your API.",
aws-appflow-connectorprofile.json:175:          "description" : "Application keys, in conjunction with your API key, give you full access to Datadog’s programmatic API. Application keys are associated with the user account that created them. The application key is used to log all requests made to the API.",
aws-appflow-connectorprofile.json:215:          "description" : "The identiﬁer for the desired client.",
aws-appflow-connectorprofile.json:245:          "description" : "The identiﬁer for the user.",
aws-appflow-connectorprofile.json:273:          "description" : "The identiﬁer for the desired client.",
aws-appflow-connectorprofile.json:327:          "description" : "The object key for the destination bucket in which Amazon AppFlow will place the ﬁles.",
aws-appflow-connectorprofile.json:398:          "description" : "A unique alphanumeric identiﬁer used to authenticate a user, developer, or calling program to your API.",
aws-appflow-connectorprofile.json:408:          "description" : "The identiﬁer for the desired client.",
aws-appflow-connectorprofile.json:454:          "description" : "The name of the Snowﬂake warehouse.",
aws-appflow-connectorprofile.json:458:          "description" : "The name of the Amazon S3 stage that was created while setting up an Amazon S3 stage in the\nSnowﬂake account. This is written in the following format: < Database>< Schema><Stage Name>.",
aws-appflow-connectorprofile.json:462:          "description" : "The name of the Amazon S3 bucket associated with Snowﬂake.",
aws-appflow-connectorprofile.json:466:          "description" : "The bucket prefix that refers to the Amazon S3 bucket associated with Snowﬂake.",
aws-appflow-connectorprofile.json:470:          "description" : "The Snowﬂake Private Link service name to be used for private data transfers.",
aws-appflow-connectorprofile.json:478:          "description" : "The region of the Snowﬂake account.",
aws-appflow-connectorprofile.json:522:          "description" : "The identiﬁer for the desired client.",
aws-databrew-job.json:9:      "pattern" : "[ -퟿-�𐀀-􏿿\t]*",
aws-databrew-job.json:27:      "pattern" : "[ -퟿-�𐀀-􏿿\t]*",
aws-databrew-job.json:72:      "pattern" : "[ -퟿-�𐀀-􏿿\t]*",
aws-databrew-recipe.json:8:      "pattern" : "[ -퟿-�𐀀-􏿿\t]*",
aws-databrew-recipe.json:16:      "pattern" : "^[ -.0-^`-퟿-�]*$",
aws-databrew-schedule.json:37:      "pattern" : "[ -퟿-�𐀀-􏿿\t]*",
aws-gamelift-gameservergroup.json:244:      "pattern" : "[ -퟿-�𐀀-􏿿\r\n\t]*",
aws-kinesis-stream.json:26:      "description" : "An arbitrary set of tags (key–value pairs) to associate with the Kinesis stream.",
aws-kinesis-stream.json:73:      "description" : "An arbitrary set of tags (key–value pairs) to associate with the Kinesis stream.",
aws-mediapackage-packagingconfiguration.json:100:          "description" : "This setting controls how ad markers are included in the packaged OriginEndpoint. “NONE” will omit all SCTE-35 ad markers from the output. “PASSTHROUGH” causes the manifest to contain a copy of the SCTE-35 ad markers (comments) taken directly from the input HTTP Live Streaming (HLS) manifest. “SCTE35_ENHANCED” generates ad markers and blackout tags based on SCTE-35 messages in the input source.",
aws-mediapackage-packagingconfiguration.json:141:          "description" : "The Dynamic Adaptive Streaming over HTTP (DASH) profile type. When set to “HBBTV_1_5”, HbbTV 1.5 compliant output is enabled.",
aws-mediapackage-packagingconfiguration.json:217:          "description" : "A list of triggers that controls when the outgoing Dynamic Adaptive Streaming over HTTP (DASH) Media Presentation Description (MPD) will be partitioned into multiple periods. If empty, the content will not be partitioned into more than one period. If the list contains “ADS”, new periods will be created where the Asset contains SCTE-35 ad markers.",
```
https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-appflow/pull/6
https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-databrew/pull/2
https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-gamelift/pull/13
https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-kinesis/pull/6
https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-mediapackage/pull/3
(also added to the TODO list in https://github.com/aws-cloudformation/cloudformation-cli/pull/663)

PatMyron:
Manual testing instructions:
```bash
# checkout and install PR in new virtual environment
git clone https://github.com/aws-cloudformation/cfn-python-lint
cd cfn-python-lint
brew install gh
gh pr checkout 1867
python3 -m venv env
source env/bin/activate
pip3 install -e .

# test on templates
cfn-lint template.yaml

# deactivate virtual environment when done
deactivate
```

jlongtine:
@PatMyron @kddejong When do you think this might make it into `master`?

kddejong:
@jlongtine this week ideally. I'm going to put in another patch release with the most recent version of the specs and then I'm going to move on merging this. There are so many extra requirements we are a little concerned about merging in things that will break people.  We are trying to test as much as we can before merging.

One example we ran into is related to allowed patterns. Since the AWS resources are written in Java it turns out that some of the allowed patterns are written to take advantage of some of the Java allowed patterns (so they won't work in Python). There are also some escaping issues with unicode characters in those patterns and how they are written to the source we access.  We have tried hard to filter those out as much as we could but this part is making me the most nervous.  We will try to get this testing done this week and get this shipped.  Thanks for your patience.

src/cfnlint/maintenance.py
```diff
@@ -18,6 +27,8 @@
 
 LOGGER = logging.getLogger(__name__)
 
+REGISTRY_SCHEMA_ZIP = 'https://schema.cloudformation.us-east-1.amazonaws.com/CloudformationSchema.zip'
```
benbridts, line 30: - Is this documented somewhere (this is much more user-friendly than how I was querying the registry in my account)?
- Do we need to download this for multiple regions (the eu-west-1  version of this URL works)?

PatMyron: * [documented here](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resource-type-schemas.html)
* [`us-east-1` is almost always roughly a superset of all regions](https://github.com/aws-cloudformation/cfn-python-lint/issues/51#issuecomment-515735722), which is why `us-east-1` is used as the default for `cfn-lint` itself. I'm okay with just using `us-east-1` for additional constraints for now. Looks like [mostly just `AWS::GroundStation` constraints might be missed at the moment](https://github.com/ScriptAutomate/aws-cfn-resource-specs/blob/master/CHANGELOG.md#existing-resourcetypes-and-propertytypes-not-in-us-east-1)

kddejong:
Something is a little off here.  We are missing the references to the value types in some scenarios. Doing some digging into this now

kddejong:
This is going to add a little time to build `update-specs` process because the amount of jsonpatches.  I think the better long term solution is to separate out our spec update process into another repo that we can build and release them.

jlongtine:
> This is going to add a little time to build `update-specs` process because the amount of jsonpatches. I think the better long term solution is to separate out our spec update process into another repo that we can build and release them.

I like this idea. Separating this from the python codebase would help me with `cfn-cue`.

src/cfnlint/maintenance.py
```diff
@@ -186,3 +243,171 @@ def update_iam_policies():
...
+                    LOGGER.info('Type for %s object and %s property is a list', names[0], propname)
+                else:
+                    LOGGER.info('Unable to handle %s object for %s property', names[0], propname)
+            elif t == 'string':
+                if not results.get('.'.join(names + [propname])):
+                    if propdetails.get('pattern') or (propdetails.get('minLength') and propdetails.get('maxLength')) or propdetails.get('enum'):
+                        results['.'.join(names + [propname])] = {}
+                if propdetails.get('pattern'):
+                    p = propdetails.get('pattern')
+                    if '.'.join(names + [propname]) == 'AWS::OpsWorksCM::Server.CustomPrivateKey':
```
PatMyron, line 307: [linking for reference](https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-opsworkscm/blob/a9c53dda37360c4b6497fa9283ee9a8922687ef1/aws-opsworkscm-server/aws-opsworkscm-server.json#L82)

src/cfnlint/maintenance.py
```diff
@@ -186,3 +243,171 @@ def update_iam_policies():
...
+                            'Unable to handle regex for type %s and property %s with regex %s', names[0], propname, p)
+                if propdetails.get('minLength') and propdetails.get('maxLength'):
+                    results['.'.join(names + [propname])].update({
+                        'StringMin': propdetails.get('minLength'),
+                        'StringMax': propdetails.get('maxLength'),
+                    })
+                if propdetails.get('enum'):
+                    results['.'.join(names + [propname])].update({
+                        'AllowedValues': propdetails.get('enum')
+                    })
```
PatMyron, line 342: [Could grab `enums` for types other than `string` as well](https://json-schema.org/understanding-json-schema/reference/generic.html#enumerated-values)

---

Especially since we seem to already be handling more types in the rule itself:
https://github.com/aws-cloudformation/cfn-python-lint/blob/853522ff0e4c977331ac677a47059cf96b6f3189/src/cfnlint/rules/resources/properties/AllowedValue.py#L38

src/cfnlint/maintenance.py
```diff
@@ -186,3 +243,171 @@ def update_iam_policies():
...
+            elif t in ['number', 'integer']:
+                if not results.get('.'.join(names + [propname])):
+                    if propdetails.get('minimum') and propdetails.get('maximum'):
+                        results['.'.join(names + [propname])] = {}
+                if propdetails.get('minimum') and propdetails.get('maximum'):
+                    results['.'.join(names + [propname])].update({
+                        'NumberMin': propdetails.get('minimum'),
+                        'NumberMax': propdetails.get('maximum'),
```
PatMyron, line 350: [Could use `exclusiveMinimum`/`exclusiveMaximum` as well](https://json-schema.org/understanding-json-schema/reference/numeric.html#range)

---

[Although no AWS resource schemas use these keywords yet anyway](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resource-type-schemas.html)

PatMyron (commented):
[`minItems`/`maxItems`](https://json-schema.org/understanding-json-schema/reference/array.html#length) could be translated to [`ListMin`/`ListMax`](https://github.com/aws-cloudformation/cfn-python-lint/issues/903#issuecomment-595356853) as well

PR #1913, opened 2021-02-22
AWS::Route53::RecordSet.TTL property to String as per docs

sameer-ranjan:
*Issue #, if available:*

*Description of changes:*

As per the AWS [cloudformation docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-route53-recordset.html#cfn-route53-recordset-ttl) TTL propert in `AWS::Route53::RecordSet` is `String` while in the linter it says `LONG` .

### DIFF
--- a/src/cfnlint/data/CloudSpecs/ap-east-1.json
+++ b/src/cfnlint/data/CloudSpecs/ap-east-1.json
@@ -47647,7 +47647,7 @@
         },
         "TTL": {
           "Documentation": "http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-route53-recordset.html#cfn-route53-recordset-ttl",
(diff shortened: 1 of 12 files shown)


kddejong:
The change here would be to remove this line.  https://github.com/aws-cloudformation/cfn-python-lint/blob/0b9f7251fb40b05e66b02ebeac2bbfd70d25e71b/src/cfnlint/data/ExtendedSpecs/all/01_spec_patch.json#L100-L104

The reason this is changed because the underlying property requires it to be long (https://docs.aws.amazon.com/Route53/latest/APIReference/API_ResourceRecordSet.html#Route53-Type-ResourceRecordSet-TTL).  For now if you want to use a string still I would set the configuration to allow non strict checking of primitive types.

PR #1929, opened 2021-03-10
add tests for formatters that use rules (currently failing)

jonathanmorley:
*Issue #, if available:* https://github.com/aws-cloudformation/cfn-python-lint/issues/1930

*Description of changes:* This adds some tests for formatters that make use of the `rules` argument.

They are currently failing because I believe the current code to be incorrect. It only uses the rules from the last template that is checked.


kddejong:
replacing with #2330 and includes fixes

PR #1952, opened 2021-03-23
Fix some changes to the API gateway iam policies

kddejong:
*Issue #, if available:*
fix #1951 
*Description of changes:*
- Update IAM patches to include the new API Gateway references

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #1954, opened 2021-03-24
Revert/1952

kddejong:
*Issue #, if available:*
#1952 
*Description of changes:*
- Make update iam policies not fail if a key isn't found and future proof possible changes in API Gateway permissions

### DIFF
--- a/src/cfnlint/data/AdditionalSpecs/Policies.json
+++ b/src/cfnlint/data/AdditionalSpecs/Policies.json
@@ -3376,6 +3376,27 @@
         "iotevents:keyValue"
       ]
     },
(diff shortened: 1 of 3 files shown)


burningalchemist:
@kddejong will you release the fix to PyPI anytime soon? :) The reason is there are several GitHub-actions (e.g. https://github.com/ScottBrenner/cfn-lint-action) which pull `cfn-python-lint` from the PyPI, and they seem to be broken at the moment.

P.S. the `policies.js` file has been changed again to the latest revision, but I guess this is covered now, so just for your information.

kddejong:
Yea let me get the release out today.

PR #1978, opened 2021-04-12
Fix rules for Iops property and EBS volume types.

### LINKED ISSUE #1977: E2504 incorrectly rejects "Iops" property for io2/gp3 volumes

cfn-lint 0.44.6

*Description of issue.*

cfn-lint produces an error "E2504: Iops shouldn't be defined for type io2 for Resource ... LaunchConfiguration/Properties/BlockDeviceMappings/0/Ebs/Iops" when setting Iops on a io2 EBS volume.

The Iops property is required for io2 and optional for gp3. [1]

Cfn-lint treats the Iops property as required for io1 and forbidden for all other volume types, which is very much not correct 

[1] https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-blockdev-template.html#cfn-ec2-blockdev-template-iops

direvus:
*Issue #, if available:*

Fixes #1977.

*Description of changes:*

The 'Iops' property on an EBS volume is required for io1 and io2
volumes, optional for gp3 volumes, and not supported for gp2, st1, sc1
and standard volumes.

The previous code treated the property as required for io1 and not
supported for all other volume types.  This gave the wrong outcome for
io2 and gp3 specifically, and is generally the wrong move for any new
volume types that are added in the future.

### DIFF
--- a/src/cfnlint/rules/resources/ectwo/Ebs.py
+++ b/src/cfnlint/rules/resources/ectwo/Ebs.py
@@ -9,10 +9,10 @@ from cfnlint.rules import RuleMatch
 
 
 class Ebs(CloudFormationLintRule):
-    """Check if Ec2 Ebs Resource Properties"""
+    """Check Ec2 Ebs Resource Properties"""
     id = 'E2504'
     shortdesc = 'Check Ec2 Ebs Properties'
-    description = 'See if Ec2 Eb2 Properties are valid'
+    description = 'See if Ec2 Ebs Properties are valid'
     source_url = 'https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-blockdev-template.html'
     tags = ['properties', 'ec2', 'ebs']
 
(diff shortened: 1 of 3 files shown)


src/cfnlint/rules/resources/ectwo/Ebs.py
```diff
@@ -26,13 +26,15 @@ def _checkEbs(self, cfn, ebs, path):
                 for volume_type_obj in volume_types_obj:
                     volume_type = volume_type_obj.get('Value')
                     if isinstance(volume_type, six.string_types):
-                        if volume_type == 'io1':
+                        if volume_type in ('io1', 'io2'):
```
PatMyron, line 29: I wonder if this is a safe assumption going forward:
```suggestion
                        if volume_type.startswith('io'):
```

direvus: I did consider that approach, but steered away from it.  When io3 comes out, will the Iops argument be required for it?  Maybe.  We don't know the answer.  What if AWS creates a different volume type that just happens to start with the letters "io"?

I don't feel good about letting the linter raise an error based on a "maybe".  In my view, we should only be telling the user their code is wrong, if we *know*, for sure, it is in fact wrong.

PatMyron: agreed

kddejong (approved):

PR #1993, opened 2021-04-26
Add back glue resources into us-gov-west-1

kddejong:
*Issue #, if available:*

*Description of changes:*
- Add other glue resources into us-gov-west-1


PatMyron (changes requested):
Prefer getting these published upstream

kddejong:
We used to have these in the specs because of the SSM patching scripts but we lost them because some Glue resources started appearing in us-gov-west-1.  They are still valid resources and should be in this spec.  While I agree they should be patched upstream our automated patching has been fixing this for a long time and this manual change will bring it back until its fixed in upstream

PR #2017, opened 2021-05-21
Ignore Mapping key issues when using Fn::Transform

### LINKED ISSUE #2016: E7003 Errors when using Fn::Transform inside a Mapping
*cfn-lint version: 0.49.2*

*Description of issue.*
#2006 tightened what is considered valid for use in a Mapping.  This causes it to reject what otherwise appears to be a valid use of `Fn::Transform` as the body of a Mapping.

For example, this snippet is valid CFN:

```yaml
Mappings:
  AwsAgentPlatformMap:
    Fn::Transform:
      Name: AWS::Include
      Parameters:
        Location: s3://my-bucket-name/version/3.0.1/amazonlinux2/a-json-file.json
```

This usage trips the newly enhanced regex:

```
E7003 Mapping key (Fn::Transform) has invalid name. Name has to be alphanumeric, '-' or '.'
```

kddejong:
*Issue #, if available:*
fix #2016
*Description of changes:*
- Ignore Fn::Transform when evaluating the mapping key in a mapping rule `E7003`

### DIFF
--- a/src/cfnlint/rules/mappings/KeyName.py
+++ b/src/cfnlint/rules/mappings/KeyName.py
@@ -35,7 +35,7 @@ class KeyName(CloudFormationLintRule):
         if not isinstance(key, six.string_types):
             message = 'Mapping key ({0}) has to be a string.'
             matches.append(RuleMatch(path[:], message.format(key)))
-        elif not re.match('^[a-zA-Z0-9.-]{1,255}$', key):
+        elif not re.match('^[a-zA-Z0-9.-]{1,255}$', key) and key != 'Fn::Transform':
             message = 'Mapping key ({0}) has invalid name. Name has to be alphanumeric, \'-\' or \'.\''
             matches.append(RuleMatch(path[:], message.format(key)))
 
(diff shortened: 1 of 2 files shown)


src/cfnlint/rules/mappings/KeyName.py
```diff
@@ -35,7 +35,7 @@ def check_key(self, key, path):
         if not isinstance(key, six.string_types):
             message = 'Mapping key ({0}) has to be a string.'
             matches.append(RuleMatch(path[:], message.format(key)))
-        elif not re.match('^[a-zA-Z0-9.-]{1,255}$', key):
+        elif not re.match('^[a-zA-Z0-9.-]{1,255}$', key) and key != 'Fn::Transform':
```
PatMyron, line 38: Cool with this, but had a wilder general thought on https://github.com/aws-cloudformation/cfn-lint/issues/476

What if every violation on [`Transform`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-transform.html)  lines were ignored, regardless of individual rule logic?

kddejong: I was more filtering this out because we didn't flag this as a failure, then we started flagging it as a failure so wanted to bring it back to where it was.

Interesting.  It could be possible. We would want to evaluate the start and end mark of the finding is within the transform so if someone is doing a one line json document we don't filter out everything.  

As we continue to evaluate the need of credentials we may be able to test resolving the Transform as well.  Your option would be a great way to not have an issue if we can't resolve credentials.  We may be able to implement your idea fairly easily.  I may test it out today actually :)

PR #2031, opened 2021-06-02
Update rule E1029 to only alert when its a valid ref or getatt

### LINKED ISSUE #2026: E1029 - false positive with X.509 Certificate AWS IoT Core policy variables
cfn-lint version: (cfn-lint 0.49.2)

I got a false positive on X.509 Certificate AWS IoT Core policy variables.

I think that `cfn-lint` may not support X.509 Certificate AWS IoT Core policy variables.

https://docs.aws.amazon.com/iot/latest/developerguide/cert-policy-variables.html

As noted above, X.509 certificate attributes can contain one or more values, but `resource_excludes` does not support specification by regex. This problem is a similar case in the past, `${iot:Connection.Thing.Attributes[attributeName]}` could not be supported (see also #765).

However, I think that just supporting the first value covers many cases.

If you are okay with this idea, I can contribute.

Do you have any comments?

**Output:**
```shell-session
❯ cfn-lint iot-policy.cfn.yml
E1019 Parameter iot:Certificate.Subject.CommonName for Fn::Sub not found at Resources/IoTPolicy/Properties/PolicyDocument/Statement/1/Resource/0/Fn::Sub
iot-policy.cfn.yml:18:13
```

**For reproduction:**
```yaml

Resources:
  IoTPolicy:
    Type: AWS::IoT::Policy
    Properties:
      PolicyDocument:
        Version: 2012-10-17
        Statement:
          - Effect: Allow
            Action:
              - iot:Connect
            Resource:
              - arn:aws:iot:us-east-1:123456789012:client/${iot:Connection.Thing.ThingName}
          - Effect: Allow
            Action:
              - iot:Publish
            Resource:
              - !Sub "arn:aws:iot:us-east-1:123456789012:topic/my/topic/${iot:Certificate.Subject.CommonName}"
```

kddejong:
*Issue #, if available:*
fix #2026 
*Description of changes:*
-Update rule E1029 to only alert when its a valid ref or getatt


test/unit/rules/functions/test_sub_needed.py
```diff
@@ -27,13 +27,13 @@ def test_template_config(self):
         """Test custom excludes configuration"""
         self.helper_file_rule_config(
             'test/fixtures/templates/good/functions/sub_needed_custom_excludes.yaml',
-            {'custom_excludes': '^\\$\\{self.+\\}$'}, 0
+            {'custom_excludes': '^\\$\\{Stage\\}$'}, 0
```
PatMyron, line 30: wonder if `custom_excludes` are even needed anymore

although we don't want to break the interface yet and should allow them to be passed in for backwards compatibility

PR #2034, opened 2021-06-03
Add validate-registry-types and update-registry-type-specs parameters

nhienton:
Added the 2 parameters --validate-registry-types and --update-registry-type-specs to the list of optional arguments for the client. Validation added to only accept the module type. 

### DIFF
--- a/src/cfnlint/core.py
+++ b/src/cfnlint/core.py
@@ -16,11 +16,12 @@ import cfnlint.config
 import cfnlint.formatters
 import cfnlint.decode
 import cfnlint.maintenance
-from cfnlint.helpers import REGIONS, REGISTRY_SCHEMAS
+from cfnlint.helpers import REGIONS, REGISTRY_SCHEMAS, REGISTRY_TYPES
 
 LOGGER = logging.getLogger('cfnlint')
 DEFAULT_RULESDIR = os.path.join(os.path.dirname(__file__), 'rules')
 __CACHED_RULES = None
+
 
 class CfnLintExitException(Exception):
     """Generic exception used when the cli should exit"""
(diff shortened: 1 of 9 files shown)


PatMyron:
might make sense to add `validate-registry-types` to [`data/CfnLintCli/config/schema.json`](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/data/CfnLintCli/config/schema.json) if [`.cfnlintrc`](https://github.com/aws-cloudformation/cfn-lint#config-file) support is desired as well

nhienton:
> might make sense to add `validate-registry-types` to [`data/CfnLintCli/config/schema.json`](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/data/CfnLintCli/config/schema.json) if [`.cfnlintrc`](https://github.com/aws-cloudformation/cfn-lint#config-file) support is desired as well

Good catch, will do. Thanks!

src/cfnlint/core.py
```diff
@@ -156,19 +164,19 @@ def get_args_filenames(cli_args):
         sys.exit(0)
 
     if not sys.stdin.isatty() and not config.templates:
-        return(config, [None], formatter)
+        return (config, [None], formatter)
```
lejiati, line 167: Not sure if this change is made by formatter (along with others below). Since there are no logical changes, let's keep style consistent across different files.

nhienton: Done!

lejiati (approved):
Ship it!

src/cfnlint/core.py
```diff
@@ -58,7 +63,7 @@ def run_cli(filename, template, rules, regions, override_spec, build_graph, regi
                     with open(os.path.join(path, f)) as schema:
                         REGISTRY_SCHEMAS.append(json.load(schema))
 
-    return run_checks(filename, template, rules, regions, mandatory_rules)
+    return run_checks(filename, template, rules, regions, validate_registry_types, mandatory_rules)
```
kddejong, line 66: We should add this to the and set a default value for backwards compatibility with people doing integration.
https://github.com/aws-cloudformation/cfn-lint/blob/main/docs/getting_started/integration.md.

Left as is we may break the integrations that some customers are doing.

This is why you will some functions have things like `mandatory_rules=None`.  Its all for backwards compatibility.

kddejong (changes requested):

src/cfnlint/core.py
```diff
@@ -215,15 +222,26 @@ def get_template_rules(filename, args):
     return(template, __CACHED_RULES, [])
 
 
-def run_checks(filename, template, rules, regions, mandatory_rules=None):
+def run_checks(filename, template, rules, regions, validate_registry_types=None, mandatory_rules=None):
```
kddejong, line 225: I think this should be last.  Just in case people are using positional arguments

kddejong (approved):

PR #2035, opened 2021-06-04
Fix networkx import

b0urb0n:
Package nesting is fixed in latest `networkx` release.  `setup.py` installs `networkx~=2.4` which grabs `2.6rc1` as of a few days ago.  In 2.6, the nested import issue is fixed.  Attempting to import the nested package now causes an `ImportError`.

*Issue #, if available:*

*Description of changes:*

Fix bad import of a nested package which was resolved by https://github.com/networkx/networkx/pull/4839

### DIFF
--- a/src/cfnlint/graph.py
+++ b/src/cfnlint/graph.py
@@ -7,7 +7,7 @@ SPDX-License-Identifier: MIT-0
 import logging
 import re
 import six
-from networkx import networkx
+import networkx
 
 LOGGER = logging.getLogger('cfnlint.graph')
 


kddejong:
Should we also make sure we are grabbing at least 2.6 of `networkx` in `setup.py`

jarrodmillman:
The fact that
```
from networkx import networkx
```
worked was a mistake on our part (i.e., the NetworkX developers).  We didn't think anyone would have run into this bug, but the correct way to import networkx has always been
```
import networkx
```
or
```
import networkx as nx
```

b0urb0n:
> Should we also make sure we are grabbing at least 2.6 of `networkx` in `setup.py`

There's no real need to pin the latest.  2.6 is still an RC at the moment.  You could pin 2.5 but `~=2.4` works just fine.

kddejong:
thanks for the feedback

kddejong (approved):

lorengordon:
Not too big a deal, but just fyi that the networkx change does make cfn-lint 0.50.0 incompatible with networkx 2.6. We were pinning the cfn-lint version in CI (so tests don't inadvertently introduce new checks), but not pinning the networkx version, and that just broke with the networkx 2.6 release. Obviously updating cfn-lint to 0.52.0 fixes things for us.

PR #2055, opened 2021-06-17
Update schema.json for custom_rules

hakanson:
Update `schema.json` to change custom_rules to type string because of either execution or validation failures when attempting to use config file.

After creating a `cfn_custom_rules.txt` file, executing `cfn-lint` from the CLI worked as expected:
```console
$ cfn-lint template.yaml --custom-rules cfn_custom_rules.txt
```

However, after attempting to update `.cfnlintrc.yaml` to this
```yaml
custom_rules: 
  - cfn_custom_rules.txt
```

This error would occur since `custom_rules_file` from [cfnlint/rules/__init__.py", line 429](https://github.com/aws-cloudformation/cfn-lint/blob/main/src/cfnlint/rules/__init__.py#L429) was looking for a string value (and not the array/list of strings from the YAML)
```console
$ cfn-lint template.yaml
  File "/Users/kevhak/.pyenv/versions/3.8.9/bin/cfn-lint", line 8, in <module>
    sys.exit(main())
  File "/Users/kevhak/.pyenv/versions/3.8.9/lib/python3.8/site-packages/cfnlint/__main__.py", line 28, in main
    (template, rules, errors) = cfnlint.core.get_template_rules(filename, args)
  File "/Users/kevhak/.pyenv/versions/3.8.9/lib/python3.8/site-packages/cfnlint/core.py", line 205, in get_template_rules
    __CACHED_RULES = cfnlint.core.get_rules(
  File "/Users/kevhak/.pyenv/versions/3.8.9/lib/python3.8/site-packages/cfnlint/core.py", line 110, in get_rules
    rules.create_from_custom_rules_file(custom_rules)
  File "/Users/kevhak/.pyenv/versions/3.8.9/lib/python3.8/site-packages/cfnlint/rules/__init__.py", line 429, in create_from_custom_rules_file
    with open(custom_rules_file) as customRules:
TypeError: expected str, bytes or os.PathLike object, not list_node
```

An edit of `.cfnlintrc.yaml`  to make a YAML string
```yaml
custom_rules: cfn_custom_rules.txt
```

Resulted in this error because of the current schema requiring an array of strings
```console
$ cfn-lint template.yaml
2021-06-17 17:03:36,559 - cfnlint - ERROR - Error parsing config file: 'cfn_custom_rules.txt' is not of type 'array'
```

*Description of changes:*

Changed

```json
    "custom_rules": {
      "description": "custom rule file to use",
      "items": {
        "type": "string"
      },
      "type": "array"
    },
```

to
```json
    "custom_rules": {
      "description": "custom rule file to use",
      "type": "string"
    },
```


yokogawa-k:
With this change, in 0.52.0 and later versions, do I need to set it as a **String** instead of an **Array**, unlike the example provided in the [README](https://github.com/aws-cloudformation/cfn-lint#config-file)?

kddejong:
Yes that is correct.  We need to get the README updated.  Thanks for pointing that out.  It was a miss that the CLI parameter was looking for a singular value but the .cfnlintrc was looking for a list and the two didn't work well together.

PR #2056, opened 2021-06-18
Feature/manage schemas

nhienton:
Methods added to extract the list of modules from a cloudformation template submitted by the customer, verify for each of them if they are already locally cached and if not, make a call against the cloudformation registry to download the module schemas and metadata to save them in the appropriate folder. Unit tests updated and added. 


src/cfnlint/registry.py
```diff
@@ -0,0 +1,87 @@
...
+import logging
+import os
+import platform
+import subprocess
+from cfnlint.template import Template
+
+LOGGER = logging.getLogger(__name__)
+
+
+class Registry(object):
```
lejiati, line 16: np: since this class does both managing resource schemas and interact with CFN registry. What about renaming it `SchemaManager` or some similar names?

src/cfnlint/registry.py
```diff
@@ -0,0 +1,87 @@
...
+            index = path_split.index('Users')
+            username = path_split[index + 1]
+        except ValueError:
+            raise ValueError
+
+        is_windows = platform.system() == 'win32'
+
+        for region in self.regions:
+            if is_windows:
+                path = 'C:/Users/%s/AppData/cloudformation/%s/%s/%s' % (username, account_id, region, name)
```
lejiati, line 40: Would suggest use [string format](https://docs.python.org/3/library/stdtypes.html#str.format) instead.

src/cfnlint/registry.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def create_folder(self, path, region, name, registry_type):
+        try:
+            os.makedirs(path)
+            self.aws_call_registry(region, name, registry_type, path)
+        except OSError:
+            raise OSError
+
+    def aws_call_registry(self, region, name, registry_type, path):
+        cmd = ['aws', '--region', region, 'cloudformation', 'describe-type', '--type', registry_type, '--type-name',
```
lejiati, line 56: Since you already initiated a Boto client, why not using Boto client to make a request. You can pass a Boto client during initialization.

src/cfnlint/registry.py
```diff
@@ -0,0 +1,87 @@
...
+                path = '/Users/%s/.cloudformation/%s/%s/%s' % (username, account_id, region, name)
+
+            if not os.path.isdir(path):
+                self.create_folder(path, region, name, registry_type)
+            return path
+
+    def create_folder(self, path, region, name, registry_type):
+        try:
+            os.makedirs(path)
+            self.aws_call_registry(region, name, registry_type, path)
```
lejiati, line 51: Looks like this method calls registry and save data into target folders. Would suggest split them into different methods and name them based on their functionalities: 
1. Calling registry.
2. Save them into folders.

Also `aws_call_registry` seems return 2 values but they are not used anywhere. If it's not used let's omit return values.

src/cfnlint/registry.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def aws_call_registry(self, region, name, registry_type, path):
+        cmd = ['aws', '--region', region, 'cloudformation', 'describe-type', '--type', registry_type, '--type-name',
+               name]
+
+        metadata = {}
+        schema = {}
+
+        # Recuperate detailed information about a registered extension
+        extension = json.loads(subprocess.check_output(cmd))
```
lejiati, line 63: When calling an AWS service, you need some error handling in case: 1. registry service fails to process request. 2. This type does not exist in that account.

test/unit/module/registry/test_aws_call_registry.py
```diff
@@ -0,0 +1,30 @@
...
+import cfnlint.core
+import cfnlint.registry
+from test.testlib.testcase import BaseTestCase
+
+from mock import patch
+
+LOGGER = logging.getLogger('cfnlint')
+
+
+class MyTestAWSCallRegistry(BaseTestCase):
```
lejiati, line 12: You don't need prefix `My` in unit test. Also test cases should cover failure scenarios as well.

test/unit/module/registry/test_create_files.py
```diff
@@ -0,0 +1,92 @@
...
+import cfnlint.core
+import cfnlint.registry
+from test.testlib.testcase import BaseTestCase
+
+from mock import patch, mock_open
+
+LOGGER = logging.getLogger('cfnlint')
+
+
+class MyTestCreateFiles(BaseTestCase):
```
lejiati, line 13: Same as above, you don't need prefix here.

test/unit/module/registry/test_create_folder.py
```diff
@@ -0,0 +1,39 @@
...
+import cfnlint.core
+import cfnlint.registry
+from test.testlib.testcase import BaseTestCase
+
+from mock import patch
+
+LOGGER = logging.getLogger('cfnlint')
+
+
+class MyTestCreateFolder(BaseTestCase):
```
lejiati, line 12: Same as above.

src/cfnlint/registry.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class Registry(object):
+    """Download schemas if necessary and validate modules"""
+
+    def __init__(
+            self, filename, template, regions):
+        self.filename = filename
+        self.regions = regions
+        self.cfn = Template(filename, template, regions)
```
lejiati, line 23: If it's a template, let's make sure its name reflect that.

src/cfnlint/registry.py
```diff
@@ -0,0 +1,87 @@
...
+from cfnlint.template import Template
+
+LOGGER = logging.getLogger(__name__)
+
+
+class Registry(object):
+    """Download schemas if necessary and validate modules"""
+
+    def __init__(
+            self, filename, template, regions):
```
lejiati, line 20: No need for a new line.

src/cfnlint/registry.py
```diff
@@ -0,0 +1,87 @@
...
+"""
+
+import json
+import logging
+import os
+import platform
+import subprocess
+from cfnlint.template import Template
+
+LOGGER = logging.getLogger(__name__)
```
lejiati, line 13: I didn't find any usage of logger in this file. Maybe we can either remove it or add logs somewhere such as when successfully download a schema?

src/cfnlint/registry.py
nhienton: Good suggestion

src/cfnlint/registry.py
nhienton: Great point!

setup.py
```diff
@@ -57,6 +57,7 @@ def get_version(filename):
         'networkx<=2.2;python_version<"3.5"',
         'junit-xml~=1.9',
         'pyrsistent<=0.16.0;python_version<"3.5"',
+        'future'
```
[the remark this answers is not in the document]
nhienton: To resolve the compatibility issue with Python 2.7:  `ImportError: No module named builtins`. The module builtins is used to mock the open() function in the unit tests

Issue #1963, opened 2021-03-29
Feature request: Validated parameters provided to nested stack against nested template.


hauboldj:
*cfn-lint version: (`cfn-lint --version`)*
0.48.0

*Description of issue.*
I frequently use nested stacks with SAM build. When I lint the root template and the nested stack is setup using `TemplateURL` pointing to a local file, I would like to have cfn-lint check that any parameters in the nested template without a default value are set by parameters in the root stack. Also I would like it to flag any parameters set by the root template that are not present in the nested template.

Checking the parameter types would be an amazing feature but might be difficult, so just matching the parameter names would be sufficient.

farski:
We tend to end up with 3 or 4 layers of nested stacks, and passing a value from a lower layer back to the top and then back down to a different stack happens pretty often. It's also a common reason for stack update failures, since you have to get outputs, template parameters, and stack parameters all updated. It's easy, but also easy to miss something. I have long wanted to request this feature, but didn't have any good ideas on how the linter would know where to look, since the structure of the resources in the stack and the files on disk likely don't match.

If it required some amount of hinting in a YAML comment:

```yaml
# LocalTemplateURL: ./asg.yml
TemplateURL: !Sub ${TemplateUrlPrefix}/asg.yml
```

That'd be work I'd be very happy to do to help catch these kinds of issues.

PR #2074, opened 2021-07-02
Add rule E3043 to validate nested stack parameters

kddejong:
*Issue #, if available:*
#1963

*Description of changes:*
- Add rule E3043 to validate nested stack parameters

This rule will not do the following.  It could be possible to add these capabilities on a future release. As a result I'm not closing the issue.
1. Fail if the template isn't found
2. Not validate parameters if the TemplateURL is not a string
3. Not validate parameters if the TemplateURL is pointed at a s3 bucket
4. Does not validate nested stack parameter outputs

### DIFF
--- a/test/unit/rules/resources/cloudformation/test_nested_stack_parameters.py
+++ b/test/unit/rules/resources/cloudformation/test_nested_stack_parameters.py
@@ -1,0 +1,28 @@
+"""
+Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+SPDX-License-Identifier: MIT-0
+"""
+from test.unit.rules import BaseRuleTestCase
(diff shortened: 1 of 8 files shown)


test/unit/rules/resources/cloudformation/test_nested_stack_parameters.py
```diff
@@ -0,0 +1,28 @@
+"""
+Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+SPDX-License-Identifier: MIT-0
+"""
+from test.unit.rules import BaseRuleTestCase
+from cfnlint.rules.resources.cloudformation.NestedStackParameters import NestedStackParameters  # pylint: disable=E0401
+
+
+class TestRulesRefCircular(BaseRuleTestCase):
+    """Test Rules Get Att """
```
nhienton, line 10: Detail : Forgot to change the description of the test

nhienton (approved):

PR #2076, opened 2021-07-08
Manage module schemas

nhienton:
*Issue #, if available:*

*Description of changes:*
Extract the list of modules from the CloudFormation template. For each of them, access the CloudFormation registry to retrieve module schemas and metadatas. Store the json files in the appropriate folder. 


src/cfnlint/registry.py
```diff
@@ -0,0 +1,87 @@
...
+import logging
+import os
+import platform
+import subprocess
+from cfnlint.template import Template
+
+LOGGER = logging.getLogger(__name__)
+
+
+class Registry(object):
```
PatMyron, line 16: Is this file just leftover from renaming this class? Can this file be removed if so?
https://github.com/aws-cloudformation/cfn-lint/pull/2056#discussion_r657306850

nhienton: I didn't catch it while resolving the merge conflicts. Thank you, will remove it!

src/cfnlint/schemaManager.py
```diff
@@ -0,0 +1,96 @@
...
+LOGGER = logging.getLogger(__name__)
+
+
+class SchemaManager(object):
+    """Download schemas if necessary and validate modules"""
+    def __init__(
+            self, filename, template, regions):
+        self.filename = filename
+        self.regions = regions
+        self.template = Template(filename, template, regions)
```
lejiati, line 24: Would it be better if STS client and CFN client are initialized as instance variables of `SchemaManager`? So you don't have to initialize and passed them to method every single time.

src/cfnlint/schemaManager.py
```diff
@@ -0,0 +1,96 @@
...
+        except ValueError:
+            raise ValueError
+
+        is_windows = platform.system() == 'win32'
+
+        for region in self.regions:
+            if is_windows:
+                path = 'C:/Users/{0}/AppData/cloudformation/{1}/{2}/{3}'.format(username, account_id, region, name)
+            else:
+                path = '/Users/{0}/.cloudformation/{1}/{2}/{3}'.format(username, account_id, region, name)
```
lejiati, line 43: nitpick - you can extract them into separate method.

src/cfnlint/schemaManager.py
```diff
@@ -0,0 +1,96 @@
...
+        # Recuperate detailed information about a registered extension
+        response = None
+        try:
+            response = client.describe_type(
+                Type=registry_type,
+                TypeName=name
+            )
+        except botocore.exceptions.ClientError as e:
+            if e.response['Error']['Code'] == 'TypeNotFoundException':
+                raise e
```
lejiati, line 66: Would this show as a lint error or just a general python exception? This looks like some errors lint should catch and surface to user.

src/cfnlint/core.py
```diff
@@ -240,6 +243,20 @@ def run_checks(filename, template, rules, regions, mandatory_rules=None, validat
...
 
+
+    # MODULE is the only accepted value right now, but eventually other values will be supported such as RESOURCE
+    for registry_type in validate_registry_types:
+        if registry_type == 'MODULE':
+            # Check the presence of modules in the template
+            template_obj = Template(filename, template, regions)
+            modules = template_obj.get_modules()
+            if modules:
+                schema_manager = SchemaManager(filename, template, regions)
```
lejiati, line 254: Is it possible to cover these lines?

lejiati (commented):
Overall looks good! Just some minor comments.

src/cfnlint/schemaManager.py
nhienton: Great idea!

src/cfnlint/schemaManager.py
nhienton: Done

lejiati (approved):

src/cfnlint/schemaManager.py
```diff
@@ -0,0 +1,103 @@
...
+        account_id = self.boto3_sts.get_caller_identity().get('Account')
+        username = None
+        path_split = os.getcwd().split('/')
+        try:
+            index = path_split.index('Users')
+            username = path_split[index + 1]
+        except ValueError:
+            raise ValueError
+
+        is_windows = platform.system() == 'win32'
```
PatMyron, line 39: 📟  paging Windows expert @miparnisari / python expert @kddejong

thinking there might be some directory path logic that's more OS-agnostic and user-agnostic

is there some standard directory for Python packages that can be referenced instead?

src/cfnlint/core.py
```diff
@@ -240,6 +241,18 @@ def run_checks(filename, template, rules, regions, mandatory_rules=None, validat
...
+                # For each module extracted from the template, verify if it's already locally cached
+                for module in modules:
```
PatMyron, line 253: (not necessary, but future idea depending on speed)
could probably be done in [parallel](https://docs.python.org/3/library/multiprocessing.html) like we did for https://github.com/aws-cloudformation/cfn-lint/issues/1249#issuecomment-562801400

src/cfnlint/schemaManager.py
```diff
@@ -0,0 +1,103 @@
...
+
+        for region in self.regions:
```
PatMyron, line 41: (not necessary, but future idea depending on speed)
could probably be done in [parallel](https://docs.python.org/3/library/multiprocessing.html) like we did for https://github.com/aws-cloudformation/cfn-lint/issues/1249#issuecomment-562801400

src/cfnlint/schemaManager.py
PatMyron: maybe something like these?

https://github.com/aws-cloudformation/cfn-lint/blob/29886f432989e4b32fa81f98e76208d740a857cb/src/cfnlint/helpers.py#L244

https://github.com/aws-cloudformation/cfn-lint/blob/29886f432989e4b32fa81f98e76208d740a857cb/src/cfnlint/maintenance.py#L53

https://github.com/aws-cloudformation/cfn-lint/blob/29886f432989e4b32fa81f98e76208d740a857cb/src/cfnlint/maintenance.py#L199

test/unit/module/registry/test_create_path.py
```diff
@@ -0,0 +1,42 @@
...
+
+from mock import patch
+
+LOGGER = logging.getLogger('cfnlint')
+
+
+class TestCreatePath(BaseTestCase):
+    """Test Create Path """
+
+    @patch('platform.system')
```
lejiati, line 16: Based on [this stack overflow answer](https://stackoverflow.com/questions/18191275/using-pythons-mock-patch-object-to-change-the-return-value-of-a-method-called-w), `@patch` allows you to control the return value of certain calls. So instead of using `if-else`, you could set the return value of this call. Since this test case is intend for Windows, you can change to 
```python
    @patch('platform.system')
    def test_create_path_windows(self, platform):
        """Create path (windows)"""
        filename = 'test/fixtures/templates/good/generic.yaml'
        (args, filenames, _) = cfnlint.core.get_args_filenames(['--template', filename])
        (template, rules, _) = cfnlint.core.get_template_rules(filename, args)
        platform.system.return_value = 'win32'
        schema_manager = cfnlint.schemaManager.SchemaManager(filename, template, ['us-east-1'])
        assert schema_manager.create_path(True, 'username', 'account_id', 'region', 'name') == \
                r'C:\Users\username\AppData\cloudformation\account_id\region\name'
```

setup.py
```diff
@@ -57,6 +57,7 @@ def get_version(filename):
         'networkx<=2.2;python_version<"3.5"',
         'junit-xml~=1.9',
         'pyrsistent<=0.16.0;python_version<"3.5"',
+        'future'
```
kddejong, line 60: where is this coming from?

nhienton: To resolve the compatibility issue with Python 2.7: ImportError: No module named builtins. The module builtins is used to mock the open() function in the unit tests

src/cfnlint/schemaManager.py
```diff
@@ -0,0 +1,103 @@
+"""
```
kddejong, line 1: I feel like this is a new naming standard?  The file name should be `schema_manager`

src/cfnlint/schemaManager.py
kddejong: Hmm, yea I wonder if this should be like this or not.  A lot of very OS and possible issues with specific configurations that has me a little concerned.

src/cfnlint/schemaManager.py
```diff
@@ -0,0 +1,103 @@
...
+        self.regions = regions
+        self.template = Template(filename, template, regions)
+        self.boto3_sts = boto3.client('sts')
+        self.boto3_cfn = boto3.client('cloudformation')
+
+    # Check if the appropriate folder already exists
+    def check_folders(self, name, registry_type):
+        account_id = self.boto3_sts.get_caller_identity().get('Account')
+        username = None
+        path_split = os.getcwd().split('/')
```
kddejong, line 32: We should talk through what the goal is here.  Are we trying to put the files where cfn-lint is installed? into a user cache folder?  Doesn't cwd return where we are running cfn-lint?  What if it isn't in the users folder?

kddejong: Additionally what if it its a docker container? So many cfn-lint runs are inside a docker container.

src/cfnlint/schemaManager.py
```diff
@@ -0,0 +1,103 @@
...
+
+        for region in self.regions:
+            path = self.create_path(is_windows, username, account_id, region, name)
+
+            if not os.path.isdir(path):
+                self.create_folder(path, name, registry_type)
+
+    def create_path(self, is_windows, username, account_id, region, name):
+        if is_windows:
+            path = os.path.join(r'C:\Users', username, 'AppData', 'cloudformation', account_id, region, name)
```
kddejong, line 49: I think there are better, more python methods, for getting a users home directory.  https://docs.python.org/3/library/os.path.html#os.path.expanduser and https://docs.python.org/3/library/pathlib.html#pathlib.Path.home

test/fixtures/results/quickstart/openshift.json
```diff
@@ -394,6 +394,33 @@
...
+                "KeyGen",
+                "Properties",
+                "Runtime"
+            ],
+            "Start": {
+                "ColumnNumber": 7,
+                "LineNumber": 810
+            }
+        },
+        "Message": "EOL runtime (python2.7) specified. Runtime is EOL since 2021-07-15 and updating will be disabled at 2021-09-30. Please consider updating to python3.8",
```
kddejong, line 416: Is this because we haven't synced to main?

kddejong (changes requested):

src/cfnlint/schema_manager.py
```diff
@@ -0,0 +1,87 @@
...
+        self.template = Template(filename, template, regions)
+        self.boto3_sts = boto3.client('sts')
+        self.boto3_cfn = boto3.client('cloudformation')
+
+    # Check if the appropriate folder already exists
+    def check_folders(self, name, registry_type):
+        account_id = self.boto3_sts.get_caller_identity().get('Account')
+        for region in self.regions:
+            path = self.create_path(account_id, region, name)
+            if not os.path.isdir(path):
```
kddejong, line 32: So we only download the metadata once?  What if the version of the resource type changes?

src/cfnlint/schema_manager.py
```diff
@@ -0,0 +1,87 @@
...
+        self.regions = regions
+        self.template = Template(filename, template, regions)
+        self.boto3_sts = boto3.client('sts')
+        self.boto3_cfn = boto3.client('cloudformation')
+
+    # Check if the appropriate folder already exists
+    def check_folders(self, name, registry_type):
+        account_id = self.boto3_sts.get_caller_identity().get('Account')
+        for region in self.regions:
+            path = self.create_path(account_id, region, name)
```
kddejong, line 31: It seems like we pass the path around a lot.  Is there a better way to do this?  Like can we set the base path once in the init `base_path = os.path.join(os.path.expanduser('~'), '.cloudformation', account_id)`

setup.py
kddejong: Weird.  Thought we were doing some tests with the open that didn't require this.  I'm okay with this now but would like to see if we can figure out a better way to do this.

kddejong (approved):

src/cfnlint/schema_manager.py
nhienton: In a future PR, the --update-registry-type-specs parameter will allow the customer to refresh their locally cached registry type information. A check of the current version id will also throw a warning message to the customer to notify them that a later version is available in the CloudFormation registry every time that they validate their private registry types.

PR #2091, opened 2021-07-30
Refresh locally cached registry type information

nhienton:
*Issue #, if available:*

*Description of changes:*

--update-registry-type-specs allows customers to refresh their locally cached registry type information. The check is also made during the validation with --validate-registry-types. New rules added to indicate if a registry type is not found in the CloudFormation registry and if a later version is available in the CloudFormation registry.


src/cfnlint/core.py
```diff
@@ -161,6 +162,12 @@ def get_args_filenames(cli_args):
         print(rules)
         sys.exit(0)
 
+    if config.update_registry_type_specs:
+        schema_manager = SchemaManager(config.regions)
+        schema_manager.update_locally_cached_schemas()
+        print('Locally cached registry type information from specified regions is now refreshed.')
```
lejiati, line 168: Suggestion: maybe explicitly return what region is used here helps?

lejiati (approved):

PR #2096, opened 2021-08-04
Feature/module validation

nhienton:
*Issue #, if available:*

*Description of changes:*
Added 3 new rules for module validation (module properties, !Ref and !GetAtt). Added unit tests for these rules.


src/cfnlint/rules/modules/GettAttModule.py
```diff
@@ -0,0 +1,83 @@
...
+                resname = None
+                restype = None
+                # YAML template
+                if isinstance(getatt[-1][1], six.string_types):
+                    resname = getatt[-1][0]
+                    restype = '.'.join(getatt[-1][1:])
+            # if it's a module
+            if any(resname.startswith(s) for s in valid_getatts) and any(resname.startswith(m) for m in
+                                                                         modules_keys):
+                for module in modules_keys:
```
lejiati, line 41: Ideally, this block should be split into its own method.

src/cfnlint/rules/modules/RefModuleExists.py
```diff
@@ -0,0 +1,51 @@
...
+
+    def match(self, cfn):
+        matches = []
+
+        # Build the list of refs
+        reftrees = cfn.search_deep_keys('Ref')
+        valid_refs = cfn.get_valid_refs()
+
+        # start with the basic ref calls
+        for reftree in reftrees:
```
lejiati, line 27: Same applies here. We don't want too many indentations.

lejiati (approved):

PR #2103, opened 2021-08-18
Change the default configuration for rule E3012 to not be strict

kddejong:
*Issue #, if available:*

*Description of changes:*
- Change the default configuration of rule E3012 to not be strict

There are a few reasons for this change:
1. All values are passed in to the code that does the provisioning using strings
https://github.com/aws-cloudformation/cloudformation-cli-python-plugin/blob/master/src/cloudformation_cli_python_lib/recast.py
https://github.com/aws-cloudformation/cloudformation-cli-go-plugin/blob/master/cfn/encoding/stringify.go
and more examples.
2. Certain properties can have multiple types defined with the new resource type schemas 


PatMyron:
linking https://github.com/aws-cloudformation/cfn-lint/issues/547

PR #2126, opened 2021-09-26
Adds support for outputting results in SARIF

kadrach:
Adds support for outputting results in the Static Analysis Results Interchange Format (SARIF) Version 2.1.0.

*Issue #, if available:*
n/a

*Description of changes:*
Adds a `-f sarif` option that will output `cfn-lint` results in a SARIF JSON file. This file can be ingested by code scanning/reporting tools. For example, on Github:

Screen Shot 2021-09-26 at 11 50 30 am
Screen Shot 2021-09-26 at 11 50 06 am


src/cfnlint/formatters/__init__.py
```diff
@@ -246,3 +251,93 @@ def _format_matches(self, matches):
...
+        if not rules:
+            rules = []
+
+        run = sarif.Run(
+            tool=sarif.Tool(
+                driver=sarif.ToolComponent(
+                    name='cfn-lint',
+                    short_description=sarif.MultiformatMessageString(
+                        text=('Validates AWS CloudFormation templates against'
+                              'the resource specification and additional checks')
```
PatMyron, line 291:
```suggestion
                              ' the resource specification and additional checks')
```

PatMyron (commented):
can update this README snippet as well:
https://github.com/aws-cloudformation/cfn-lint/blob/34affe23fde9526b50f0800f2fdbb83d6c6bab1b/README.md#L156

PatMyron:
Love standardized formats, posting a sample since this one's new to me:

Screen Shot 2021-09-26 at 3 03 34 PM

@grantmcconnaughey another format that might help https://github.com/grantmcconnaughey/Lintly

@lgolding would love a quick comparison between SARIF and other formats like JUnit XML/Checkstyle

kadrach:
The spec itself is quite long, the SARIF output produced here uses the information that is immediately available to `cfn-lint` formatters. Also a call-out that GitHub only supports a subset of the spec, and it is right now the primary tool I am looking to use this with. 

Happy to add additional information into the output if there's something valuable that's currently missing.

lgolding:
@PatMyron Glad to see you're using SARIF! Although I no longer work on the format, @michaelcfanning (also an Editor of the specification) might have some thoughts.

michaelcfanning:
Hello, nice to see this support. We have a fair amount of experience helping prep SARIF for GitHub advanced security ingestion. A very good start on ramping up on this complex format (as you've noted) is to put it through the [SARIF Validator](https://sarifweb.azurewebsites.net/Validation). Be sure to check the `GitHub ingestion rules` checkbox, to get some special analysis for that scenario.

If you can provide a sample log file to us here at MS, we'll provide some additional review. 

@EddyNaka, FYI

lgolding:
@PatMyron Also (since as you rightly point out, the spec is quite long), you might like the [SARIF Tutorials](https://github.com/microsoft/sarif-tutorials).

eddynaka:
@kadrach ,

Below, a few suggestions:
- only output the rules that you have a result (that will make your SARIF more optimized)
- a rule can contain a property called `messageStrings` and that can be a reference to the result (that will make your SARIF more optimized) (https://github.com/microsoft/sarif-tutorials/blob/main/samples/3-Beyond-basics/message-with-arguments.sarif)
- all phrases should end with a '.'

More like a question: I saw that your result has a location, and the location is a relative path.
does the analysis generate the full path and u are trimming or is it something else?
If you are trimming, you could leverage the `OriginalBaseUriIds` property: https://github.com/microsoft/sarif-tutorials/blob/main/samples/OriginalUriBaseIds.sarif

Let us know if you need help with that.
Thank you!

kadrach:
@eddynaka Thank for your your insights. Is there value in knowing which rules ran, but did not trigger?

I will dig into the paths. At the moment the path is relative to the directory the linter is invoked in. From my personal experience that's typically repository root, but this is not guaranteed. I can add a `originalUriBaseId` for it.

`CloudFormationLintRule` do not current expose the message strings in a way that is easily consumable for formatting (outside of the rule class). In regards to messages ending with '.', consistency with existing output formats is probably preferable (but I'm not above appending a '.' to all messages in the SARIF formatter). @PatMyron

eddynaka:
> @eddynaka Thank for your your insights. Is there value in knowing which rules ran, but did not trigger?
> 
> I will dig into the paths. At the moment the path is relative to the directory the linter is invoked in. From my personal experience that's typically repository root, but this is not guaranteed. I can add a `originalUriBaseId` for it.
> 
> `CloudFormationLintRule` do not current expose the message strings in a way that is easily consumable for formatting (outside of the rule class). In regards to messages ending with '.', consistency with existing output formats is probably preferable (but I'm not above appending a '.' to all messages in the SARIF formatter). @PatMyron

@kadrach , instead of starting to write the rules first, start with the results. When you do that, you can create a mapper result <-> rule and once you walked in all results, you will see the distinct rules. From that, you output it.

kadrach:
[Updated sample](https://gist.github.com/kadrach/f09193d895ba4570892df56c83633259).

Just noticed it barfs on `E0000` (parsing) rule, will update this shortly.

eddynaka:
> [Updated sample](https://gist.github.com/kadrach/f09193d895ba4570892df56c83633259).
> 
> Just noticed it barfs on `E0000` (parsing) rule, will update this shortly.

the latest SARIF looks good!

eddynaka:
Once we publish this version that exports SARIF, we can create an action here https://github.com/actions/starter-workflows/pulls that bundles everything together: cfn-lint + exporting sarif -> send to github.

This action normally appears in the `security/code-scanning/setup` tab.

NickLiffen:
@eddynaka there is already an action [here](https://github.com/marketplace/actions/cfn-lint-action). 

Once this PR is merged and published was looking to update that action to support SARIF upload to GitHub Code Scanning. 👍 

I think it would be great to add a starter workflow that uses `cfn-lint` here: https://github.com/actions/starter-workflows/tree/main/code-scanning as this is where the code scanning workflows go 👍 

This is going to be great btw 💯

eddynaka:
> Hey @eddynaka, if I understand your suggestion correctly, you are asking to move `runs[].results[].message` into the rules block?
> 
> The messages currently contain the individual findings, e.g. the [latest example](https://gist.github.com/kadrach/f09193d895ba4570892df56c83633259#file-lint-sarif-json-L29) warns about a reference `PrivateSecurityGroup` that does not exist.
> 
> https://github.com/aws-cloudformation/cfn-lint/blob/0c5d63e0f681fde3c385514c08aefb9aac13fbaa/test/fixtures/templates/bad/issues.yaml#L9-L12
> 
> These are generated within the individual rules, and are specific to each finding.
> 
> https://github.com/aws-cloudformation/cfn-lint/blob/4a7af2bd53a9ad1ccaba3a509437c53102ade522/src/cfnlint/rules/functions/RefExist.py#L37-L42
> 
> Moving this message block into the rule is not possible unless we utilize `messageStrings`. This probably involves refactoring all existing rules, but happy to be proven wrong :)

Hello,

yes, you would need to refactor to accept the messageStrings.
This is an optimization in sarif that prevents duplicated information.

kadrach:
Thank you! I don't see this as in-scope of this PR. @kddejong, what do you reckon?

kddejong (approved):

Issue #2136, opened 2021-10-13
Unsafe call to glob causes error if filename contains special characters


alex-harvey-z3q:
*cfn-lint version: 0.54.1*

Given a CloudFormation template whose filename contains `[` and `]`, due to unsafe call to `glob.glob` the following is seen:

```
▶ cfn-lint emr/tests/__snapshots__/test_snapshot\[emr--tests--emr.template.2.yaml\].yaml
Traceback (most recent call last):
  File "/Users/alexharvey/git/fox/streamotion-platform-sceptre/venv/bin/cfn-lint", line 8, in <module>
    sys.exit(main())
  File "/Users/alexharvey/git/fox/streamotion-platform-sceptre/venv/lib/python3.8/site-packages/cfnlint/__main__.py", line 26, in main
    (args, filenames, formatter) = cfnlint.core.get_args_filenames(sys.argv[1:])
  File "/Users/alexharvey/git/fox/streamotion-platform-sceptre/venv/lib/python3.8/site-packages/cfnlint/core.py", line 177, in get_args_filenames
    if not config.templates:
  File "/Users/alexharvey/git/fox/streamotion-platform-sceptre/venv/lib/python3.8/site-packages/cfnlint/config.py", line 587, in templates
    add_filenames = glob.glob(filename, recursive=True)
  File "/Users/alexharvey/.pyenv/versions/3.8.9/lib/python3.8/glob.py", line 21, in glob
    return list(iglob(pathname, recursive=recursive))
  File "/Users/alexharvey/.pyenv/versions/3.8.9/lib/python3.8/glob.py", line 74, in _iglob
    for name in glob_in_dir(dirname, basename, dironly):
  File "/Users/alexharvey/.pyenv/versions/3.8.9/lib/python3.8/glob.py", line 85, in _glob1
    return fnmatch.filter(names, pattern)
  File "/Users/alexharvey/.pyenv/versions/3.8.9/lib/python3.8/fnmatch.py", line 52, in filter
    match = _compile_pattern(pat)
  File "/Users/alexharvey/.pyenv/versions/3.8.9/lib/python3.8/fnmatch.py", line 46, in _compile_pattern
    return re.compile(res).match
  File "/Users/alexharvey/.pyenv/versions/3.8.9/lib/python3.8/re.py", line 252, in compile
    return _compile(pattern, flags)
[...]

PR #2137, opened 2021-10-13
Escape filename in glob in ConfigMixIn (#2136)

alexharv074:
Call glob.escape on filename to handle any unsafe characters prior to
globbing.


kddejong:
I cannot replicate the issue locally.  Trying to figure out why this is failing.

alexharv074:
@kddejong Are you saying you weren't able to reproduce this even with the details I gave in the issue here? https://github.com/aws-cloudformation/cfn-lint/issues/2136

kddejong:
Here is the issue from what I can tell `test/fixtures/templates/public/*.yaml` is becoming `test/fixtures/templates/public/[*].yaml` which is resulting in the `*` not being expanded.

kddejong:
Going to close this one.  We can re-open as needed.

PR #2173, opened 2021-11-15
Extending linter to fully support translated SAM templates

andrew-glenn:
Depends on: https://github.com/aws/serverless-application-model/pull/2224

In translated SAM templates, metadata is not currently copied to new resources. In addition, line-numbers are not updated to reflect the new template data. As a result of both concerns, it's impossible to add in-template metadata to exclude rules in cfn-lint. 

This PR round-trips the transformed template so that the internal loading mechanism can apply proper line numbers. This PR additionally copies parent-resouce-metadata to child-resources that are created in the transform workflow.

Exposing the parent/child relationship is dependent on the referenced PR being merged.

### DIFF
--- a/src/cfnlint/transform.py
+++ b/src/cfnlint/transform.py
@@ -7,6 +7,7 @@ import logging
 import six
 import samtranslator
 import copy
+import json
 from samtranslator.parser import parser
 from samtranslator.translator.translator import Translator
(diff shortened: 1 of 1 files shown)


andrew-glenn:
This has been handled via an upstream merge to the SAM CLI. Closing.

PR #2210, opened 2022-02-02
W2506: Avoid false positives against resources

teekennedy:
*Description of changes:*

W2506, the rule that checks that `ImageId` refs are for the appropriate parameter type, produces false positives when the reference is to a resource. I encountered this issue when creating a custom resource type that builds an AMI out of a given EC2 instance, but I could see this applying to CloudFormation registry resources or even official resources released in the future.

Fixing the false positive involved adding one more condition to the rule to check that the `Ref` is for a parameter like the rule is intended.


jfriend-citrine (approved):
lgtm

kddejong (approved):

Issue #2228, opened 2022-03-04
[cfn-lint] E3030 for config resources


theksi:
*cfn-lint version: (`cfn-lint --version`)*
cfn-lint 0.58.2
*Description of issue.*
Error on an invalid value for Compliance resourceTypes that is in fact valid

Please provide as much information as possible:
* Template linting issues:*
```Shell
[{
	"resource": "XXXXX/dev_local/aws-config-library/account-quality/templates/cloudwatch.yaml",
	"owner": "_generated_diagnostic_collection_name_#4",
	"severity": 8,
	"message": "[cfn-lint] E3030: You must specify a valid value for ComplianceResourceTypes (AWS::Logs::LogGroup). Valid values are [\"AWS::ACM::Certificate\", \"AWS::ApiGateway::RestApi\", \"AWS::ApiGateway::Stage\", \"AWS::ApiGatewayV2::Api\", ...]
	"startLineNumber": 57,
	"startColumn": 15,
	"endLineNumber": 57,
	"endColumn": 34
}]
```
  * Please provide a CloudFormation sample that generated the issue.
```yaml
  QnACwLogRetentionRule:
      Type: AWS::Config::ConfigRule
      Properties:
        ConfigRuleName: !Sub ${RuleNamePrefix}-QnACwLogRetention-${AWS::Region}
        Description: "Every Cloudwatch log must have a set retention"
        Scope:
          ComplianceResourceTypes:
            - AWS::Logs::LogGroup
        Source:
          Owner: AWS
          SourceIdentifier: CW_LOGGROUP_RETENTION_PERIOD_CHECK
```
  * If present, please add links to the (official) documentation for clarification.
  - https://docs.aws.amazon.com/config/latest/developerguide/cw-loggroup-retention-period-check.html
  * Validate if the issue still exists with the latest version of `cfn-lint` and/or the latest Spec files
  N/A using the latest version
* Feature request:
  * Please provide argumentation about the missing feature. Context is key!
When designing cfn template for scheduled rules, any value for ComplianceResourceTypes can be used as long as it is a valid AWS resource type (and not only a supported resource type).
When a scheduled rule is described in the template, cfn-lint should verify all AWS resources.
Suggestion : 
Identifying an AWS scheduled rule is not easy as `MaximumExecutionFrequency` is not mandatory and cfn-lint evaluation could work this way:
[...]

PR #2231, opened 2022-03-07
Remove Config resource type allowed values

kddejong:
*Issue #, if available:*
Fix issue #2228 

*Description of changes:*
- AWS Config supports custom resource types from the CloudFormation registry making this value hard to keep accurate.  We will remove it leaving the value to be open.


src/cfnlint/data/ExtendedSpecs/all/03_value_types/aws_config.json
```diff
@@ -1,11 +1,4 @@
 [
-  {
-    "op": "add",
-    "path": "/ValueTypes/AWS::Config::ConfigurationRecorder.ResourceTypes",
```
PatMyron, line 4: think the references in https://github.com/aws-cloudformation/cfn-lint/blob/main/src/cfnlint/data/ExtendedSpecs/all/04_property_values/aws_config.json would also need to be removed

(also wondering if we should have a test to catch that)

kddejong: the only test we have for this is on the spec itself.  I updated the spec in this PR now too based on the changes so we should have this tested now.

PR #2242, opened 2022-03-25
Move null value checks from template decoding to E3012

### LINKED ISSUE #1657: Using null in AWS::Serverless::StateMachine in the ResultPath throws incorrect E0000 error
StateMachines allow for a null value in the ResultPath Property to [pass input directly to the output](https://docs.aws.amazon.com/step-functions/latest/dg/input-output-resultpath.html)  

Below is a minimal example of a template that is proper but flagged as invalid

```

Resources:
  rLambda:
    Properties:
      Handler: index.handler
      Runtime: python3.6
      InlineCode: |
        def handler(event, context):
          print(event)
          return


  rTestMachine:
    Type: AWS::Serverless::StateMachine
    Properties:
      Definition: 
        StartAt: myState
        States:
          myState:
            Type: Task
            Resource:  !GetAtt rLambda.Arn
            ResultPath: null
            End: true
      Policies:
      - Statement:
        - Sid: Test
          Effect: Allow
          Action:
          - lambda:Invoke*
          Resource: !GetAtt rLambda.Arn
```

Adding quotes around null does not resolve the issue as the value of the property needs to evaluate to null. An exception for null checking should be added for this Parameter

kddejong:
*Issue #, if available:*
fix #1657

CloudFormation service does not accept null values but it does accept JSON/YAML properties that have null values in them. As a result we need to move the null value check from the decoding process and put it into a rule.

Possible impacts from this change is that rules will fail.  Before this change rule writers didn't have to worry about a property value having a null value as it was found in the decoding process.  After this change it is possible some rules will fail when encountering a null value.

*Description of changes:*
- Remove null checks from the decoding process
- Add checks for null values into rule E3012
- Update template `check_value` to have a parameter `pass_if_null` to pass through null values. Previously this wasn't passed so rule writers wouldn't have to account for null values.  To maintain backwards compatibility the default is `false`


kddejong (commented):
I think this results in missing null values in other areas of the template in which there is no spec. Null values in parameters, conditions, etc.

kddejong:
Added in all the appropriate checks for parameters and null values

PR #2276, opened 2022-05-24
Point to the general "Rules" documentation when a rule doesn't specify a `source_url`

ankon:
The SARIF schema defines `help_uri` as a `uri`, and it cannot be empty -- if it is then Github's codeql-action
to upload the SARIF will fail with validation errors, for instance.

I'm a bit unsure who is exactly at fault here, maybe the problem is actually that the sarif-om library shouldn't write the `help_uri` if it is `None`?

### DIFF
--- a/src/cfnlint/formatters/__init__.py
+++ b/src/cfnlint/formatters/__init__.py
@@ -327,7 +327,7 @@ class SARIFFormatter(BaseFormatter):
                 full_description=sarif.MultiformatMessageString(
                     text=rules_map[rule_id].description
                 ),
-                help_uri=rules_map[rule_id].source_url if rules_map[rule_id] else "https://github.com/aws-cloudformation/cfn-lint/blob/main/docs/rules.md"
+                help_uri=rules_map[rule_id].source_url if rules_map[rule_id] else 'https://github.com/aws-cloudformation/cfn-lint/blob/main/docs/rules.md'
             )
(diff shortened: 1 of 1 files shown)


ankon:
Workaround: `cfn-lint -f sarif TEMPLATE | jq 'del(.runs[].tool.driver.rules[].helpUri|select(.==""))' > output.sarif` removes the empty entries so that the sarif validates against the schema.

kddejong (approved):

src/cfnlint/formatters/__init__.py
```diff
@@ -327,7 +327,7 @@ def print_matches(self, matches, rules=None, filenames=None):
                 full_description=sarif.MultiformatMessageString(
                     text=rules_map[rule_id].description
                 ),
-                help_uri=rules_map[rule_id].source_url if rules_map[rule_id] else None
+                help_uri=rules_map[rule_id].source_url if rules_map[rule_id] else "https://github.com/aws-cloudformation/cfn-lint/blob/main/docs/rules.md"
```
PatMyron, line 330: `pylint` failing:
```suggestion
                help_uri=rules_map[rule_id].source_url if rules_map[rule_id] else 'https://github.com/aws-cloudformation/cfn-lint/blob/main/docs/rules.md'
```

PatMyron (approved):

PR #2285, opened 2022-06-11
Proposal: New Simplified API Accepting Strings

LaurenceWarne:
Hi, I was wondering if I could get your thoughts on this API proposal.  The main feature as opposed to https://github.com/aws-cloudformation/cfn-lint/blob/main/docs/getting_started/integration.md is that you can now validate strings directly (so useful if your template is not on your file system).  The API is fairly simple, we have:

-  `lint(s, rules, regions)` which would allow you validate a string against a custom set of rules and regions
-  `lint_all(s)`, a convenience function which validates against everything (all rules and regions).

~The functions (well the main one) are stubbed for now, I thought this would make more sense than submitting a feature request since I've done something similar on a project I'm working on, so in terms of the implementation, I'd be happy to work on it on top of this if you'd be open to it.~ EDIT: there's now an implementation.

Possibly related is https://github.com/aws-cloudformation/cfn-lint/issues/2243, apologies if something like this has been discussed before, I couldn't find anything related in the issue tracker or similar PRs.

Thanks, let me know what you think!


LaurenceWarne:
Great!  I've pushed an initial implementation if you want to take a look.

I've extracted out the string decoding from  `cfnlint.decode` (possibly there's some deduplication to be done here, but I think that would involve making `decode_str` take a optional filename argument which I don't really like), from there most of the work is simply done by allowing `Match` to take a `None` for a file name.

Example usage:
```python
>>> from cfnlint.api import lint_all
>>> s = """AWSTemplateFormatVersion: 2010-09-09
... Description: An example CloudFormation template for Fargate.
... Resources:
...   Cluster:
...     Type: AWS::ECS::Cluster
...     Properties:
...       ClusterName: !Join ['', [!Ref ServiceName, Cluster]]
... """
>>> lint_all(s)
[[I1022: Use Sub instead of Join] (Prefer using Fn::Sub over Fn::Join with an empty delimiter) matched 7, [E1012: Check if Refs exist] (Ref ServiceName not found as a resource or parameter) matched 7]
```

Let me know if this addresses what you've written above, still TODO is to expand on the unit tests and write documentation.

src/cfnlint/decode/__init__.py
```diff
@@ -16,6 +16,52 @@
 LOGGER = logging.getLogger(__name__)
 
 
+def decode_str(s):
```
kddejong, line 19: Seems like repetitive logic which is unfortunate but looking at the structure of the current code I understand why you did this.

src/cfnlint/rules/__init__.py
```diff
@@ -474,22 +474,30 @@ def __init__(
...
+        """The message of this match"""
         if rulematch_obj:
             for k, v in vars(rulematch_obj).items():
                 if not hasattr(self, k):
                     setattr(self, k, v)
 
     def __repr__(self):
         """Represent"""
-        formatstr = u'[{0}] ({1}) matched {2}:{3}'
+        file_str = self.filename + ':' if self.filename else ''
```
kddejong, line 497: Whats this do?  Didn't change any tests but trying to understand

kddejong (commented):
Looks pretty good just wish you didn't have to create a `decode_str`

kddejong:
Oh and we should also update the integration documentation to show how to use the APIs.

The result of this would allow you to pass in strings which will add to stdin and using filenames

src/cfnlint/rules/__init__.py
LaurenceWarne: Since `filename` can now take the value `None`, this makes it so that in such cases `filename` is omitted, so you would just get:
`"[E1012: Check if Refs exist] (Ref ServiceName not found as a resource or parameter) matched 7"` 
Instead of:
`"[E1012: Check if Refs exist] (Ref ServiceName not found as a resource or parameter) matched None:7"`

LaurenceWarne:
> Looks pretty good just wish you didn't have to create a decode_str

Yeah, I don't like it either.  I had another look, but it looks like I would have to do stuff like `if filename: LOGGER.log("foo")` in `decode_str` to stay consistent with the behaviour of `decode`, if we were to allow `decode_str` to take an optional `filename` argument.  Of course other suggestions on how to deduplicate welcome :slightly_smiling_face: 

> Oh and we should also update the integration documentation to show how to use the APIs.

:+1: I've done so in https://github.com/aws-cloudformation/cfn-lint/pull/2285/commits/cd8245d50f4fe11b65759f1a9645943c3f1b9f1c, I've removed the existing doc. I can re-add it if you like, I wasn't sure if it was necessary since if you wanted to lint a file you could just:

```python
with open(file, 'r') as f:
    lint_all(f)
```

src/cfnlint/rules/__init__.py
kddejong: Got it.

src/cfnlint/decode/__init__.py
kddejong: What about doing something like this? 

https://github.com/kddejong/cfn-python-lint/blob/LaurenceWarne-new-api/src/cfnlint/decode/__init__.py#L19-L104

LaurenceWarne: Yeah lgtm, I've pushed ^ with a few minor tweaks, I think I was overthinking re logging since it's unconditionally disabled in `lint(s)` anyway.

kddejong (approved):

LaurenceWarne:
Great, thanks :tada:

PR #2293, opened 2022-06-26
Add explicit encoding

shunkt:
*Issue #, if available:*

*Description of changes:*

Add explicit `encoding='utf-8'` argument when cfn-lint loads .yml file.

  Some languages' environment such as Windows in Japanese, default encodings is not UTF-8. UTF-8 encoded template file could not be read without implicit `encoding` argument. For example, when cfn-lint runs template with comment in Japanese in Japanese-Windows environment, it cannot work because system interprets that file is written in cp932 despite done in UTF-8. These users must set `encoding` option when use `read()` method in Python.

### DIFF
--- a/src/cfnlint/decode/cfn_yaml.py
+++ b/src/cfnlint/decode/cfn_yaml.py
@@ -233,7 +233,7 @@ def load(filename):
         for line in fileinput.input(files=filename):
             content = content + line
     else:
-        with open(filename) as fp:
+        with open(filename, encoding='utf-8') as fp:
             content = fp.read()
 
     return loads(content, filename)


src/cfnlint/decode/cfn_yaml.py
```diff
@@ -233,7 +233,7 @@ def load(filename):
         for line in fileinput.input(files=filename):
             content = content + line
     else:
-        with open(filename) as fp:
+        with open(filename, encoding='utf-8') as fp:
```
PatMyron, line 236: Should this be done in more places as well?
https://sourcegraph.com/search?q=context:global+r:aws-cloudformation/cfn-lint+with+open(+-file:test/|scripts/
If so, wonder if there's any way to set a package-level default rather than hardcoding everywhere

shunkt: > Should this be done in more places as well?

In testing, yes.

> If so, wonder if there's any way to set a package-level default rather than hardcoding everywhere

I found a enforcing utf-8 codec option (PEP 540).
https://peps.python.org/pep-0540/
But I'm not familiar with setuptools to combine with `-x utf8` option and it's valid after python 3.7.

kddejong:
Closing in favor of #2298.  Thanks @shunkt just added a new commit to cover the rest of the files in the main program.  I didn't include the test files.

PR #2299, opened 2022-07-07
Added transform parameter which outputs transformed template to a jso…

zanders2:
*Issue #, if available:*

*Description of changes:*

A new parameter was added --transform which causes cfn-lint to output the transformed template to a .json file in the same directory as the original template.

Created a new method in the Template class which creates a new file containing the transformed template.

The run_cli method was  updated to call the above mentioned method.


kddejong:
@zanders2 any reason why we couldn't just use `sam-translator.py` directly?  

https://github.com/aws/serverless-application-model/blob/develop/bin/sam-translate.py#L5-L18

kddejong:
Closing for now.  We can re-open as needed.

Issue #1930, opened 2021-03-10
Incorrect 'used rules' when using the pretty or junit formatters


jonathanmorley:
If you have a set of templates, that use a subset of rules, and pass them all into cfn-lint with `--format pretty` or `--format junit`, then only the last template's rules get used for evaluating the rules used for all the templates.

kddejong:
I believe this is fixed now.  We can re-open as needed.

PR #2330, opened 2022-08-18
test formatters and create used rules

kddejong:
*Issue #, if available:*
#1930 

*Description of changes:* This adds some tests for formatters that make use of the `rules` argument.

They are currently failing because I believe the current code to be incorrect. It only uses the rules from the last template that is checked.

Replacing PR #1929

### DIFF
--- a/src/cfnlint/__main__.py
+++ b/src/cfnlint/__main__.py
@@ -25,7 +25,7 @@ def main():
     try:
         (args, filenames, formatter) = cfnlint.core.get_args_filenames(sys.argv[1:])
         matches = list(cfnlint.core.get_matches(filenames, args))
-        (_, rules, _) = cfnlint.core.get_template_rules(filenames[-1], args)
+        rules = cfnlint.core.get_used_rules()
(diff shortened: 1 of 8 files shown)


src/cfnlint/__main__.py
```diff
@@ -25,7 +25,7 @@ def main():
     try:
         (args, filenames, formatter) = cfnlint.core.get_args_filenames(sys.argv[1:])
         matches = list(cfnlint.core.get_matches(filenames, args))
-        (_, rules, _) = cfnlint.core.get_template_rules(filenames[-1], args)
```
kddejong, line 28: only pulling the rules from the last run is causing the issue we are seeing

src/cfnlint/core.py
```diff
@@ -184,6 +184,14 @@ def get_args_filenames(cli_args):
     return(config, config.templates, formatter)
 
 
+def get_used_rules():
+    return __CACHED_RULES
+
+def _reset_rule_cache():
```
kddejong, line 190: I want to reset the cache so used_rules is reset

PR #2345, opened 2022-08-31
Configure dependabot

gliptak:
*Issue #, if available:*

*Description of changes:*


PatMyron:
Last time I looked into this:
https://github.com/aws-cloudformation/cfn-lint/pull/1917

Dependabot wouldn't do much since `cfn-lint` doesn't pin dependencies as a library trying to stay compatible with users' Python environments:
https://github.com/aws-cloudformation/cfn-lint/blob/main/setup.py#L44

Although hopefully over time we can start recommending separate Python environments for `cfn-lint` using something like `pipx`:
https://github.com/aws-cloudformation/cfn-lint/issues/1291#issuecomment-949047626

gliptak:
@PatMyron thank you for the ponters. proposed setup keeps GHA up-to-date

PatMyron:
Whoops, completely missed that line that this is for GitHub actions rather than Python, understand now 👍

kddejong (approved):

PR #2371, opened 2022-09-14
reduce JSON indent size

PatMyron:
https://github.com/aws-cloudformation/cfn-lint/issues/2368

`gsed -i 's/indent=./indent=1/' src/cfnlint/*.py src/cfnlint/formatters/__init__.py scripts/*.py`

---

Reducing indentation from 2 to 1 should remove ~1/4 of the file size while still preserving indentation formatting

EDIT: not as much size removed as I hoped, but still something


whoDoneItAgain:
is readability a requirement? removing linefeeds and indents would further reduce file size without sacrificing usability. I know if I need to look at a json that isn't pretty, I drop it in np++ and use the json plugin to prettify it.

PatMyron:
We do look at the diffs, so there is some use to preserving some formatting in the git repo at least. Hopefully this is just a quick stopgap solution and eventually we compress more under the hood

kddejong:
Agreed with both of you @whoDoneItAgain and @PatMyron .  I was debating dropping all readability but we have caught issues with the CloudFormation Specs that would be much harder to track down if it was completely reduced.

I'm not sure if there is another way Pat we could identify those possible issues but also open to some ideas

kddejong:
My quick tests:
141M    before
129M    after

Removing all spacing:
112M

src/cfnlint/formatters/__init__.py
```diff
@@ -167,7 +167,7 @@ def print_matches(self, matches, rules=None, filenames=None):
         del rules
 
         return json.dumps(
-            matches, indent=4, cls=self.CustomEncoder,
+            matches, indent=1, cls=self.CustomEncoder,
```
kddejong, line 170: Is this one necessary?  Seems unrelated to the issue

PatMyron: Agreed, not necessary and unrelated, just included it in the sweep for no strong reason 🤷 Feel free to keep or remove this one

kddejong, line 170:
```suggestion
            matches, indent=4, cls=self.CustomEncoder,
```

kddejong (changes requested):

kddejong (approved):

PR #2403, opened 2022-10-08
Switch from pylint to flake8

kddejong:
*Issue #, if available:*

*Description of changes:*
- Convert from pylint to flake8 to bring more 
- Run `black src/cflint` to auto-format files

No tests were modified to make sure consistency between previous to after changes.
Documentation for the rules was updated but that is the result of fixing a few spacing issues in the descriptions.


kddejong:
Black did most of the formatting changes.  I mostly had to add commas and split long strings

PR #2431, opened 2022-10-21
chore(specs): Limit Lambda::Function.Code/Properties/Zipfile to 4096 characters

### LINKED ISSUE #1870: New rule request: ZipFile length cannot exceed 4096 characters
*cfn-lint version: 0.38.0*

CloudFormation Lambda functions with inline code (through the `{"Code": {"ZipFile": "...."}}` property, or with SAM functions with the `InlineCode` property) have a size limit of 4096 bytes enforced by CloudFormation. It would be very helpful if cfn-lint could warn about crossing this threshold, since it's difficult to measure the code size in bytes in the text editor. Additionally, it would be helpful to not just report the failure, but also display the current size, to make it easier to know how far the code is over the limit.

Template to reproduce:

```yaml
Resources:
  Role:
    Type: AWS::IAM::Role
    Properties:
      ManagedPolicyArns: ["arn:aws:iam::aws:policy/service-role/AWSLambdaBasicExecutionRole"]
      AssumeRolePolicyDocument:
        Statement:
          - Action: sts:AssumeRole
            Effect: Allow
            Principal: {"Service": "lambda.amazonaws.com"}
  BigLambda:
    Type: AWS::Lambda::Function
    Properties:
      Role: !Ref Role
      Runtime: python3.8
      Code:
        ZipFile: |
          import json
          def handler(event, context):
            print(json.dumps(event))
            padding_data = "Call me Ishmael. Some years ago-never mind how long precisely-having little or no money in my purse, and nothing particular to interest me on shore, I thought I would sail about a little and see the watery part of the world. It is a way I have of driving off the spleen and regulating the circulation. Whenever I find myself growing grim about the mouth; whenever it is a damp, drizzly November in my soul; whenever I find myself involuntarily pausing before coffin warehouses, and bringing up the rear of every funeral I meet; and especially whenever my hypos get such an upper hand of me, that it requires a strong moral principle to prevent me from deliberately stepping into the street, and methodically knocking people's hats off-then, I account it high time to get to sea as soon as I can. This is my substitute for pistol and ball. With a philosophical flourish Cato throws himself upon his sword; I quietly take to the ship. There is nothing surprising in this. If they but knew it, almost all men in their degree, some time or other, cherish very nearly the same feelings towards the ocean with me.\nThere now is your insular city of the Manhattoes, belted round by wharves as Indian isles by coral reefs-commerce surrounds it with her surf. Right and left, the streets take you waterward. Its extreme downtown is the battery, where that noble mole is washed by waves, and cooled by breezes, which a few hours previous were out of sight of land. Look at the crowds of water-gazers there.\nCircumambulate the city of a dreamy Sabbath afternoon. Go from Corlears Hook to Coenties Slip, and from thence, by Whitehall, northward. What do you see?-Posted like silent sentinels all around the town, stand thousands upon thousands of mortal men fixed in ocean reveries. Some leaning against the spiles; some seated upon the pier-heads; some looking over the bulwarks of ships from China; some high aloft in the rigging, as if striving to get a still better seaward peep. But these are all landsmen; of week days pent up in lath and plaster-tied to counters, nailed to benches, clinched to desks. How then is this? Are the green fields gone? What do they here?\nBut look! here come more crowds, pacing straight for the water, and seemingly bound for a dive. Strange! Nothing will content them but the extremest limit of the land; loitering under the shady lee of yonder warehouses will not suffice. No. They must get just as nigh the water as they possibly can without falling in. And there they stand-miles of them-leagues. Inlanders all, they come from lanes and alleys, streets and avenues-north, east, south, and west. Yet here they all unite. Tell me, does the magnetic virtue of the needles of the compasses of all those ships attract them thither?\nOnce more. Say you are in the country; in some high land of lakes. Take almost any path you please, and ten to one it carries you down in a dale, and leaves you there by a pool in the stream. There is magic in it. Let the most absent-minded of men be plunged in his deepest reveries-stand that man on his legs, set his feet a-going, and he will infallibly lead you to water, if water there be in all that region. Should you ever be athirst in the great American desert, try this experiment, if your caravan happen to be supplied with a metaphysical professor. Yes, as every one knows, meditation and water are wedded for ever.\nBut here is an artist. He desires to paint you the dreamiest, shadiest, quietest, most enchanting bit of romantic landscape in all the valley of the Saco. What is the chief element he employs? There stand his trees, each with a hollow trunk, as if a hermit and a crucifix were within; and here sleeps his meadow, and there sleep his cattle; and up from yonder cottage goes a sleepy smoke. Deep into distant woodlands winds a mazy way, reaching to overlapping spurs of mountains bathed in their hill-side blue. But though the picture lies thus tranced, and though this pine-tree shakes down its sighs like leaves upon this shepherd's head, yet all were vain, unless the shepherd’s eye were fixed upon the magic stream"
            return {'moby': padding_data}
```  

Result of the template above:
* BigLambda CREATE_FAILED: ZipFile length cannot exceed 4096 characters. For larger source use S3Bucket/S3Key properties instead.

mmaeng:
Closes #1870 

Added maximum size of 4096 characters to Lambda::Function.Code/Properties/Zipfile 


kddejong (approved):

bgardner-noggin:
This is actually now causing a false positive.

Approximately 3 months ago, AWS silently changed the behaviour of this. Now so long as the resulting zip file does not exceed 4MB, Cloudformation will accept it

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-lambda-function-code.html

bgardner-noggin:
I just saw the commit which still looks incorrect. The limit applied by cloudformation is to the resulting zip file, not the uncompressed raw file.

kddejong:
Got it.  Good catch again.  I'm going to probably just remove the constraint all together.  I want to test a few things tonight and see if I understand the different impacts here then I'll push out a fix. 

@bgardner-noggin have you put in a lambda of that size before?  How is this impacted by the max size of a template?

bgardner-noggin:
No I've never gotten close to a lambda of that size compressed. Our max would be somewhere around 10k uncompressed.

My two cents is that 4mb compressed is whoppingly huge and that cfn-lint shouldn't be checking at all, as it probably can't tell what the final compressed size will be accurately.

kddejong:
When testing this on a large lambda (~4MB) I get `Template may not exceed 1000000 bytes in size.` which is the template size limit (1MB from S3).  My guess here is that we can safely remove this check as you will be limited by the size of the template before you will be impacted by this limit.  Which would make this check useless.

Issue #2219, opened 2022-02-22
Tag values are not validated against permitted character regex


trav-c:
*cfn-lint version: (`cfn-lint --version`)*
cfn-lint 0.58.1

*Description of issue.*
Tag values are not validated against the permissible regex ^([\p{L}\p{Z}\p{N}_.:/=+\-@]*)$ which is documented here https://docs.aws.amazon.com/directoryservice/latest/devguide/API_Tag.html specifically in this test case tag values containing an & character, which is not valid are not detected on either AWS::SSM:Parameter or AWS::S3::Bucket (or I suspect anywhere)

Please provide as much information as possible:
As an example, no errors are detected in the following template by cfn-lint, but attempting to launch the template yields a validation error (see below)

Sample Template
```
Description: "Test CloudFormation Template"
Resources:
    Param:
        Type: AWS::SSM::Parameter
        Properties:
            Type: String
            Value: SomeValue
            Tags:
              Test: 'A & B'
    
    Bucket:
        Type: AWS::S3::Bucket
        Properties:
            Tags:
              - Key: Test
                Value: 'A & B'
```

Stack launch error:
```
1 validation error detected: Value 'A & B' at 'tags.1.member.value' failed to satisfy constraint: Member must satisfy regular expression pattern: ^([\p{L}\p{Z}\p{N}_.:/=+\-@]*)$. (Service: AmazonSSM; Status Code: 400; Error Code: ValidationException; Request ID: e428b41c-4952-439d-87c4-f30a597990b9; Proxy: null)
```

PatMyron:
https://github.com/aws-cloudformation/cfn-lint/issues/903#issuecomment-595356853, https://github.com/aws-cloudformation/cfn-lint/pull/1867#discussion_r556178172
Probably won't be able to use that pattern as-is due to [unicode categories unsupported by Python `re`](https://stackoverflow.com/questions/1832893/python-regex-matching-unicode-properties), but might be able to at least construct a regex pattern disallowing certain characters to catch issues like that

trav-c:
@PatMyron is the [regex](https://pypi.org/project/regex/) (as opposed to re) module an option here?
It appears to support \p and at least in a cursory test appears to work correctly with the 'official' regex

PatMyron:
[...]

PR #2437, opened 2022-10-24
Convert from re to regex

kddejong:
*Issue #, if available:*
#2219 
*Description of changes:*
- Replace re with regex allowing us to expand the regex expressions we support


kddejong:
Across all of our tests I've noted ~2 second slowdown (on my mac)

kddejong:
I don't think this is worth it so I'm going to close and we will continue to re-evaluate.

PR #2497, opened 2022-11-22
Add option to filter for Cloudformation templates

stampflit:
*Issue #, if available:*

*Description of changes:*

Adding support for the `--detect-cfn-lint` argument.  This only lints files which contain the string `AWSTemplateFormatVersion`.

After some internal tests, this check neither excludes any files that should be checked nor does it usually check any file that are intended to be left out.  But if requested I can implement a more specific detection mechanism.

I didn't add any tests for this code since I'm not sure what tests would be expected.  I can add some tests if needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,6 +179,7 @@ Optional parameters:
 | -l, --list-rules | | | List all the rules |
 | -r, --regions | regions | [REGIONS [REGIONS ...]], ALL_REGIONS  | Test the template against many regions.  [Supported regions](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html) |
 | -b, --ignore-bad-template | ignore_bad_template | | Ignores bad template errors |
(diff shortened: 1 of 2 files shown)


kddejong:
Ideally we would go beyond just `AWSTemplateFormatVersion` since that isn't a required property.  `Resources` is the only required property so ideally we would look for `Resources` and then a property of `Type` with a common naming standard including the `Publisher::Service::Resource` standard.

kddejong:
Via different issues and topics this has been requested a few times so is probably a good solution.

Also we implement similar logic into the VS Code plugin

stampflit:
Should I change the check to use the same Regexps as here: https://github.com/aws-cloudformation/cfn-lint-visual-studio-code/blob/e5f98b7fbdbba9ef502b1fa2f0f2c14383116123/server/src/server.ts#L134

Newer version actually look at the parsed yaml.  I don't know how well this integrates with the current architecture of cfn-lint: https://github.com/aws-cloudformation/cfn-lint-visual-studio-code/blob/main/server/src/utils/cfnParser.ts#L115

Issue #1756, opened 2020-10-27
New rule to test parameter attributes based on NumberMax, NumberMin values


kddejong:
Create a new rule to use the NumberMax and NumberMin value attributes to determine if a Parameters Default, AllowedValues, MinValue, and MaxValue are within the limits specified.

kddejong:
MinValue, and MaxValue still need to be done.

kddejong:
This is complete in v1

PR #2521, opened 2022-12-14
Add support for AllowedValues/AllowedPattern for CommaDelimitedLists

MalikAtalla-AWS:
This change adds support for AllowedValues and AllowedPattern for CommaDelimitedLists by warning the user when the default value is invalid according to the same rules enforced on the server-side. Both these contraints are now applied to every item of a CommaDelimitedList parameter.


kddejong (approved):

kddejong:
@MalikAtalla-AWS I think this resolves #1756  correct?

PR #2558, opened 2023-01-18
Fix PreviousGenerationInstanceType family string matching

sp-ricard-valverde:
*Description of changes:*

`PreviousGenerationInstanceType` rule is currently using a validation regex that is incorrectly targeting instance types like `mac1.metal` due to `mac1.` being matched because of an incomplete regex approach.

The matching mechanism has been improved to allow for every case of `InstanceType` and similar properties.


src/cfnlint/rules/resources/PreviousGenerationInstanceType.py
```diff
@@ -91,3 +91,14 @@ def match(self, cfn):
                             )
                         )
         return matches
+
+    def __is_previous_generation_instance_type(self, property_type, instance_type):
+        sections = instance_type.split(".")
+        if len(sections) == 1:
+            return False
+
+        family_section = sections[0] if property_type == "InstanceType" else sections[1]
```
PatMyron, line 100: Good catch on `mac1.` This new logic should work for the services currently covered by this rule, but reading different sections based on if the property type is named `InstanceType` might not work if certain other services are ever added to this rule like [`AWS::Redshift::Cluster.NodeType`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-redshift-cluster.html#cfn-redshift-cluster-nodetype)

Something like just prefixing the pre-existing regex with `(^|\.)` to make sure it's at the start of the value or after `.` might be more general

src/cfnlint/rules/resources/PreviousGenerationInstanceType.py
```diff
@@ -91,3 +89,11 @@ def match(self, cfn):
                             )
                         )
         return matches
+
+    def __is_previous_generation_instance_type(self, instance_type):
+        return (
+            re.search(
+                r"(^|\.)[cmr][1-3]|cc2|cg1|cr1|g2|hi1|hs1|i2|t1($|\.)", instance_type
```
PatMyron, line 96:
```suggestion
                r"(^|\.)([cmr][1-3]|cc2|cg1|cr1|g2|hi1|hs1|i2|t1)($|\.)", instance_type
```

kddejong (approved):

PR #2577, opened 2023-02-10
Added ItemProcessor State Machine Definition key for State of Type (Map)

jayasimha537:
As per https://docs.aws.amazon.com/step-functions/latest/dg/concepts-asl-use-map-state-inline.html ,  "ItemProcessor" State Machine Definition key for State of Type (Map)  is valid.

As of now, getting below error : 
E2532 State Machine Definition key (ItemProcessor) for State (Map Api Calls) of Type (Map) is not valid


src/cfnlint/rules/resources/stepfunctions/StateMachine.py
```diff
@@ -60,6 +60,7 @@ def _check_state_json(self, def_json, state_name, path):
                 "MaxConcurrency",
                 "Iterator",
                 "ItemsPath",
+                "ItemProcessor",
```
kddejong, line 63:
```suggestion
                "ItemProcessor",
                "ItemSelector",
```

kddejong (changes requested):

src/cfnlint/rules/resources/stepfunctions/StateMachine.py
kddejong: I think we are also missing `ItemSelector`

kddejong (approved):

Issue #2612, opened 2023-03-08
Custom rule greater and lesser operators are not numerical


kftsehk:
https://github.com/aws-cloudformation/cfn-lint/blob/55752dc40ee5935704c48ad7422dc98e1e71c88f/src/cfnlint/rules/custom/Operators.py#L132-L191

These lines are not comparing the Greater or Lesser operator numerically, and also not taken care of int input in yaml/json

Custom rule
```
AWS::Lambda::Function Timeout >= "5" ERROR "Have you forgot to set Timeout? 5s ain't doing anything."
```

# Case 1:

```yaml
Resources:
  TestNumericalPropertyLambda:
    Properties:
      PackageType: Zip
      Runtime: nodejs16.x
      Timeout: 100
    Type: AWS::Lambda::Function
```

fails for YAML numerical type due to error

```
E0002 Unknown exception while processing rule E9001: 'int' object has no attribute 'strip'
test.yml:1:1
```

# Case 2

```yaml
AWSTemplateFormatVersion: '2010-09-09'
Resources:
  TestNumericalPropertyLambda:
    Properties:
      PackageType: Zip
      Runtime: nodejs16.x
      Timeout: '100'
    Type: AWS::Lambda::Function
```

fails due to '100' < '5' of string comparison

```
E9001 Have you forgot to set Timeout? 5s ain't doing anything.
test.yml:5:7
```

PR #2614, opened 2023-03-09
Fix custom rule inequality comparison (#2612)

kftsehk:
*Issue #, if available:*
#2612 

*Description of changes:*
- Fix corresponding operators by type conversions
- Add 1 good 1 bad test case for custom rule Greater / Lesser operator regarding
  - numerical comparison
  - number presented in python `int` type

### DIFF
--- a/test/unit/rules/custom/test_greater_operator.py
+++ b/test/unit/rules/custom/test_greater_operator.py
@@ -4,7 +4,7 @@ SPDX-License-Identifier: MIT-0
 """
 from test.unit.rules import BaseRuleTestCase
 
-from cfnlint.rules.custom.Operators import CreateGreaterRule # pylint: disable=E0401
+from cfnlint.rules.custom.Operators import CreateGreaterRule  # pylint: disable=E0401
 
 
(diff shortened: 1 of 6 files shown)


kddejong (approved):

test/unit/rules/custom/test_greater_operator.py
```diff
@@ -0,0 +1,26 @@
+"""
```
kddejong, line 1: Can we move some of these tests into https://github.com/aws-cloudformation/cfn-lint/blob/main/test/unit/module/custom_rules/test_custom_rules.py

kftsehk: Can but not sure if it is a good idea tho, 
https://github.com/aws-cloudformation/cfn-lint/blob/main/test/unit/module/custom_rules/test_custom_rules.py
mostly deal with the rule parser.

While the added test cases are end-to-end assertions based on some template fixtures, which I find similar with rule tests in the same level, e.g.
https://github.com/aws-cloudformation/cfn-lint/blob/main/test/unit/rules/conditions/test_and.py

PR #2624, opened 2023-03-13
Convert conditions to SymPy

kddejong:
*Issue #, if available:*

*Description of changes:*
- Convert conditions logic to use [SymPy](https://www.sympy.org/en/index.html) to determine scenarios and relationship of conditions

---
Currently resolving conditions is bottom up.  Meaning if we are determining the relationship of two conditions we look at equals and how they overlap.  We then create a slew of scenarios based on the different equal scenarios.  The more equals the more scenarios, the more unrelated equals the more scenarios.   If you have 6 equals in a set of 2 conditions that are unrelated you have 2^6 = 64 scenarios.  In the end from all those scenarios you really only have 4 max possibilities.

This new approach works the other way.  It uses SymPy to validate if the 4 scenarios are satisfied.  To relate equals that have 1 similar parameter (but not 2) we use `Not(And(...))` which will not allow each equals to be `True` at the same time.

---

Possible remaining issues:

- [x] A condition that is always `True`/`False`.  Sometimes a person will codify `!Equals ["True", "True"]`
- [x] A condition based on values limited by AllowedValues parameter may never create a scenario.  Environments parameter may have `AllowedValues: ["prod", "dev"]` which means condition `IsProduction` and `IsDevelopment` can never both be `False`


ericzbeard (approved):

kddejong:
This will cause issues for installation on Alpine

PR #2646, opened 2023-03-26
Fix SAM templates treated as normal by api

LaurenceWarne:
*Issue #, if available:*
N/A

*Description of changes:*

Hi, this PR fixes an issue where SAM templates were treated as normal Cloudformation templates by the api, by calling `.transform` on the runner.


test/unit/module/test_api.py
```diff
@@ -66,3 +66,8 @@ def test_issues_template(self):
         filename = "test/fixtures/templates/bad/issues.yaml"
         matches = self.helper_lint_string_from_file(filename)
         self.assertEqual(["E1012"], [match.rule.id for match in matches])
+
+    def test_sam_template(self):
```
LaurenceWarne, line 70: This would fail prior to this change since the globals section is invalid in normal Cloudformation templates.

Issue #2296, opened 2022-07-05
[feature request] Validate Cloudfront::Distribution.DistributionConfig.DefaultCacheBehavior.TargetOriginId exists


alex-a-pereira:
*cfn-lint version: v0.60.1*

*Description of issue.*

The linter seems like it should validate that `TargetOriginId` must be set to an origin that exists in `DistributionConfig.Origins` array (please let me know and close if I'm incorrect). We're using AWS SAM to deploy the cloudformation template.

Currently this error is only caught (with a generic error message) at the time that cloudformation is deployed via `sam deploy`. 
It would be great if this could be caught in static analysis - especially considering how updating cloudfront distributions via cloudformation is slow.

Example:
```yml
Resources:
  ReverseProxyCloudfrontDistribution:
    Type: 'AWS::CloudFront::Distribution'
    DependsOn:
      - ReactProjectCloudfrontDistribution
    Properties:
      DistributionConfig:
        Enabled: true
        Comment: 'My Website Reverse Proxy'
        HttpVersion: http2
        Origins:
          - Id: proxy-to-react-project
            DomainName: !GetAtt ReactProjectCloudfrontDistribution.DomainName
            CustomOriginConfig:
              OriginProtocolPolicy: "https-only"
          - Id: proxy-to-backend-orm
            DomainName: 'staging.mywebsite.com'
            CustomOriginConfig:
              OriginProtocolPolicy: "https-only"
        DefaultCacheBehavior:
          Compress: 'true'
          AllowedMethods:
            - GET
            - HEAD
            - OPTIONS
          ForwardedValues:
            QueryString: true
          # 
          # !!!!!!!!!! SEE HERE !!!!!!!!!!
          # I made a (rather obvious) typo to show that the linter does not catch this
          # it should be set to `proxy-to-react-project`
          #
          TargetOriginId: proxy-to-REACTTTTTTT-project
          ViewerProtocolPolicy : allow-all

  ReactProjectCloudfrontDistribution:
    Type: 'AWS::CloudFront::Distribution'
    Properties:
      # omitted for this example 
      # not relevent to the issue being discussed
```

Error message when deploying:
```
"Invalid request
provided: One or more of
your origins or origin
groups do not exist.
(Service: CloudFront, Status
Code: 404, Request ID: 47dd8
910-c4e0-4c2c-9b55-6df586af6
56e)" (RequestToken: 57323dc
4-fb71-c4be-c472-e738ec6d3de
[...]

Issue #2619, opened 2023-03-10
Checking AWS::StepFunctions::StateMachine.Definition


kftsehk:
### Is this feature request related to a new rule or cfn-lint capabilities?

New capability

### Describe the feature you'd like to request

Validate the state machine definition against the spec, if it is written in json / yaml form.
Understood that if a string form is used it probably cannot be validated.

### Describe the solution you'd like

Example:
```
Resources:
  SampleWorkflow:
    Type: AWS::StepFunctions::StateMachine
    Properties:
      Definition:
        StartAt: 'SampleChoice'
        States:
          SampleChoice:
            Default: __SucceedEntryPoint__
            Type: Choice
          __SucceedEntryPoint__:
            Type: Pass
```

A cfn-lint spec like static check should already show the above has 2 errors:

```
Definition of "SampleChoice" should include at least one "Choices"
Definition of "__SucceedEntryPoint__" should have either "Next" or "End"
```

See if a spec for the Definition can be retrieved from your side. Let's not discuss the part where validating the reachability of the workflow at this time, that would probably be breaking change.

### Is this something that you'd be interested in working on?

- [X] 👋 I may be able to implement this feature request

### Would this feature include a breaking change?

- [ ] ⚠️ This feature might incur a breaking change

PR #2672, opened 2023-04-08
Generic Match/Project/Validate pattern rule

kftsehk:
*Issue #, if available:*

Targeting the implementation of cross-checking, such as

#2296 #1059 #1439

*Description of changes:*

This is an attempt to make v0.x linter more capable of cross-validation that depends on multiple fields
- A generic Match/Project/Validate pattern rule
- YAML declared Match/Project/Validate parameters
- Implement #2296
  - [ ] Assign rule ID (temporarily E2554)
  - [ ] Update changelog
- Implement #1059
  - [ ] Assign rule ID (temporarily E2555 / E2556)
  - [ ] Update changelog
- Implement #1439
  - [ ] Assign rule ID (temporarily E2553)
  - [ ] Update changelog

Some thoughts not implemented
- And/Or operator to Match/Validate is useful
- Good to have a way to show more useful message for different validation that failed, maybe with each `Validations` item should accept a string template with one `%s`
- Sometimes might want to group multiple rules under same error ID (E2555 / E2556 are quite similar in this case)
- Custom function `fn = (checked_value) => bool` needed, met a case need to check `10 | 30 | divisible by 60` cannot be enforced even with regex

### DIFF
--- a/docs/custom_rules.md
+++ b/docs/custom_rules.md
@@ -41,9 +41,12 @@ The specified operator to be used for this rule. The supported values are define
 | == | Identical to `EQUALS` |
 | NOT_EQUALS | Checks the specified property is not equal to the value given |
 | != | Identical to `NOT_EQUALS` |
+| REGEX_MATCH | Checks the specified property matches regex given (using python `regex` module) |
 | IN | Checks the specified property is equal to or contained by the array value |
 | NOT_IN | Checks the specified property is not equal to or not contained by the array value |
 | \>= | Checks the specified property is greater than or equal to the value given |
+| \> | Checks the specified property is greater than the value given |
+| < | Checks the specified property is less than the value given |
 | <= | Checks the specified property is less than or equal to the value given |
 | IS | Checks the specified property is defined or not defined, the value must be one of DEFINED or NOT_DEFINED |
 
--- a/src/cfnlint/rules/custom/Operators.py
+++ b/src/cfnlint/rules/custom/Operators.py
@@ -452,7 +452,9 @@ def CreateLesserEqualRule(rule_id, resourceType, prop, value, error_message):
 def CreateInSetRule(rule_id, resourceType, prop, value, error_message):
     def rule_func(value, expected_values, path):
         matches = []
-        if value not in expected_values:
+        if isinstance(expected_values, list):
+            expected_values = [str(x) for x in expected_values]
+        if str(value) not in expected_values:
             matches.append(
                 cfnlint.rules.RuleMatch(path, error_message or "In set check failed")
             )
@@ -474,7 +476,9 @@ def CreateInSetRule(rule_id, resourceType, prop, value, error_message):
 def CreateNotInSetRule(rule_id, resourceType, prop, value, error_message):
     def rule_func(value, expected_values, path):
         matches = []
-        if value in expected_values:
+        if isinstance(expected_values, list):
+            expected_values = [str(x) for x in expected_values]
+        if str(value) in expected_values:
             matches.append(
                 cfnlint.rules.RuleMatch(
                     path, error_message or "Not in set check failed"
--- a/src/cfnlint/rules/custom/Operators.py
+++ b/src/cfnlint/rules/custom/Operators.py
@@ -3,16 +3,21 @@ Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
 SPDX-License-Identifier: MIT-0
 """
 
+import regex as re
+
 # pylint: disable=cyclic-import
 import cfnlint.rules
 
 OPERATOR = [
     "EQUALS",
     "NOT_EQUALS",
+    "REGEX_MATCH",
     "==",
     "!=",
     "IN",
     "NOT_IN",
+    ">",
+    "<",
     ">=",
     "<=",
     "IS DEFINED",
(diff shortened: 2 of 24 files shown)


kddejong:
I think some of this makes sense.  As we move to v1 and the registry provider schemas we are doing a lot more json schema validation as well.  In that approach we are actually building template scenarios that remove conditions and focus on the actual values.  We also modified how json schema validation works to handle the other intrinsic functions.

Why I say all of this is we can also write rules by extending the schema validation and writing json schemas.
https://github.com/aws-cloudformation/cfn-lint/pull/2685/commits/fefa8240e6b22beb31b9e6cc3d5c461581107768

I don't think all the linked scenarios will work with this approach.  I haven't written a mechanism for people to write their own json schemas at this point and this will not work across resources.

kftsehk:
Surely will look into v1 preview. It is good to hear that these might be included in v1.

I much agree the v1 change that it is way better to focus on
- (A) possible scenarios generated by conditions
- (B) static value inference: `!Ref AWS:NoValue (set it back to undefined!)`, `!FindInMap`, `!Ref Parameter`, `!Ref AWS::PseudoParameters`
- (C) static type inference: `!Ref X`, `!GetAtt X.Att`
before rules are enforced.

We have used cfn for quite some years and definitely met some workarounds, issues that we want be able to configure the linters to check, and ended up arbitrary script that just load the yaml and check that.
Some use case and preliminary thoughts from experience using and contributing to cfn-lint v0.7 below in case it help

Please take it lightly in case it had been considered, since I have not dive into details of v1.

### (A) Handling condition

My main concern is with linter performance if repeatedly evaluating parameter combinations on the whole template, which can be a pretty large if number of conditions are more than handful.

**How condition can affect a template**

*1. By adding or removing resources*
*2. By choosing among different property values within a resources*

It maybe possible to compute and aggregate `<ConditionConsumed, [ResourceConsumedByRefOrGetAtt]>` for each resources, and check against `<Scenario, AvailableResource>`, so 1. is solved, and 2. can be done locally only considering possible values of that particular condition.

### (B) Static value inference

The current API seem to leave static value inference to rule implementation by `check_find_in_map` etc, it is definitely better to handle centrally by having concrete value at the rule

In this issue I've specifically *added "Projection"* for these use case I can think of
- #2296, the list of origin id are arbitrarily named by user
- #2619 Where `StartAt` `Next` etc must be one of the states which we name it arbitrarily

### (C) Static type inference

cfn-lint v0.7 will miss on mismatched !Ref / !GetAtt, for example, `!Ref LogicalName` instead of `!GetAtt LogicalName.Arn` when an arn is actually required and `!Ref` returns only the logical name, prominently in IAM policy, role, SNS subscription etc.

The susceptibility to human error, for this Arn case, is probably due to the fact that it is quite arbitrary for each resource type that

*Does !Ref return an ARN or logical name?*: my guess is majority is for logical name, and substantial minority returns arn
  - Arn in !Ref https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-secretsmanager-secret.html#aws-resource-secretsmanager-secret-return-values
  - Logical name in !Ref https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-iam-role.html#aws-resource-iam-role-return-values

_Does !GetAtt X.Arn is available_: most resources has, but one can always find exception
  - No Arn attr.: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-secretsmanager-secret.html#aws-resource-secretsmanager-secret-return-values

_When arn is required by some field not explicitly naming someArn_
  - AWS::Serverless::Function.Role: Role is (almost?) always an Arn, although it is not named RoleArn as in `AWS::Lambda::Function`
  - AWS::SNS::Subscription.Endpoint: Is either email, url or arn, never a logical name
  - AWS::ECS::TaskDefinition.ExecutionRoleArn/TaskRoleArn: Surprisingly, I am able to find one instance, where it is named Arn, but you can actually use `!Ref` https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-taskdefinition.html#cfn-ecs-taskdefinition-taskrolearn

Generally the returns will fall into several types, being one of `url`, `ip`, `hostedZoneId`, `cidr`, `arn`, `string` (usually logical names) etc. I am not sure if return values are available in cfn spec, if so, return value can then be checked against the requirement of that field to catch the above !Ref/GetAtt confusion

Some additional use cases

### (D: suggestion) a stage for simple string substitution to dynamic references from downloaded values: `!ImportValue`,  `{{resolve::ssm::*}}`, `!Sub {{resolve:ssm:*}}`

I guess there is no solution for a linter to check against external value, where everyone is happy and still lightweighted. I would suggest allowing to provide a string substitution file so we can pull samples from dev/uat to check against, there is only limited ability with string substitution, but it can be better than nothing.

### (E: suggestion) includes non-`Properties` into check, `Metadata`, `DeletionPolicy` those sometimes matters

Met this when dealing with stack deployed with changeset, end up having to put arbitrary metadata around
- https://github.com/aws-cloudformation/cloudformation-coverage-roadmap/issues/844

It is also useful to enforce some deletion policy config on IaC source code level.

kddejong:
@kftsehk merge conflict on this one. Do we still need this PR?

kftsehk:
@kddejong any rule to assign rule id? I think I only miss that here. The conflict can be solved.

kddejong:
Need to rework this to handle it in v1

PR #2682, opened 2023-04-11
New rule for Lambda zip deployment

### LINKED ISSUE #2676: New rule for package command and Lambda

_No response_

Error message when using the package command and Lambda: `Runtime and Handler are mandatory parameters for functions created with deployment packages. `

A new warning or information rule that determines if you are using the package command and alerts that these parameters are required.

### Additional context

_No response_

- [X] 👋 I may be able to implement this feature request

tal66:
(closes #2676)

Hi, implemented this new rule. what do you think?

*quick failure test:
`cfn-lint test/fixtures/templates/bad/resources/lambda/zipfile_required_properties.yaml`


kftsehk:
@tal66 When implementing 
- #2656
I did missed a whole chunk of those !Ref !GetAtt ... stuff, and fixed in
- #2673

You can check the implementation of #2673, isn't simple to compare these cases including correctly check `!Ref AWS::NoValue` as undefined

@tal66 @kddejong What do you think of this proposal
- https://github.com/aws-cloudformation/cfn-lint/pull/2672

Would generally solve the problem of implementing rules enforcing `if A assert B is <op> C`, which can be less prone to problems of correctly re-implementing checking to !Ref / !GetAtt in each rule.

tal66:
Do you mean templates such as this (*)?

is there already a way that you use to resolve possible scenarios and attributes?  
or, since the [Code](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-lambda-function-code.html)(**) property is always required here, maybe it's enough just to see if by removing `ImageUri` (sub)property there's still something left under `Code`, which means zip file scenario (and then search for `Handler` and `Runtime`)

(*)
```yaml
Description: Properties with conditions.

Parameters:
  UseZipDeployment:
    Type: String
    Default: true
    AllowedValues:
      - true
      - false
Conditions:
  IsZipDeployment:
    !Equals [true, !Ref UseZipDeployment]

Resources:
  Function1:
    Type: AWS::Lambda::Function
    Properties:
      Role: arn:aws:iam::123456789012:role/lambda-role
      Code:
        !If
        - IsZipDeployment
        - S3Bucket: my-bucket
          S3Key: my-lambda-function.zip
        - ImageUri: 111122223333.dkr.ecr.us-east-1.amazonaws.com/app:latest
      PackageType: !If [IsZipDeployment, "Zip", "Image"]
```

(**) 
```yaml
# Code:
  ImageUri: String
  S3Bucket: String
  S3Key: String
  S3ObjectVersion: String
  ZipFile: String
```

kftsehk:
afaik, `Code` (or `CodeURI` in `Serverless::Function`)
can be 
- String, e.g. `"./my-lambda-function"`, where the artifacts are automatically uploaded and substituted into S3 url when using `sam deploy`, requires `Handler` and `PackageType: Zip`
- S3, requires `Handler` and `PackageType: Zip`
- ImageUri, `PackageType: Image`

kftsehk:
Not sure if everyone agree with this, what do you think?
As
- cfn-lint is best effort basis
- there is probably not a good reason one would maintain 2 different deployment of the same lambda function (one zip, one image)

we can opt to implement only simple cases of the following, ignoring the very complicated one

```yaml
Conditions:
  IsZipDeployment: !Equals [true, !Ref 'UseZipDeployment']
Description: Properties with conditions.
Parameters:
  MyImage:
    Type: String
  MyImageVersion:
    Type: String
  UseZipDeployment:
    AllowedValues:
      - true
      - false
    Default: true
    Type: String
Resources:
  Bucket:
    Properties:
      BucketName: my-bucket
    Type: AWS::S3::Bucket
  SimpleLambdaImage:
    Properties:
      Code:
        ImageUri: !Ref 'MyImage'
      PackageType: Image
      Role: arn:aws:iam::123456789012:role/lambda-role
    Type: AWS::Lambda::Function
  SimpleLambdaImageWithVersion:
    Properties:
      Code:
        ImageUri: !Sub
          - '${S}:${V}'
          - S: !Ref 'MyImage'
            V: !Ref 'MyImageVersion'
      PackageType: Image
      Role: arn:aws:iam::123456789012:role/lambda-role
    Type: AWS::Lambda::Function
  SimpleLambdaInplaceSource:
    Properties:
      Code:
        ZipFile: !Sub
          - |
            def handler(event, context):
              return "Hello World ${Region}"
          - Region: !Ref 'AWS::Region'
      Handler: index.handler
      PackageType: Zip
      Role: arn:aws:iam::123456789012:role/lambda-role
    Type: AWS::Lambda::Function
  SimpleLambdaZipLocalSource:
    Properties:
      Code: ./src
      Handler: !Join
        - '.'
        - - index
          - !Ref 'AWS::Region'
      PackageType: Zip
      Role: arn:aws:iam::123456789012:role/lambda-role
    Type: AWS::Lambda::Function
  SimpleLambdaZipS3:
    Properties:
      Code:
        S3Bucket: !Ref 'Bucket'
        S3Key: !Sub
          - 'my-function-${R}'
          - R: !Ref 'AWS::Region'
      Handler: index.handler
      PackageType: Zip
      Role: arn:aws:iam::123456789012:role/lambda-role
    Type: AWS::Lambda::Function
  VeryComplicatedConditionUnlikelyToBeUsed:
    Properties:
      Code: !If
        - IsZipDeployment
        - S3Bucket: !Ref 'Bucket'
          S3Key: !Sub
            - "my-function-${R}"
            - R: !Ref 'AWS::Region'
        - ImageUri: 111122223333.dkr.ecr.us-east-1.amazonaws.com/app:latest
      Handler: !If
        - IsZipDeployment
        - !Sub "my.${AWS::Region}.handler}"
        - !Ref 'AWS::NoValue'
      PackageType: !If
        - IsZipDeployment
        - "Zip"
        - "Image"
      Role: arn:aws:iam::123456789012:role/lambda-role
    Type: AWS::Lambda::Function

```

src/cfnlint/rules/resources/lmbd/ZipPackageRequiredProperties.py
```diff
@@ -0,0 +1,64 @@
...
+            "Handler",
+            "Runtime",
+        ]  # required if package is a .zip file
+
+        resources = cfn.get_resources(
+            ["AWS::Lambda::Function", "AWS::Serverless::Function"]
+        )
+        for resource_name, resource in resources.items():
+            properties = resource.get("Properties")
+            if not properties:
```
kddejong, line 34:
```suggestion
            if not isinstance(properties, dict):
```

src/cfnlint/rules/resources/lmbd/ZipPackageRequiredProperties.py
```diff
@@ -0,0 +1,64 @@
...
+                continue
+
+            path = ["Resources", resource_name, "Properties"]
+
```
kddejong, line 38:
```suggestion
for scenario in cfn.get_object_without_conditions(properties, ["PackageType", "Code", "Handler", "Runtime"]):
                properties = scenario.get('Object')
```

src/cfnlint/rules/resources/lmbd/ZipPackageRequiredProperties.py
```diff
@@ -0,0 +1,64 @@
...
+            is_zip_deployment = True
+            if properties.get("PackageType") == "Zip":
+                path.append("PackageType")
+            elif properties.get("Code", {}).get("ZipFile") or properties.get(
+                "Code", {}
+            ).get("S3Key", "").endswith(".zip"):
```
kddejong, line 44: Do we need the `.endswith(".zip")`?  Assuming that S3Key may be a string is probably going to cause issues.  If its specified doesn't it automatically mean its a zip?

src/cfnlint/rules/resources/lmbd/ZipPackageRequiredProperties.py
```diff
@@ -0,0 +1,64 @@
...
+            properties = resource.get("Properties")
+            if not properties:
+                continue
+
+            path = ["Resources", resource_name, "Properties"]
+
+            is_zip_deployment = True
+            if properties.get("PackageType") == "Zip":
+                path.append("PackageType")
+            elif properties.get("Code", {}).get("ZipFile") or properties.get(
```
kddejong, line 42: We should probably store code off into a parameter.  We need to make sure its a dict before treating it as a dict.  Other rules will validate the correct time so in this rule we just don't want a false positive for bad formatting.  
```
code = properties.get("Code")
if isinstance(code, dict):
```

kddejong (changes requested):

kddejong (approved):

kddejong:
Can you run and submit the updates?
```
black src/ test/
isort --profile black src/cfnlint test/
```

Issue #2797, opened 2023-07-17
physical resource name limit


blade2005:
### Describe the feature you'd like to request

I want to be able to determine before launching CFN if any resources have names which exceeds the service limitations.

Lambda functions limit is 64 characters
buckets are limited to 64 characters
glue jobs are 255 characters
event bridge rules are 64 characters.

If I've use CDK to dynamically generate the names (so they are static in the outputted cfn) the names may exceed the limitations of the individual service

### Describe the solution you'd like

an error generated if any resources with static names(no sub or ref support) has a name which exceeds the max length supported by those resources.

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request

### Would this feature include a breaking change?

- [ ] ⚠️ This feature might incur a breaking change

ashemedai:
@kddejong Did a quick check of the code base, but isn't this already (largely) implemented via src/cfnlint/rules/resources/properties/StringSize.py that fetches `StringMin` and `StringMax` (E3033)?

If I create a template of an `AWS::S3::Bucket` with a `BucketName` longer than 63 characters I get this:
```sh
$ cfn-lint s3-bucket.yml
E3033 String has to have length between 3 and 63 at Resources/S3Bucket/Properties/BucketName. Current length 71.
```

kddejong:
@ashemedai I think you're right

- [X] Lambda functions limit is 64 characters
- [X] buckets are limited to 64 characters
- [ ] glue jobs are 255 characters
- [ ] event bridge rules are 64 characters.

Shouldn't be hard to add the other two as we have the process to do this.

kddejong:
This PR should take care of the other two

ashemedai:
@kddejong Fastest would be to get a proper entry to src/cfnlint/data/CloudSpecs/us-east-1.json e.g.

```json
  "AWS::Glue::Job.Name": {
  "AllowedPatternRegex": "^[\u0020-\uD7FF\uE000-\uFFFD\uD800\uDC00-\uDBFF\uDFFF\t]*$",
  "StringMax": 255,
  "StringMin": 1
  },
  "AWS::Events::Rule.Name": {
  "AllowedPatternRegex": "^[\.\-_A-Za-z0-9]+$",
  "StringMax": 64,
  "StringMin": 1
  },
```

[...]

PR #2820, opened 2023-07-27
Add various resource constraints for StringMax/StringMin

ashemedai:
*Description of changes:*

Based on the exchange in #2797 with @kddejong I whipped up a small script to scan the various resources for `Name` and `Description` and see if they were missing their constraints. This is the result of that.

For `Description` if an explicit maximum was given, but no minimum I reasoned that `0` is valid since it's only descriptive and an empty string is okay. For `Name`, as it is typically (always) an identifier it should always have a minimum of `1`. Is this a fair assumption?

I initially wanted to add the `AllowedPatternRegexp` information, but the syntax at first seems all over the place in the CloudFormation documentation.

There are still some mistakes in this that I need to rectify, but they also depend on my question above.

I also have not done any checking against botocore to reference their specified constraints for this. What do we prefer, reference their data to avoid repeating and potential discrepencies?

### DIFF
--- a/src/cfnlint/data/ExtendedSpecs/all/03_value_types/aws_appflow.json
+++ b/src/cfnlint/data/ExtendedSpecs/all/03_value_types/aws_appflow.json
@@ -1,0 +1,82 @@
+[
+ {
+  "op": "add",
+  "path": "/ValueTypes/AWS::AppFlow::Connector.Description",
+  "value": {
+   "StringMax": 1024,
+   "StringMin": 1
+  }
+ },
+ {
+  "op": "add",
+  "path": "/ValueTypes/AWS::AppFlow::ConnectorProfile.ConnectorProfileName",
+  "value": {
+   "StringMax": 256,
+   "StringMin": 1
+  }
+ },
+ {
+  "op": "add",
+  "path": "/ValueTypes/AWS::AppFlow::ConnectorProfile.SAPODataConnectorProfileProperties.PrivateLinkServiceName",
+  "value": {
+   "StringMax": 512,
+   "StringMin": 1
+  }
+ },
+ {
+  "op": "add",
+  "path": "/ValueTypes/AWS::AppFlow::ConnectorProfile.SnowflakeConnectorProfileProperties.AccountName",
+  "value": {
+   "StringMax": 512,
+   "StringMin": 1
+  }
+ },
+ {
+  "op": "add",
+  "path": "/ValueTypes/AWS::AppFlow::ConnectorProfile.SnowflakeConnectorProfileProperties.PrivateLinkServiceName",
+  "value": {
+   "StringMax": 512,
+   "StringMin": 1
+  }
+ },
+ {
+  "op": "add",
+  "path": "/ValueTypes/AWS::AppFlow::Flow.CustomConnectorDestinationProperties.EntityName",
+  "value": {
+   "StringMax": 1024,
+   "StringMin": 1
+  }
+ },
+ {
+  "op": "add",
+  "path": "/ValueTypes/AWS::AppFlow::Flow.CustomConnectorSourceProperties.EntityName",
+  "value": {
+   "StringMax": 1024,
+   "StringMin": 1
+  }
+ },
+ {
+  "op": "add",
+  "path": "/ValueTypes/AWS::AppFlow::Flow.Description",
+  "value": {
+   "StringMax": 2048,
+   "StringMin": 1
+  }
+ },
+ {
+  "op": "add",
+  "path": "/ValueTypes/AWS::AppFlow::Flow.DestinationFlowConfig.ConnectorProfileName",
+  "value": {
+   "StringMax": 256,
+   "StringMin": 1
+  }
+ },
+ {
+  "op": "add",
+  "path": "/ValueTypes/AWS::AppFlow::Flow.SourceFlowConfig.ConnectorProfileName",
+  "value": {
+   "StringMax": 256,
+   "StringMin": 1
+  }
+ }
+]
--- a/src/cfnlint/data/ExtendedSpecs/all/03_value_types/aws_athena.json
+++ b/src/cfnlint/data/ExtendedSpecs/all/03_value_types/aws_athena.json
@@ -1,0 +1,18 @@
+[
+ {
+  "op": "add",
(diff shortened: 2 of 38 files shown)


kddejong:
@ashemedai  Interesting script any chance you would be willing to share it.  There are min/max numbers, regex patterns, string min/max, allowed values would be interesting to see what more I could pull.

We do test regex patterns to validate they are valid with python regex package. We now use regex package instead of re as it provides more support for some of the regex patterns that we are using.

ashemedai:
@kddejong That was the intention at least. :)

Let me toss it up as a gist tomorrow.

kddejong:
I made a fix do you want to look at this pr #2824

src/cfnlint/data/ExtendedSpecs/all/03_value_types/aws_cassandra.json
```diff
@@ -0,0 +1,18 @@
+[
+ {
+  "op": "add",
+  "path": "/ValueTypes/AWS::Cassandra::Keyspace.KeyspaceName",
+  "value": {
+   "StringMax": 255,
```
kddejong, line 6: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cassandra-keyspace.html#cfn-cassandra-keyspace-keyspacename


We are pulling this regex which doesn't match the described length.

src/cfnlint/data/ExtendedSpecs/all/03_value_types/aws_cassandra.json
```diff
@@ -0,0 +1,18 @@
...
+   "StringMin": 3
+  }
+ },
+ {
+  "op": "add",
+  "path": "/ValueTypes/AWS::Cassandra::Table.TableName",
```
kddejong, line 12: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cassandra-table.html#cfn-cassandra-table-tablename

Same we have the pattern that defines the length and the pattern and length descriptions don't match.

src/cfnlint/data/ExtendedSpecs/all/03_value_types/aws_athena.json
```diff
@@ -0,0 +1,18 @@
+[
+ {
+  "op": "add",
+  "path": "/ValueTypes/AWS::Athena::CapacityReservation.Name",
+  "value": {
+   "StringMax": 128,
```
kddejong, line 6: Covered by regex:
```
"AWS::Athena::CapacityReservation.Name": {
   "AllowedPatternRegex": "[a-zA-Z0-9._-]{1,128}"
  },
```

src/cfnlint/data/ExtendedSpecs/all/03_value_types/aws_athena.json
```diff
@@ -0,0 +1,18 @@
...
+   "StringMin": 1
+  }
+ },
+ {
+  "op": "add",
+  "path": "/ValueTypes/AWS::Athena::WorkGroup.Name",
+  "value": {
+   "StringMax": 128,
```
kddejong, line 14: Covered by regex:

```
"AWS::Athena::WorkGroup.Name": {
   "AllowedPatternRegex": "[a-zA-Z0-9._-]{1,128}"
  },
```

src/cfnlint/data/ExtendedSpecs/all/03_value_types/aws_appflow.json
```diff
@@ -0,0 +1,82 @@
+[
+ {
+  "op": "add",
+  "path": "/ValueTypes/AWS::AppFlow::Connector.Description",
+  "value": {
+   "StringMax": 1024,
```
kddejong, line 6: Need to check this one.  The schema defines 2048 and the docs have 1024

"Description": {
   "description": "A description about the connector that's being registered.",
   "maxLength": 2048,
   "pattern": "[\\s\\w/!@#+=.-]*",
   "type": "string"
  }

ashemedai:
@kddejong I put the current version at https://gist.github.com/ashemedai/962d3449c3712e40bb54d2fa3fba4e00

Things I would like to do:

- Properly update the project json files, I think my current code should allow it to run within the cfn-lint project, but I need to adjust the output path properly, maybe a parameter?
- Depending on how many parameters need to be added, use `argparse`.
- Tighten up existing regular expressions.
- Add regex matches for some other common cases.
- Maybe parse out/check for the `{x,y}` bits at the end of some regexp so we know to skip `StringMax`/`StringMin` since the regexp already takes care of it.

src/cfnlint/data/ExtendedSpecs/all/03_value_types/aws_athena.json
ashemedai: So when the regexp covers it, it overrides the need for any explicit `StringMax`/`StringMin`? In these case I checked the documentation and filled these in based on my reading of the regexp.

src/cfnlint/data/ExtendedSpecs/all/03_value_types/aws_redshift.json
```diff
@@ -6,5 +6,45 @@
...
+  "value": {
+   "StringMax": 2147483647,
+   "StringMin": 0
+  }
+ },
+ {
+  "op": "add",
+  "path": "/ValueTypes/AWS::Redshift::EventSubscription.SubscriptionName",
+  "value": {
+   "StringMax": 2147483647,
```
blade2005, line 30: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-redshift-eventsubscription.html#cfn-redshift-eventsubscription-subscriptionname
While the printed text says the max is 2147483647 the conditions says the max is 255. I think additional fact finding may be needed on some of these.

I've found inconsistencies with different services in how the constraints are describe and quite often conflict with each other or omit a constraint and the api throws an error.

I would bet whoever wrote the docs for cloudformation just put the 32 bit signed int max and wrote in their docs more refined constraints. Like `AWS::Redshift::EventSubscription.SourceType` says it's max length is `2147483647` but it's a valid values are: cluster, cluster-parameter-group, cluster-security-group, cluster-snapshot, and scheduled-action.

kddejong:
Lots of conflicts now as we have converted to v1.  Lets see how and where we can add these changes as needed.

PR #2831, opened 2023-07-30
Fix incorrect attribute names for Glue Job Name constraint check

gruebel:
*Description of changes:*

PR #2816 added constraints for the Glue Job resource type, but with a wrong attribute name


gruebel:
hey @kddejong would be great, if you can review my PR, because it currently fails all `AWS::Glue::Job` resources, which set a value for `Name` 😄

kddejong (approved):

kddejong:
Released in v0.79.6

PR #2905, opened 2023-10-12
Add rule W2530 to warn when SnapStart isn't configured correctly

kddejong:
*Issue #, if available:*

*Description of changes:*
- Add rule `W2530` to warn when `SnapStart` isn't configured correctly


src/cfnlint/rules/resources/lmbd/SnapStart.py
```diff
@@ -0,0 +1,65 @@
+"""
+Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+SPDX-License-Identifier: MIT-0
+"""
+
+from cfnlint.rules import CloudFormationLintRule, RuleMatch
+
+
+class SnapStart(CloudFormationLintRule):
+    """Check if the settings of multiple subscriptions are included for one LogGroup"""
```
msailes, line 10: Is this correct?

src/cfnlint/rules/resources/lmbd/SnapStart.py
```diff
@@ -0,0 +1,65 @@
...
+        "To properly leverage SnapStart, you must configure both the lambda function "
+        "and attach a Lambda version resource"
+    )
+    source_url = "https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#user-content-cloudwatchlogs"
+    tags = ["resources", "lambda"]
+
+    def __init__(self):
+        super().__init__()
+        self.resource_property_types.append("AWS::Lambda::Function")
+        self.snapstart_runtimes = ["java11", "java17"]
```
msailes, line 24: As more runtime versions are added this list will need to be maintained. Is there any way to say all Java runtimes except `jvm` and `java8.al2` to save you some work in the future?

kddejong: We can switch this to be anything starting with java that isn't java8.al2

src/cfnlint/rules/resources/lmbd/SnapStart.py
kddejong: Fixed.

src/cfnlint/rules/resources/lmbd/SnapStart.py
```diff
@@ -0,0 +1,71 @@
...
+
+            runtime = props.get("Runtime")
+            # future proofing this rule.  This will apply to newer java runtimes
+            # so we are validating its java and not java8.al2
+            if (
+                runtime
+                and (not runtime.startswith("java"))
+                and runtime not in ["java8.al2", "java8"]
+            ):
+                continue
```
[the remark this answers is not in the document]
kddejong: I'm going to add another rule `E2530` so I can cover SnapStart configured on unsupported runtimes.  `Resource handler returned message: "python3.10 is not supported for SnapStart enabled functions.` 

This can be removed then as it would be picked up by the other rule if the runtime isn't supported.

kddejong:
@greenwoodcm and @msailes some updates have been made based on your feedback

src/cfnlint/rules/resources/lmbd/SnapStart.py
```diff
@@ -0,0 +1,64 @@
...
+            props = scenario.get("Object")
+
+            snap_start = props.get("SnapStart")
+            if not snap_start:
+                continue
+
+            # Get safe removes any conditions nested
+            # since we aren't checking runtimes we just need to know if
+            # there is a scenario in which PublishedVersions is present
+            for apply_on in snap_start.get_safe("ApplyOn"):
```
greenwoodcm, line 45: can we do get_safe directly on props?

src/cfnlint/rules/resources/lmbd/SnapStartSupported.py
```diff
@@ -0,0 +1,54 @@
+"""
+Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+SPDX-License-Identifier: MIT-0
+"""
+
+from cfnlint.rules import CloudFormationLintRule, RuleMatch
+
+
+class SnapStartSupported(CloudFormationLintRule):
```
greenwoodcm, line 9: curious people's thoughts on if we want this lint or not.  the set of runtimes supporting snapstart may change in the future.  do we want to have to change this lint as that happens?  should we instead just let the service be the authority on which runtimes support the feature?

src/cfnlint/rules/resources/lmbd/SnapStart.py
kddejong: Yea, it helps remove conditions.  So if we get the following template we will validate both the `None` and `PublishedVersions`.

```yaml
ApplyOn: !If [IsUsEast1, None, PublishedVersions]
```

src/cfnlint/rules/resources/lmbd/SnapStartSupported.py
kddejong: cfn-lint is used to identify issues before they are implemented. Since a template with the improper configuration will fail at deployment time I would like to have it implemented before I try create/update a stack. 

As the service changes this capability we will update the rule. Plus we provide options to suppress this rule if needed.

kddejong: @greenwoodcm if the service has a spot in which we can pull this information I would gladly do so over configuring this manually.  We pull a lot of data from SSM, pricing APIs, etc to build the data for our rules.

src/cfnlint/rules/resources/lmbd/SnapStart.py
kddejong: Switched to the check_value helper function

PR #2955, opened 2023-11-20
Add Trunk Check instructions

joshmarinacci:
Hi. I'm an engineer at Trunk, a free extensible super-linter that uses and automates cfn-lint. This PR just adds instructions for installing and enabling cfn-lint if you already have Trunk Check installed. Please let me know if this isn't appropriate. Thanks.


kddejong:
@joshmarinacci appreciate the PR after talking to some of my team members internally I think we need to talk through our partner organization. I need to find the best way to articulate what your tools does.  I apologize on the delay.

PR #2994, opened 2024-01-02
Specifiy license in PULL_REQUEST_TEMPLATE.md

benbridts:
*Description of changes:*
This seems more fair to me (and is similar to the wording in e.g https://github.com/aws-cloudformation/cloudformation-guard/blob/main/.github/pull_request_template.md and https://github.com/aws-cloudformation/cloudformation-cli/blob/master/.github/PULL_REQUEST_TEMPLATE.md

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -3,4 +3,6 @@
 *Description of changes:*
 
 
-By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.
+By submitting this pull request, I confirm that my contribution is made under the terms of the MIT No Attribution license.
+
+


kddejong:
This is some standard wording that we are suppose to use so running this PR with some internal resources.

hyandell:
Hi Ben,

Hopping in from the Amazon/AWS Open Source Program Office. I agree that the change is basically the same (moving from the MIT-0 grant to the MIT-0 license), but I'm a bit hesitant to change one repository when we've got thousands and thousands of repositories with this text.

What was your thinking for wanting the change?

benbridts:
Hi Hyandell,

Thanks for looking at this!

I've noticed that some repositories use an explicit reference to a licence (the ones linked above, but also e.g. [bottlerocket](https://github.com/bottlerocket-os/bottlerocket/blob/develop/.github/pull_request_template.md).

In this case, like you said, there is not a practical difference between the current text and MIT-0, for other projects that are Apache or MIT (not -0) licensed , there might be.

In any case I'm mostly doing it because:
- I noticed the inconsistency
- I think for not MIT-0 licensed works there is a (slight) difference, and explicitly stating "we won't change the license" is more contributor friendly.

However, I'm not that invested (not even for the apache ones, the difference is small), so feel free to close this if that works better.

kddejong:
Chatted with @hyandell we are going to close this for now.  We can re-open as needed.

PR #3116, opened 2024-03-23
Lambda runtime deprecation updates

PatMyron:
fix https://github.com/aws-cloudformation/cfn-lint/issues/3105
https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html#runtimes-supported
https://github.com/terraform-linters/tflint-ruleset-aws/blob/master/rules/aws_lambda_function_deprecated_runtime.go

updated test results with:
```bash
gsed -i 's/Please consider updating to nodejs16/Please consider updating to nodejs20/' test/integration/*.py
gsed -i 's/Please consider updating to nodejs16/Please consider updating to nodejs20/' test/fixtures/results/*/*.json
```

### DIFF
--- a/src/cfnlint/data/AdditionalSpecs/LmbdRuntimeLifecycle.json
+++ b/src/cfnlint/data/AdditionalSpecs/LmbdRuntimeLifecycle.json
@@ -25,22 +25,22 @@
   "successor": "dotnet8"
  },
  "dotnet6": {
-  "deprecated": "",
+  "deprecated": "2025-02-11",
   "eol": "2024-11-12",
   "successor": "dotnet8"
  },
  "dotnet7": {
-  "deprecated": "",
+  "deprecated": "2024-05-14",
   "eol": "2024-05-14",
   "successor": "dotnet8"
  },
  "go1.x": {
-  "deprecated": "",
+  "deprecated": "2024-03-12",
   "eol": "2024-01-08",
   "successor": "provided.al2023"
  },
  "java8": {
-  "deprecated": "",
+  "deprecated": "2024-03-12",
   "eol": "2024-01-08",
   "successor": "java21"
  },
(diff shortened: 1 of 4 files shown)


src/cfnlint/data/AdditionalSpecs/LmbdRuntimeLifecycle.json
```diff
@@ -2,71 +2,126 @@
...
-  "successor": "dotnet6"
+  "successor": "dotnet8"
+ },
+ "dotnet5.0": {
+  "deprecated": "2022-05-10",
+  "eol": "2022-05-10",
+  "successor": "dotnet8"
+ },
+ "dotnet6": {
+  "deprecated": "",
```
PatMyron, line 28: https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html#runtimes-supported

Docs now split deprecation dates into 3 columns rather than 2 and mention:
`For new regions, Lambda will not support runtimes that are set to be deprecated within the next 6 months.`
Rather not maintain more dates for dozens of runtimes and skeptical that precision's needed anyways. Think we either:
1) Just continue picking 2 columns out of the 3
2) Pick a column and simply calculate based on it (`Error` on date, `Warn` months before, `Info` years before)

@kddejong thoughts?

kddejong: I like 2 cause it would make life a lot easier but I get a little nervous about exceptions to that. So may need to have an exception process as it comes up.

PatMyron: > exception process

Thinking no exceptions for option 2. Most rows currently span a couple months between all 3 columns. We'd err earlier (like `Info` full year in advance, `Warn` half year in advance) and modify the wording of non-error rule(s) accordingly:

https://github.com/aws-cloudformation/cfn-lint/blob/dba16cb09e142d89144577aa6505e8871589e2d0/src/cfnlint/rules/resources/lmbd/DeprecatedRuntimeEol.py#L36

If timeframes between columns increase beyond that, we'd increase advance notices for all runtimes. We haven't updated these dates in almost 2 years; it's more important to give devs more notice than bother translating all the specific documentation columns. Devs can read docs themselves as long as we surface some notice

kddejong:
Going to have to rework this for v1 and v0.  Closing this now so we can clean the approach.

Issue #3453, opened 2024-07-01
Validate Cpu/Memory combinations for Fargate AWS::ECS::TaskDefinition


trav-c:
### Is this feature request related to a new rule or cfn-lint capabilities?

rules

### Describe the feature you'd like to request

When creating an [AWS::ECS::TaskDefinition](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-taskdefinition.html) with `FARGATE` in [RequiresCompatibilities](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-taskdefinition.html#cfn-ecs-taskdefinition-requirescompatibilities) it would be helpful if cfn-lint could validate that the [Cpu](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-taskdefinition.html#cfn-ecs-taskdefinition-cpu) and [Memory](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-taskdefinition.html#cfn-ecs-taskdefinition-memory) values are a combination that is compatible with Fargate.

Fargate ECS tasks impose restrictions on the specific values, and more to the point combinations of values that can be used for the Cpu and Memory as per https://docs.aws.amazon.com/AmazonECS/latest/developerguide/task-cpu-memory-error.html and the CloudFormation documentation for [AWS::ECS::TaskDefinition.Memory](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-taskdefinition.html#cfn-ecs-taskdefinition-memory)

For example it is not possible to create a task with 1 vCPU (1024 CPU Shares) and 1GB (1024 MB) of Memory, the closest supported options are either
* 0.5 vCPU with 1GB of Memory
* OR 1 vCPU with 2GB of Memory

### Describe the solution you'd like

Conditionally validate that the CPU and Memory properties are a supported combination when `FARGATE` is present in [AWS::ECS::TaskDefinition.RequiresCompatibilities](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ecs-taskdefinition.html#cfn-ecs-taskdefinition-requirescompatibilities)

### Is this something that you'd be interested in working on?

- [ ] 👋 I may be able to implement this feature request

### Would this feature include a breaking change?

[...]

PR #3464, opened 2024-07-04
Add two new rules to validate fargate tasks

kddejong:
*Issue #, if available:*
#3453 

*Description of changes:*
- Add rule E3047 to validate ECS Fargate cpu/memory on a task definition
- Add rule E3048 to validate ECS Fargate properties on a task definition


kddejong:
While the schema specifies the values as strings they are integers.  These rules will accept both and work accordingly.  Tried to favor the string method as much as possible.

PR #4210, opened 2025-08-01
Update CreationPolicy StartFleet type

syldyer:
*Description of changes:*

AWS Documentation [1] for the `AWS::AppStream::Fleet` resource's CreationPolicy attribute is inconsistent with CloudFormation behavior.

CloudFormation only accepts template syntax such as:

```
CreationPolicy:
  StartFleet: True
```

However, cfn-lint validation is consistent with AWS Documentation and expects:
```
CreationPolicy:
  StartFleet: 
    Type: True
```

**This change is meant to update `cfn-lint` validations to align with CloudFormation CreationPolicy behavior for the `AWS::AppStream::Fleet` resource**

[1] https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-attribute-creationpolicy.html

### DIFF
--- a/src/cfnlint/rules/resources/CreationPolicy.py
+++ b/src/cfnlint/rules/resources/CreationPolicy.py
@@ -31,9 +31,7 @@ class CreationPolicy(CfnLintJsonSchema):
                 "additionalProperties": False,
                 "properties": {
                     "StartFleet": {
-                        "additionalProperties": False,
-                        "type": "object",
-                        "properties": {"Type": {"type": "boolean"}},
+                        "type": "boolean",
                     }
                 },
             }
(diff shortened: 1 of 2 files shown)


src/cfnlint/rules/resources/CreationPolicy.py
```diff
@@ -32,8 +32,7 @@ def _get_schema(self, resource_type: str) -> dict[str, Any]:
                 "properties": {
                     "StartFleet": {
                         "additionalProperties": False,
```
kddejong, line 34: We can get rid of `additionalProperties` too since it only applies to `objects`. This will drop the first error in the 2nd set of tests you changed.  `"{'Type': {}} is not of type 'boolean'",` this would be the important test based on the test case.

kddejong (changes requested):

kddejong (approved):

Issue #4307, opened 2025-12-04
Crash in StepFunctions Map validation when Map uses Iterator (AttributeError: 'NoneType' object has no attribute 'get')


tyukei:
### CloudFormation Lint Version

cfn-lint 1.42.0

### What operating system are you using?

macOS15.6.1

### Describe the bug

After upgrading `cfn-lint` from **1.41.x** to **1.42.0**, `cfn-lint` crashes while linting a CloudFormation template with `AWS::StepFunctions::StateMachine`.

When the ASL definition includes a **Map** state using **`Iterator`**, `cfn-lint` crashes with:

```
AttributeError: 'NoneType' object has no attribute 'get'
```

This did not happen on **1.41.x**. 
Also, if I change the Map state to use **`ItemProcessor`** (with `ProcessorConfig`), `cfn-lint` works without crashing.

I think this regression is related to **PR #4264** and the StartAt validation logic here:
https://github.com/aws-cloudformation/cfn-lint/pull/4264/files#diff-e8b58ac1b7ae1ace868148cc87ca85ec1152e81ab615fef42faaf3e11b39ba59R154-R159

AWS docs: Map state has two forms (**Iterator** and **ItemProcessor**):
https://docs.aws.amazon.com/step-functions/latest/dg/state-map.html

### Expected behavior

cfn-lint should not crash. 
For Map states, it should validate StartAt/States for whichever structure is present (Iterator or ItemProcessor), and safely skip validation when optional sections (like ItemProcessor) are missing.

### Reproduction template

Running cfn-lint -t inline-map.yaml (first template) crashes with the 
`AttributeError: 'NoneType' object has no attribute 'get'`
 while cfn-lint -t itemprocessor-map.yaml (second template) succeeds.

1) Crashing template (Map with Iterator)

```yml
AWSTemplateFormatVersion: "2010-09-09"
Description: Step Functions Map - Iterator (crashes)

Resources:
  StateMachineInlineMap:
    Type: AWS::StepFunctions::StateMachine
    Properties:
      StateMachineName: InlineMapExample
      RoleArn: arn:aws:iam::123456789012:role/DummyRole
      DefinitionString: |
        {
          "StartAt": "MyMap",
          "States": {
            "MyMap": {
              "Type": "Map",
              "ItemsPath": "$.items",
              "MaxConcurrency": 1,
              "Iterator": {
                "StartAt": "DoSomething",
                "States": {
[...]

PR #4264, opened 2025-10-15
State Machine: Implement check & Unit Tests for "StartAt"

r-heimann:
*Issue #, if available:*

Fixes https://github.com/aws-cloudformation/cfn-lint/issues/4074

*Description of changes:*

This PR implements a check for 

https://states-language.net/spec.html#toplevelfields

> A State Machine MUST have a string field named "StartAt", whose value MUST match exactly the name of one of the "States" fields. The interpreter starts executing the machine at the named state.

Since i am not a professional Python programmer i used the help of a LLM to implement the check and the unit tests, while trying to keep it as simple as possible.
The cfn-lint "Error" messages are almost identical to the State Machine CloudFormation API, but i made it a little bit more human friendly to read.

I tested the following CloudFormation templates, which all displayed the error message in the correct State:

<details> <summary>StartAt JSONata / JSONPath</summary>

```yaml
  StartAtJSONata:
    Type: AWS::StepFunctions::StateMachine
    Properties:
      RoleArn: !GetAtt StepFunctionRole.Arn
      Definition:
        QueryLanguage: JSONata
        StartAt: FAIL # MISSING_TRANSITION_TARGET: Missing 'Next' target: FAIL at /StartAt
        States:
          Pass: # MISSING_TRANSITION_TARGET: State "Pass" is not reachable. at /States/Pass (not yet implemented)
            Type: Pass
            Next: Success
          Success:
            Type: Succeed

  StartAtJSONPath:
    Type: AWS::StepFunctions::StateMachine
    Properties:
      RoleArn: !GetAtt StepFunctionRole.Arn
      Definition:
        QueryLanguage: JSONPath
        StartAt: FAIL # MISSING_TRANSITION_TARGET: Missing 'Next' target: FAIL at /StartAt
        States:
          Pass: # MISSING_TRANSITION_TARGET: State "Pass" is not reachable. at /States/Pass (not yet implemented)
            Type: Pass
            Next: Success
          Success:
            Type: Succeed
```

</details>

<details> <summary>Parallel StartAt JSONata / JSONPath</summary>

```yaml
  ParallelStartAtJSONata:
    Type: AWS::StepFunctions::StateMachine
    Properties:
      RoleArn: !GetAtt StepFunctionRole.Arn
      Definition:
        QueryLanguage: JSONata
        StartAt: ParallelState
        States:
          ParallelState:
            Type: Parallel
            Branches:
              - StartAt: BranchState1
                States:
                  BranchState1:
                    Type: Pass
                    End: true
[...]

PR #4308, opened 2025-12-04
Fix Step Functions Map Iterator validation crash in StateMachineDefinition

tyukei:
# Issue: #4307 

# Description of changes:
Fix cfn-lint crash in Step Functions `Map` state validation.

- Avoid calling `_validate_start_at()` with `None` for `ItemProcessor` (prevents `AttributeError: 'NoneType' object has no attribute 'get'`).
- Validate nested `StartAt` for classic Map using `Iterator` as well as new Map using `ItemProcessor`.
- Add a unit test for missing `StartAt` target inside `Iterator`.

# Related:
- Issue #4307
- Regression introduced around PR #4264: https://github.com/aws-cloudformation/cfn-lint/pull/4264/files#diff-e8b58ac1b7ae1ace868148cc87ca85ec1152e81ab615fef42faaf3e11b39ba59R154-R159
- Step Functions Map docs (Iterator vs ItemProcessor): https://docs.aws.amazon.com/step-functions/latest/dg/state-map.html


r-heimann:
Thank you for creating the PR, i was able to reproduce your issue and also test your fix - i could not find any regressions.

tyukei:
@r-heimann 
Thank you for checking!
And now I’m looking into an issue https://github.com/aws-cloudformation/cfn-lint/issues/4311
Before I start coding,  confirm plz:
1) are you already working on this in progress? If not, I wanna take it.
my plan is to keep propagating the parent `path` in `_validate_start_at()` when recursing into nested states (Parallel branches, Map ItemProcessor, Map Iterator)

2) do you prefer one PR per issue (e.g., 1 issue = 1 PR), or is it okay to bundle related fixes into this PR?

PR #4372, opened 2026-02-19
Add lambda schema extensions

chrisqm-dev:
*Description of changes:*
Adding the following lambda Rules:
- Lambda LogLevel requires JSON LogFormat (LogLevel is not supported when LogFormat is set to Text)
- Lambda Handler not supported for container images 
- Lambda Runtime not supported for container images
- Lambda Layers not supported for container images
- Added unit tests


chrisqm-dev (commented):
addressed feedback

Issue #1167, opened 2019-10-24
Rule scanning for missing Ref on pseudo-parameters


copumpkin:
*cfn-lint version: 0.24.4*

Apologies if this has already been proposed, but I just wasted some time staring at a template containing `AWS::NoValue` without a `Ref` in front of it (in an IAM policy in this case), and trying to figure out why my stack deployment was failing cryptically.

I'm wondering if `cfn-lint` could have a new rule basically stating that in 99.9% of cases, you don't want a raw string in your template saying `AWS::NoValue` or `AWS::Partition` or any of the other standard pseudo-parameters. I probably wouldn't scan for local names since those can be pretty much anything, but maybe it could be a warning.

kddejong:
We do something similar for subs.  This sounds like a good idea to me.

PatMyron:
Like this idea. Should be able to use:
https://github.com/aws-cloudformation/cfn-python-lint/blob/342ac61db052224314ca1219a7a073b45841d78e/src/cfnlint/helpers.py#L117-L126

PR #4453, opened 2026-04-09
feat: add W1054 rule to warn on raw pseudo-parameter strings without Ref

cv-dote:
*Issue #, if available:* #1167

*Description of changes:*

Add a new warning rule W1054 that detects pseudo-parameters (e.g. `AWS::NoValue`, `AWS::Region`, `AWS::Partition`) used as plain string values instead of being wrapped with `Ref`.

For example, this common mistake:
```yaml
Resource: 'AWS::NoValue'  # W1054 warning
```
Should instead be:
```yaml
Resource: !Ref AWS::NoValue  # correct
```

Design doc: https://gist.github.com/cv-dote/ca697a81c59034203fc43345537921bb


test/unit/rules/functions/test_raw_pseudo_parameter.py
```diff
@@ -0,0 +1,154 @@
...
+                    rule=RawPseudoParameter(),
+                ),
+            ],
+        ),
+    ],
+)
+def test_validate(name, instance, expected, rule, context, cfn):
+    validator = CfnTemplateValidator(context=context, cfn=cfn)
+    errs = list(rule.rawPseudoParameter(validator, {}, instance, {}))
+    assert errs == expected, f"Test {name!r} got {errs!r}"
```
kddejong, line 154: Tests need negative cases: a pseudo-param used correctly inside `Ref` (no false positive), a non-pseudo-param `AWS::` string like `AWS::EC2::Instance`, and an integration test through `CfnTemplateValidator` to exercise the `_filter.py` logic end-to-end.

kddejong (commented):
Thanks for the contribution! A couple of items to address.

src/cfnlint/jsonschema/_filter.py
```diff
@@ -137,6 +138,9 @@ def filter(
             elif REGEX_DYN_REF_SPACES.search(instance):
                 yield (instance, {"dynamicReferenceSpaces": schema}, validator)
                 return
+            elif instance in PSEUDOPARAMS and "Ref" not in validator.context.path.path:
```
kddejong: This `elif` is inside the `if validator.is_type(instance, "string") and self.validate_dynamic_references:` block, so the pseudo-param check gets skipped when `validate_dynamic_references=False` (e.g. `CfnLintJsonSchema` rules like Step Functions definitions). Should be a separate `if` block after the dynamic reference check.

cv-dote:
Hi @kddejong,

Thanks for the review! I've addressed both comments:

**1. Decoupled pseudo-parameter check from the dynamic reference block**
Moved the check out of the `if ... and self.validate_dynamic_references:` block into a separate `if` statement, so it also runs when `validate_dynamic_references=False` (e.g. `CfnLintJsonSchema` rules).

**2. Added missing test cases in `test/unit/module/jsonschema/test_filter.py`**
- `Detect raw pseudo-parameter string` — positive detection through the filter
- `Skip pseudo-parameter when used inside Ref` — no false positive for correct `Ref: AWS::NoValue` usage
- `Skip non pseudo-parameter AWS:: string` — `AWS::EC2::Instance` and similar strings are not flagged
- `test_raw_pseudo_parameter_with_dynamic_references_disabled` — end-to-end test verifying the fix for comment 1

kddejong (approved):

PR #4526, opened 2026-05-30
feat: Add rule W4002 for parameters missing from Metadata Interface

### LINKED ISSUE #4115: W4001 reversed

_No response_

Hi,

W4001 checks if all parameters referenced in tMetadata.AWS::CloudFormation::Interface exist. I was wondering if there is a check that does the opposite. Check if all parameters are listed in Metadata.AWS::CloudFormation::Interface.

The following template should fail because `SubnetId` is not listed in the Metadata.AWS::CloudFormation::Interface:

```

Metadata:
  AWS::CloudFormation::Interface:
    ParameterGroups:
      - Label:
          default: 'Vpc Parameters'
        Parameters:
          - VpcId
Parameters:
  # Vpc Parameters
  VpcId:
    Description: VPC To Create Stack In
    Type: String
  SubnetId:
    Description: Subnet To Create Stack In
    Type: String

Resources:
  Test:
    Type: AWS::S3::Bucket
    Properties:
      BucketName: !Ref VpcId
```

### Additional context

_No response_

- [ ] 👋 I may be able to implement this feature request

johnanand707:
*Issue #, if available:* Closes #4115

*Description of changes:*

Adds a new warning rule **W4002** that implements the inverse check of W4001.

- W4001 warns when a parameter referenced in `Metadata.AWS::CloudFormation::Interface` does not exist in the `Parameters` section.
- W4002 warns when a declared `Parameter` is not listed in any `ParameterGroup`, so it would not appear in the CloudFormation Console UI.

**Changes:**
- `src/cfnlint/rules/metadata/InterfaceParameterInGroup.py` — new rule (49 lines)
- `test/unit/rules/metadata/test_interface_parameter_in_group.py` — 5 parametrized unit tests covering the all-grouped, partial, no-groups, multi-group, and non-dict cases
- Updated snapshot results in `test/fixtures/results/integration/` and `test/fixtures/results/quickstart/` (pure additions of new W4002 entries; no existing entries modified)
- `test/unit/module/cfn_json/test_cfn_json.py` — updated expected failure count for `vpc-management.json` fixture (11 → 27)

**Verification:**
- `pre-commit run` passes (ruff, isort, bandit, mypy, file hygiene)
- `pytest test/unit/` — 2568 passed
- `pytest test/integration/` — 27 passed
- Smoke-tested end-to-end against a template missing a parameter from `ParameterGroups`

### DIFF
--- a/src/cfnlint/rules/metadata/InterfaceParameterInGroup.py
+++ b/src/cfnlint/rules/metadata/InterfaceParameterInGroup.py
@@ -1,0 +1,49 @@
+"""
+Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+SPDX-License-Identifier: MIT-0
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+from cfnlint.jsonschema import ValidationError, ValidationResult, Validator
+from cfnlint.rules.jsonschema.CfnLintKeyword import CfnLintKeyword
+
+
+class InterfaceParameterInGroup(CfnLintKeyword):
+    """Check that all Parameters are listed in Metadata Interface ParameterGroups"""
+
+    id = "W4002"
+    shortdesc = "Parameter not listed in Metadata Interface ParameterGroups"
+    description = (
+        "All parameters should be listed in at least one "
+        "Metadata AWS::CloudFormation::Interface ParameterGroup so they appear "
+        "in the CloudFormation Console UI"
+    )
+    source_url = "https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-cloudformation-interface.html"
+    tags = ["metadata"]
+
+    def __init__(self) -> None:
+        super().__init__(keywords=["Metadata/AWS::CloudFormation::Interface"])
+
+    def validate(
+        self, validator: Validator, _, instance: Any, schema: Any
+    ) -> ValidationResult:
+        if not isinstance(instance, dict):
+            return
+
+        grouped: set[str] = set()
+        for group in instance.get("ParameterGroups", []):
+            if isinstance(group, dict):
+                for param in group.get("Parameters", []):
+                    if isinstance(param, str):
+                        grouped.add(param)
+
+        for param_name in validator.context.parameters:
+            if param_name not in grouped:
+                yield ValidationError(
+                    f"Parameter {param_name!r} is not listed in any "
+                    "Metadata AWS::CloudFormation::Interface ParameterGroup",
+                    rule=self,
+                )
--- a/test/unit/rules/metadata/test_interface_parameter_in_group.py
+++ b/test/unit/rules/metadata/test_interface_parameter_in_group.py
@@ -1,0 +1,83 @@
+"""
+Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.
+SPDX-License-Identifier: MIT-0
+"""
+
+import pytest
+
+from cfnlint.jsonschema import ValidationError
+from cfnlint.rules.metadata.InterfaceParameterInGroup import InterfaceParameterInGroup
+
+
+@pytest.fixture(scope="module")
+def rule():
+    return InterfaceParameterInGroup()
+
+
+@pytest.fixture
+def template():
+    return {
+        "Parameters": {
+            "VpcId": {"Type": "String"},
+            "SubnetId": {"Type": "String"},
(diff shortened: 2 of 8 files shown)


kddejong (commented):
Thanks for the contribution! A few items to address:

**1. False positives when `ParameterGroups` is absent** (`src/cfnlint/rules/metadata/InterfaceParameterInGroup.py`)

The rule fires when `AWS::CloudFormation::Interface` exists but has no `ParameterGroups` key. This will produce false positives for templates that only use `ParameterLabels` — a valid pattern where the user hasn't opted into grouping. If there are no `ParameterGroups` defined, CloudFormation shows all parameters in a default list anyway, so there's nothing to warn about.

Suggest early-returning when `ParameterGroups` is absent or empty:
```python
parameter_groups = instance.get("ParameterGroups", [])
if not parameter_groups:
    return
```

**2. Error path should include `ParameterGroups`** (`src/cfnlint/rules/metadata/InterfaceParameterInGroup.py`)

When `ParameterGroups` is specified but missing parameters, the error path should include `ParameterGroups` — that's where the user needs to add the missing entries.

**3. Test structure** (`test/unit/rules/metadata/test_interface_parameter_in_group.py`)

The project convention for rule unit tests is to assert against full `ValidationError` objects (`errs == expected`), not just counts. See the sibling tests (`test_interface_configuration.py`, `test_interface_parameter_exists.py`, `test_config.py`) — they all construct explicit expected error lists with message, rule, path, etc. This catches regressions in error messages, paths, and rule attribution that a length check would miss.

Additionally, please add a test case for when the template has no parameters defined (empty `Parameters` section).

src/cfnlint/rules/metadata/InterfaceParameterInGroup.py
```diff
@@ -0,0 +1,49 @@
...
+        super().__init__(keywords=["Metadata/AWS::CloudFormation::Interface"])
+
+    def validate(
+        self, validator: Validator, _, instance: Any, schema: Any
+    ) -> ValidationResult:
+        if not isinstance(instance, dict):
+            return
+
+        grouped: set[str] = set()
+        for group in instance.get("ParameterGroups", []):
```
kddejong, line 37: The rule fires when `AWS::CloudFormation::Interface` exists but has no `ParameterGroups` key. This will produce false positives for templates that only use `ParameterLabels`. If there are no `ParameterGroups` defined, CloudFormation shows all parameters in a default list anyway, so there is nothing to warn about.

Suggest early-returning when `ParameterGroups` is absent or empty:
```python
parameter_groups = instance.get("ParameterGroups", [])
if not parameter_groups:
    return
```

src/cfnlint/rules/metadata/InterfaceParameterInGroup.py
```diff
@@ -0,0 +1,49 @@
...
+                    if isinstance(param, str):
+                        grouped.add(param)
+
+        for param_name in validator.context.parameters:
+            if param_name not in grouped:
+                yield ValidationError(
+                    f"Parameter {param_name!r} is not listed in any "
+                    "Metadata AWS::CloudFormation::Interface ParameterGroup",
+                    rule=self,
+                )
```
kddejong, line 49: When `ParameterGroups` is specified but missing parameters, the error path should include `ParameterGroups` — that is where the user needs to add the missing entries.

test/unit/rules/metadata/test_interface_parameter_in_group.py
```diff
@@ -0,0 +1,102 @@
...
+    ],
+)
+def test_validate(name, instance, expected_count, rule, validator):
+    errs = list(rule.validate(validator, "", instance, {}))
+    assert len(errs) == expected_count, (
+        f"Test {name!r}: expected {expected_count} error(s), got {errs!r}"
+    )
+    for err in errs:
+        assert isinstance(err, ValidationError)
+        assert err.rule == rule
```
kddejong, line 102: The project convention for rule unit tests is to assert against full `ValidationError` objects (`errs == expected`), not just counts. See the sibling tests (`test_interface_configuration.py`, `test_interface_parameter_exists.py`, `test_config.py`) — they all construct explicit expected error lists with message, rule, path, etc. This catches regressions in error messages, paths, and rule attribution that a length check would miss.

Additionally, please add a test case for when the template has no parameters defined (empty `Parameters` section).

PR #4566, opened 2026-06-30
Fix Elasticsearch domain instance type enum

### LINKED ISSUE #4565: E3652 false positive: AWS::Elasticsearch::Domain InstanceType enum uses OpenSearch '.search' names instead of legacy '.elasticsearch'
### Summary

`E3652` is a false positive for the `InstanceType` of the **legacy** `AWS::Elasticsearch::Domain` resource. cfn-lint validates it against an enum containing only OpenSearch-style `.search` names, so any valid legacy `.elasticsearch` instance type (e.g. `m6g.large.elasticsearch`) is rejected.

Affects **1.51.3** through **1.52.1** (latest).

### Reproduction

`repro.yaml`:

```yaml
Resources:
  Domain:
    Type: AWS::Elasticsearch::Domain
    Properties:
      ElasticsearchVersion: "7.10"
      ElasticsearchClusterConfig:
        InstanceType: m6g.large.elasticsearch
        InstanceCount: 1
      EBSOptions: { EBSEnabled: true, VolumeSize: 10 }
```

```console
$ cfn-lint --version
cfn-lint 1.52.1

$ cfn-lint repro.yaml
E3652 'm6g.large.elasticsearch' is not one of ['c4.2xlarge.search', ..., 'm6g.large.search', ..., 'ultrawarm1.medium.search'] in 'us-east-1'
repro.yaml:7:9
```

- **Expected:** `.elasticsearch`-suffixed instance types are accepted for `AWS::Elasticsearch::Domain`.
- **Actual:** rejected — the enum contains only `.search` names (note it lists `m6g.large.search` but not `m6g.large.elasticsearch`).

### Why `.elasticsearch` is correct for this resource

1. **The AWS API still enumerates `.elasticsearch` names** for the legacy service (queried 2026-06-30):

   ```console
   $ aws es list-elasticsearch-instance-types --elasticsearch-version 7.10 --region us-east-1
   { "ElasticsearchInstanceTypes": [ "m6g.large.elasticsearch", "m6g.xlarge.elasticsearch", ... ] }
   ```

   201 entries, all `.elasticsearch`, **zero `.search`**; `m6g.large.elasticsearch` is among them.

2. **The AWS docs use the `.elasticsearch` form** — [`ElasticsearchClusterConfig`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-elasticsearch-domain-elasticsearchclusterconfig.html) gives the example *"such as `m3.medium.elasticsearch`"* for `InstanceType` (and `DedicatedMasterType`).

3. **The upstream provider schema does not constrain this property** — it defines `ElasticsearchClusterConfig.InstanceType` as free-form `{ "type": "string" }`. The enum is added only by cfn-lint's pricing-sync extension, so the rejection is an over-constraint introduced by cfn-lint.

(The separate `AWS::OpenSearchService::Domain` resource correctly uses `.search` names, validated by `E3653`.)

### Root cause

The enum at

```
src/cfnlint/data/schemas/extensions/aws_elasticsearch_domain/elasticsearchclusterconfig_instancetype_enum.json
```

was an empty placeholder (`{"description": "Automatically updated using update_specs_from_pricing"}`) through 1.51.2, then was populated with `.search`-only per-region enums in 1.51.3 (commit 7d296bb3, #4523). Per #4525, the legacy `AWS::Elasticsearch::Domain` now pulls from the same pricing data as `AWS::OpenSearchService::Domain`; since that data uses unified OpenSearch `.search` names, the legacy resource ended up with `.search` names instead of its own `.elasticsearch` names.

### Suggested fix

Populate the `aws_elasticsearch_domain` `InstanceType` enum with the legacy `.elasticsearch` values (e.g. from `aws es list-elasticsearch-instance-types`) rather than the OpenSearch `.search` values — i.e. the pricing sync should not map OpenSearch instance-type names onto the legacy resource. (Since the upstream schema leaves the property unconstrained, dropping the enum for this resource would also resolve it.)

---

*Researched and drafted with [Claude Code](https://claude.com/claude-code). The reproduction, the `aws es` query, and the provider-schema inspection were run and verified before filing.*

0xTaoZ:
## What
Fixes the regional instance type enum used by `AWS::Elasticsearch::Domain` so the legacy Elasticsearch resource accepts `.elasticsearch` instance type names instead of OpenSearch `.search` names.

## Why
`AWS::Elasticsearch::Domain` still uses legacy Elasticsearch instance type names such as `m6g.large.elasticsearch`. The enum for E3652 was populated with `.search` values, which causes valid legacy templates to fail linting.

Fixes #4565.

## Details
This updates the generated enum data for the legacy Elasticsearch resource while leaving the OpenSearch resource enum separate. I also added a good fixture covering the false-positive case from #4565.

If maintainers prefer this to be fixed in the pricing/spec update path instead of committing the generated enum directly, I am happy to adjust the PR.

## Testing
- `git diff --check`
- confirmed `m6g.large.elasticsearch` is present in the `us-east-1` enum and `m6g.large.search` is no longer present for the legacy Elasticsearch enum

Note: full local `cfn-lint` execution could not complete in my environment because enhanced schema download failed with a local SSL certificate verification error.


kddejong:
Thanks for submitting @0xTaoZ based on this I submitted #4569 if you can review that it would be great but this should resolve it going forward.

PR #4569, opened 2026-07-06
fix: Restore pricing and AWS API scripts, fix Elasticsearch instance type enum

### LINKED ISSUE #4565: E3652 false positive: AWS::Elasticsearch::Domain InstanceType enum uses OpenSearch '.search' names instead of legacy '.elasticsearch'

kddejong:
## Summary
- Restores `scripts/update_specs_from_pricing.py` and `scripts/update_schemas_from_aws_api.py` that were accidentally deleted in #4539 (enhanced-schemas refactor)
- Fixes the Elasticsearch instance type enum to use `.elasticsearch` suffixes instead of `.search` — the pricing API no longer has a separate "Elastic Search Instance" product family, so we transform the OpenSearch results
- Wires both scripts back into the maintenance workflow

## Why
The schema refactor (#4539) moved resource provider schemas to the enhanced-schemas repo but also deleted the pricing/API scripts that generate extension data (instance type enums, engine versions). These data files are still consumed by rules (E3652, E3653, etc.) but were no longer being regenerated, causing stale data. Additionally, the Elasticsearch enum was populated with wrong `.search` suffixes since #4525, causing false positives for valid legacy templates.

Fixes #4565

## Test plan
- [ ] Verify `m6g.large.elasticsearch` is accepted for `AWS::Elasticsearch::Domain`
- [ ] Verify `m6g.large.search` is accepted for `AWS::OpenSearchService::Domain`
- [ ] Run maintenance workflow to confirm scripts execute successfully


0xTaoZ (commented):
Thanks for asking me to take a look. The Elasticsearch/OpenSearch enum split looks right to me. One follow-up I noticed while reading the restored pricing script: get_results() assigns the same mutable default set to each new region, so the Neptune db.serverless baseline can share discovered instance types between regional enums. I opened #4572 with a small regression test and fix.

Issue #4585, opened 2026-07-23
Perf regression since 1.51.1: Context.module_names recomputed on every evolve()


sir-sigurd:
### CloudFormation Lint Version

1.53.1 (regression present since 1.51.1; 1.51.0 is unaffected)

### What operating system are you using?

macOS (Apple Silicon); also reproduces on `ubuntu-latest` GitHub-hosted CI runners

### Describe the bug

#4510 added `module_names` to `Context`, computed in `__post_init__` as a compute-once cache. However, `Context.evolve()` constructs a new `Context` on nearly every schema-walk descend, and every construction re-runs `__post_init__`, rescanning the entire resources dict. Validation cost therefore grows with template nodes × resources, even for templates with no `::MODULE` resources at all.

Profiling cfn-lint 1.53.0 with cProfile on one of our generated templates (~250 KB, ~270 resources, zero modules): ~84,000 `Context` constructions trigger ~22.4M `str.endswith` calls, and recomputing `module_names` (always an empty tuple here) accounts for ~46% of total runtime:

```
  ncalls  tottime  cumtime  filename:lineno(function)
   84273    2.612    4.179  cfnlint/context/context.py:201(<genexpr>)
22429420    1.594    1.594  {method 'endswith' of 'str' objects}
   83130    0.423    5.452  cfnlint/context/context.py:208(evolve)
                   (of 12.6s total for the file under cProfile)
```

Across our fleet of 66 such templates, one cfn-lint invocation went from ~2m20s (1.51.0) to ~3m55s (1.53.1) on an M-series MacBook, and from ~3m30s to ~7m30s on 2-core GitHub-hosted runners.

### Expected behavior

Validation time should not regress ~1.5× on templates that don't use modules. `module_names` is derived purely from `resources` and is read at only two sites (`_keywords_cfn.py`, `rules/functions/GetAtt.py`), so it can be computed lazily, once per instance, via `functools.cached_property` — restoring pre-1.51.1 performance while keeping the caching intent of #4510. A PR with that fix accompanies this issue.

### Reproduction template

Any sufficiently large template reproduces it. Generator for a synthetic 400-resource template:

<details><summary>gen_repro.py</summary>

```python
import json

resources = {}
for i in range(400):
    resources[f"Bucket{i}"] = {
        "Type": "AWS::S3::Bucket",
        "Properties": {
            "BucketName": {"Fn::Sub": "my-bucket-%d-${AWS::Region}" % i},
[...]

PR #4590, opened 2026-07-24
Cache RefResolver subschemas lookup per instance

sir-sigurd:
*Issue #, if available:* none — follow-up to #4586, found while profiling for #4585

*Description of changes:*

`RefResolver._get_subschemas_cache()` re-ran a full breadth-first walk of the entire referrer schema document (`_search_schema`) on every `resolve()` call, despite the name suggesting a cache. The referrer schema is immutable for the lifetime of a resolver instance, so memoize the result on the instance (a new `init=False` field).

Benchmarks (v1.53.2 baseline): on a ~330 KB real-world template, cumulative profile time in `_get_subschemas_cache` drops 1.65s → 0.05s (the top self-time hotspot remaining after #4586); end-to-end wall clock on that template goes 4.20s → 3.90s (Apple M-series, best of 6, ~−7%). On a CI runner, a same-runner interleaved 3-way comparison over 72 real-world templates measured ~−4.5% end-to-end with zero overlap between arms — full table in the comment below.

Full unit test suite passes (2644 passed, 1 skipped); ruff, ruff-format, isort, and mypy are clean on the changed file.

---

*Prepared with AI assistance (Claude Code).*

### DIFF
--- a/src/cfnlint/schema/resolver/_resolver.py
+++ b/src/cfnlint/schema/resolver/_resolver.py
@@ -64,6 +64,7 @@ class RefResolver:
     _urljoin_cache: Any = field(init=True, default=None)
     _cache: Any = field(init=True, default=None)
     _cache_cfn_pointer: Any = field(init=True, default=None)
+    _subschemas_cache: Any = field(init=False, default=None)
     store: Any = field(init=True, default=None)
 
     def __post_init__(self):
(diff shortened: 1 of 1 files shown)


sir-sigurd:
Additional measurement in a CI environment, since Apple-Silicon numbers proved unreliable for small effects elsewhere: same-runner interleaved 3-way A/B/C (ubuntu-latest; 72 real-world templates, ~250–340 KB each; one full-set `cfn-lint` invocation per pass; 3 passes per arm, Latin-square ordering; venv contents assertion-verified).

| arm | passes | min |
|---|---|---|
| 1.53.2 (PyPI wheel) | 316.4 / 317.9 / 316.2 s | 316.2 s |
| 1.53.2 base commit (git install — control) | 320.0 / 325.1 / 315.9 s | 315.9 s |
| this branch (git install) | 301.5 / 310.8 / 305.3 s | 301.5 s |

Every pass of this branch is faster than every baseline pass: ~4.5% end-to-end on this workload — somewhat below the ~7% single-template number in the description, but cleanly separated from noise (the wheel arm's own spread was 0.3%).

PR #4591, opened 2026-07-24
Avoid recomputing dataclass fields() on every evolve()

sir-sigurd:
*Issue #, if available:* none — follow-up to #4586

*Description of changes:*

`Context.evolve()` and `Validator.evolve()` call `dataclasses.fields()` on every invocation — ~230k calls when linting a large template — rebuilding the fields tuple and re-filtering for `init` fields each time. The init-field name tuple is constant per class, so compute it once per class via a small `lru_cache`d helper in `cfnlint.helpers`, shared by both call sites.

Benchmarks (Apple M-series, warm, v1.53.2 baseline): on a ~330 KB real-world template, wall clock goes 4.20s → 4.11s (best of 6 runs, ~−2%). Modest on its own, but it stacks with #4590 — the two together measured ~−8–9% on two large real-world templates.

Full unit test suite passes (2644 passed, 1 skipped); ruff, ruff-format, isort, and mypy are clean on the changed files.

---

*Prepared with AI assistance (Claude Code).*

### DIFF
--- a/src/cfnlint/context/context.py
+++ b/src/cfnlint/context/context.py
@@ -15,11 +15,6 @@ from typing import TYPE_CHECKING, Any, Deque, Iterator, Set, Tuple
 
 import regex as re
 
-
-@lru_cache(maxsize=None)
-def _init_field_names(cls) -> tuple[str, ...]:
-    return tuple(f.name for f in fields(cls) if f.init)
-
 from cfnlint.context._mappings import Mappings
 from cfnlint.context.conditions._conditions import Conditions
 from cfnlint.context.parameters import ParameterSet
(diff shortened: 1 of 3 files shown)


sir-sigurd:
Closing: I can't demonstrate this change end-to-end, so I don't think it should take up review time.

Follow-up measurement on a same-runner interleaved A/B in CI (ubuntu-latest; 72 real-world templates, ~250–340 KB each; one full-set `cfn-lint` invocation per pass; 3 passes per version, alternating; branch install assertion-verified): 1.53.2 measured 322.5 / 334.8 / 319.9 s vs this branch 353.3 / 344.6 / 340.2 s — no win, if anything slower (caveats: the branch was installed from git while the baseline was the PyPI wheel, and the branch ran second in each round).

Microbenchmarks do show ~1 µs/call saved in `Context.evolve()` / `Validator.evolve()` (~230k such calls linting a large template), but that idealized ~2–4% did not reproduce on any end-to-end instrument I tried (wall clock and CPU time on Apple Silicon, wall clock on a CI runner).

#4590 is unaffected — its improvement separates cleanly from noise (~−7% wall on the same workload).

PR #4594, opened 2026-07-26
fix(codepipeline): stop E3701 flagging Fn::If OutputArtifact names as duplicates

### LINKED ISSUE #4584: E3701 false positive: Fn::If for a CodePipeline OutputArtifacts Name flagged as duplicate output-artifact name

cfn-lint 1.53.1

Windows

### Describe the bug

When an `AWS::CodePipeline::Pipeline` source action sets its `OutputArtifacts` `Name` with an `Fn::If`, cfn-lint reports **E3701 `'<name>' is already a defined 'OutputArtifact' Name`** for *both* branches of the `Fn::If`, even though only one artifact name is ever realized in a rendered template. The two branch names are mutually exclusive under the same condition, so no template ever defines the same output artifact name twice.

This comes up naturally when conditionally using the [full clone](https://docs.aws.amazon.com/codepipeline/latest/userguide/tutorials-github-gitclone.html) output format. A `CODEBUILD_CLONE_REF` artifact may only be consumed by CodeBuild actions, so when a parameter toggles the source between full clone and `CODE_ZIP`, the output artifact is given a different `Name` per mode (via `Fn::If`) so that non-CodeBuild consumers only ever bind to the `CODE_ZIP` artifact. cfn-lint treats the two conditional names as two simultaneous definitions on a single action and flags each as a duplicate of the other.

### Expected behavior

No E3701 is raised. In any rendered template exactly one `Fn::If` branch is realized, so the source action produces a single, uniquely-named output artifact. cfn-lint should resolve the condition scenarios (as it does for other rules) rather than treating the two mutually-exclusive branch names as duplicate definitions.

### Reproduction template

```yaml
Description: >-
  Minimal reproduction of a cfn-lint E3701 false positive. A CodePipeline source
  action whose OutputArtifacts Name is a Fn::If is reported as defining a duplicate
  OutputArtifact name, even though exactly one artifact name is produced in any
  rendered template (the two Fn::If branches are mutually exclusive).

Parameters:
  UseFullClone:
    Type: String
    Default: 'false'
    AllowedValues:
      - 'true'
      - 'false'

Conditions:
  IsFullClone: !Equals [!Ref UseFullClone, 'true']

Resources:
  Pipeline:
    Type: AWS::CodePipeline::Pipeline
    Properties:
      RoleArn: !Sub 'arn:${AWS::Partition}:iam::${AWS::AccountId}:role/example-pipeline-role'
      ArtifactStore:
        Type: S3
        Location: example-artifact-bucket
      Stages:
        - Name: Source
          Actions:
            - Name: Source
              ActionTypeId:
                Category: Source
                Owner: AWS
                Provider: CodeStarSourceConnection
                Version: '1'
              Configuration:
                ConnectionArn: !Sub 'arn:${AWS::Partition}:codeconnections:${AWS::Region}:${AWS::AccountId}:connection/11111111-1111-1111-1111-111111111111'
                FullRepositoryId: example-org/example-repo
                BranchName: main
                OutputArtifactFormat: !If [IsFullClone, CODEBUILD_CLONE_REF, CODE_ZIP]
              OutputArtifacts:
                - Name: !If [IsFullClone, SourceGit, SourceZip]
              RunOrder: 1
        - Name: Build
          Actions:
            - Name: Build
              ActionTypeId:
                Category: Build
                Owner: AWS
                Provider: CodeBuild
                Version: '1'
              Configuration:
                ProjectName: example-build-project
              InputArtifacts:
                - Name: !If [IsFullClone, SourceGit, SourceZip]
              RunOrder: 1
```

### Command and output

```
> cfn-lint template.yaml
E3701 'SourceGit' is already a defined 'OutputArtifact' Name
template.yaml:46:43
E3701 'SourceZip' is already a defined 'OutputArtifact' Name
template.yaml:46:54
```

### Notes

Possibly related to (but distinct from) #4001, which is also E3701 but for a different scenario (multiple pipeline resources defined in one template). This report is specifically about a single conditional `OutputArtifacts` `Name` (`Fn::If`) on one action.

Sanjays2402:
*Issue #, if available:* Fixes #4584

*Description of changes:*

An `OutputArtifacts` `Name` given as an `Fn::If` is visited twice per branch at the identical path, so E3701 compared the second visit against the entry it had just recorded for itself and reported the name as already defined. The rule now records the instance path with each name and skips same-path entries; genuine duplicates at different paths still error.

### DIFF
--- a/src/cfnlint/rules/resources/codepipeline/PipelineArtifactNames.py
+++ b/src/cfnlint/rules/resources/codepipeline/PipelineArtifactNames.py
@@ -30,7 +30,7 @@ class PipelineArtifactNames(CfnLintKeyword):
                 "Resources/AWS::CodePipeline::Pipeline/Properties/Stages/*/Actions/*/OutputArtifacts/*/Name",
             ],
         )
-        self._output_artifact_names: dict[str, list[tuple[str, dict]]] = {}
+        self._output_artifact_names: dict[str, list[tuple[str, dict, tuple]]] = {}
 
     def initialize(self, cfn):
         self._output_artifact_names = {}
@@ -49,11 +49,14 @@ class PipelineArtifactNames(CfnLintKeyword):
         if resource_name not in self._output_artifact_names:
             self._output_artifact_names[resource_name] = []
 
+        path = tuple(validator.context.path.path)
         if "OutputArtifacts" in validator.context.path.path:
-            for output_name, output_condition in self._output_artifact_names[
-                resource_name
-            ]:
-                if output_name != instance:
+            for (
+                output_name,
+                output_condition,
+                output_path,
+            ) in self._output_artifact_names[resource_name]:
+                if output_name != instance or output_path == path:
                     continue
                 try:
                     validator.evolve(
@@ -72,10 +75,10 @@ class PipelineArtifactNames(CfnLintKeyword):
                     pass
 
             self._output_artifact_names[resource_name].append(
-                (instance, validator.context.conditions.status)
+                (instance, validator.context.conditions.status, path)
             )
         elif "InputArtifacts" in validator.context.path.path:
-            for output_name, output_condition in self._output_artifact_names[
+            for output_name, output_condition, _ in self._output_artifact_names[
                 resource_name
             ]:
                 if output_name != instance:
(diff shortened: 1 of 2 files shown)


kddejong:
Thanks for the fix and the clear diagnosis, @Sanjays2402 — your root cause is spot on: an `Fn::If` `OutputArtifacts` `Name` is visited multiple times at the same path, so E3701 compared it against its own recorded entry.

This was fixed in parallel by #4596 (merged, released in v1.53.3), which takes the same approach — records the instance path with each name and skips the same-path self-visit. I confirmed on current `main` that the #4584 repro no longer raises E3701, so this change is now superseded and would conflict with the merged code (it's based on the pre-#4596 2-tuple structure).

Closing as already-resolved by #4596. Really appreciate the contribution — sorry it got fixed concurrently. 🙏

PR #4662, opened 2026-08-28
fix: Defer schema metadata updates until install succeeds

Fail-Safe:
*Issue #, if available:*

N/A

*Description of changes:*

The enhanced schema updater previously persisted the ETag obtained by its HEAD request before the schema archive was downloaded and installed. If the download, extraction, or installation failed, the existing schemas could be paired with the new ETag, causing subsequent update checks to incorrectly report that the cache was current.

This change:

- Retains candidate metadata in memory until the schema archive is installed successfully.
- Preserves existing metadata when downloading, extracting, or replacing schema files fails.
- Keeps the cached retrieval helper behavior compatible while deferring its metadata write until retrieval succeeds.
- Reports whether metadata was preserved, committed, or could not be committed.
- Adds regression coverage for HTTP 403 responses, corrupt archives, atomic replacement failures, metadata write failures, and successful installation.

Validation:

- tox -e py313: 2814 passed, 4 deselected
- pre-commit run --all-files: all hooks passed

### DIFF
--- a/src/cfnlint/helpers.py
+++ b/src/cfnlint/helpers.py
@@ -562,6 +562,21 @@ def get_url_content(url, caching=False):
     return content
 
 
+def get_url_metadata(url: str) -> tuple[dict[str, Any], str] | None:
+    """Get updated cache metadata for a URL without persisting it."""
+    req = Request(url, method="HEAD")
+    with urlopen(req) as res:
+        etag = res.info().get("ETag")
+        if not etag:
+            return None
+
+        metadata_filename = get_metadata_filename(url)
+        metadata = load_metadata(metadata_filename)
+        metadata["etag"] = etag
+        metadata["url"] = url
+        return metadata, metadata_filename
+
+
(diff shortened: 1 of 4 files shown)


kddejong (approved):
Reviewed and validated locally — traced the logic and ran the affected suites (test/unit/module/helpers/test_get_url_retrieve.py and test/unit/module/schema/test_manager.py, plus the broader schema/helpers dirs): all pass.

The fix is correct: the ETag/metadata is now persisted only after a successful download *and* install, so a failed download/extract/replace leaves the prior ETag intact and the next update re-attempts instead of reporting a stale cache as current. Failure-path coverage (403, corrupt zip, atomic-replace OSError, post-install metadata-write OSError) is thorough.

Note: after this change, get_url_retrieve's `caching=True` path is no longer used in production (the manager manages metadata itself). Will follow up with a cleanup PR to drop the dead parameter.

PR #4707, opened 2026-09-14
Skip stale FSx Lustre PERSISTENT_2 capacity warning

### LINKED ISSUE #4706: FSx Lustre max size false positive W1030

1.56.3

Mac

### Describe the bug

Linter is throwing a warning about maximum 65536 FSx Lustre size that seems to be a historical artifact that no longer applies.
```bash
chchangu@7cf34dd94b80 templates % cfn-lint w1030-bug.yml
W1030 {'Ref': 'StorageCapacityGiB'} is greater than the maximum of 65536 when 'Ref' is resolved at 'Resources/LustreScratch/Properties/StorageCapacity'
w1030-bug.yml:11:5
```

(AI analysis follows)
  Where 65,536 GiB (64 TiB) came from
  
  At GA and for the early SCRATCH_1 / SCRATCH_2 / PERSISTENT_1 deployment types, the StorageCapacity API/CloudFormation limit was valid values from 1,200 GiB up to a maximum of 65,536 GiB (64 TiB). This cap was documented directly in the FSx API reference and CloudFormation LustreConfiguration docs for those deployment types.
  
  AWS raised the limits over time:
  - PERSISTENT_2 (introduced later) supports much larger file systems — into the hundreds of TiB / PiB range depending on PerUnitStorageThroughput, well beyond 64 TiB.
  - The old 64 TiB cap was specific to the earlier deployment types and their PerUnitStorageThroughput tiers.
  
  Current constraints (they depend on deployment type)
  StorageCapacity valid values today vary by deployment type:
  - PERSISTENT_2 (SSD): 1,200 GiB, then increments of 2,400 GiB, with a much higher ceiling
  
  The authoritative, current values live in the CloudFormation AWS::FSx::FileSystem (https://docs.aws.amazon.com/AWSCloudFormation/latest/TemplateReference/aws-resource-fsx-filesystem.html#cfn-fsx-filesystem-storagecapacity) reference. Lustre can go >> 64 TiB.

### Expected behavior

Clean lint, no warning.

### Reproduction template

```yaml

Parameters:
  SubnetId:
    Type: AWS::EC2::Subnet::Id
  StorageCapacityGiB:
    Type: Number
    Description: FSx Lustre storage capacity in GiB. Must be 1200, 2400, or a multiple of 2400.
    Default: 1200
    MaxValue: 460800
    AllowedValues:
      - 1200
      - 460800

Resources:
  LustreScratch:
    Type: AWS::FSx::FileSystem
    Properties:
      FileSystemType: LUSTRE
      FileSystemTypeVersion: "2.15"
      LustreConfiguration:
        DataCompressionType: LZ4
        DeploymentType: PERSISTENT_2
        PerUnitStorageThroughput: 125
        MetadataConfiguration:
          Mode: AUTOMATIC
      StorageCapacity: !Ref StorageCapacityGiB
      StorageType: SSD
      SubnetIds:
        - !Ref SubnetId
```

0xTaoZ:
## What

Fixes #4706 by suppressing the stale 65,536 GiB resolved-Ref maximum warning only for `AWS::FSx::FileSystem` resources that are Lustre `PERSISTENT_2`.

## Why

The current CloudFormation docs make Lustre `StorageCapacity` depend on storage type and deployment type. `PERSISTENT_2` can validly exceed the older 64 TiB cap, but the enhanced provider schema still has a generic `maximum: 65536`, which makes `W1030` warn on a parameter `MaxValue` such as 460800.

The guard is intentionally narrow: non-`PERSISTENT_2` Lustre configurations still keep the existing maximum validation.

## Validation

- `.venv/bin/python -m pytest test/unit/rules/functions/test_ref.py test/unit/rules/resources/properties/test_properties.py -q`
- `.venv/bin/python -m pytest test/unit/rules/functions -q`
- `.venv/bin/ruff check src/cfnlint/rules/functions/_BaseFn.py src/cfnlint/rules/functions/Ref.py test/unit/rules/functions/test_ref.py src/cfnlint/rules/resources/properties/Properties.py test/unit/rules/resources/properties/test_properties.py`
- `.venv/bin/ruff format --check src/cfnlint/rules/functions/_BaseFn.py src/cfnlint/rules/functions/Ref.py test/unit/rules/functions/test_ref.py src/cfnlint/rules/resources/properties/Properties.py test/unit/rules/resources/properties/test_properties.py`
- `git diff --check`
- API smoke with the latest enhanced schemas release and the issue reproducer: `count 0` matches


kddejong:
Thanks for digging into this, @0xTaoZ, and for the clear reproducer — the diagnosis of the stale 65,536 GiB cap on Lustre `PERSISTENT_2` was spot on.

We're going to fix this at the schema layer rather than in the `Ref` rule, so I'm closing this in favor of two changes:

- **aws-cloudformation/resource-provider-enhanced-schemas#37** (merged) — removes the unconditional `minimum: 32` / `maximum: 65536` patch on `StorageCapacity`. Those two numbers are actually the FSx for **Windows File Server** SSD limits, and the patch was applying them to *every* `FileSystemType` — so the stale cap was wrong for Lustre, OpenZFS, and ONTAP, not just `PERSISTENT_2`.
- **#4708** — adds rule **E3725** with an `if/then` extension that encodes the correct per-`FileSystemType` bounds (Windows 32–65536, OpenZFS 64–524288, Lustre min 1200, ONTAP min 1024). Lustre and ONTAP get **no maximum**, because the documented Lustre ceiling is an adjustable, account-level Service Quota rather than a static per-file-system limit — so there's no fixed number that belongs in a schema.

Two reasons we went this route instead of suppressing the resolved-`Ref` error:

1. **Layer.** The root cause is stale schema data, so fixing the data corrects it for literal values and every intrinsic function (not just `Ref`), in all regions, with no imperative code. A hardcoded `65536` check in `Ref.py` also silently keeps validating the wrong number and only covers the `Ref` path.
2. **Scope.** Keying on `PERSISTENT_2` leaves the same false warning in place for `SCRATCH_2` / `PERSISTENT_1` Lustre file systems above 64 TiB. The stale cap is wrong for all Lustre deployment types, so the fix keys on `FileSystemType: LUSTRE`.

Really appreciate the report and the investigation — it's what surfaced the underlying schema bug. 🙏

Issue #4653, opened 2026-08-24
`AWS::CustomerProfiles::Integration` ScheduleExpression accepts values the service rejects at create time


hakin2:
### CloudFormation Lint Version

1.55.1

### What operating system are you using?

Mac

### Describe the bug

`ScheduleExpression` on `AWS::CustomerProfiles::Integration` is modelled as `pattern: ".*"`, so cfn-lint accepts any string. Customer Profiles actually accepts only four values, compared case-insensitively:

```
rate(5minutes)
rate(15minutes)
rate(1hours)
rate(1days)
```

Anything else fails at stack create with `<value> is not a supported schedule expression`, so a template that lints clean rolls back on deploy. The trap is that the property looks like an EventBridge schedule expression but is not one: `rate(1 day)`, `rate(30minutes)` and `cron(0 6 * * ? *)` are all refused.

I confirmed the list by putting 74 candidate values through `customer-profiles put-integration`. Also rejected: every other number tried (2, 10, 30, 45 minutes; 2, 6, 12, 24 hours; 2, 7, 30 days), singular units such as `rate(1day)`, any whitespace, a leading zero (`rate(01days)`), and the `at(...)` form.

The template below exits 0 on 1.55.1 even though the first resource cannot deploy.

### Expected behavior

cfn-lint should flag a `ScheduleExpression` that Customer Profiles will refuse, and stay silent on the four accepted values in any casing.

### Reproduction template
```yaml
AWSTemplateFormatVersion: "2010-09-09"
Resources:
  # Rejected at create time: "rate(1 day) is not a supported schedule expression"
  BadSchedule:
    Type: AWS::CustomerProfiles::Integration
    Properties:
      DomainName: example-domain
      ObjectTypeName: Customer
      FlowDefinition:
        FlowName: flow-bad
        KmsArn: arn:aws:kms:us-west-2:111111111111:key/00000000-0000-0000-0000-000000000000
        SourceFlowConfig:
          ConnectorType: S3
          SourceConnectorProperties:
            S3:
              BucketName: example-bucket
              BucketPrefix: input/
        TriggerConfig:
          TriggerType: Scheduled
          TriggerProperties:
            Scheduled:
              ScheduleExpression: "rate(1 day)"
              DataPullMode: Incremental
        Tasks:
          - TaskType: Filter
            SourceFields: [CustomerId]
            ConnectorOperator:
              S3: PROJECTION
```
  # Accepted by the service, must not be flagged
[...]

Issue #38, opened 2018-05-06
Check templates for Null values


kddejong:
Null values should result in a basic parsing error.  Not catching them in the parsing can result in a lot of rule errors later on.   They also fail with this error when trying to deploy

```
AWSTemplateFormatVersion: "2010-09-09"
Resources:
```

or JSON
```
{
  "AWSTemplateFormatVersion": "2010-09-09",
  "Resources": null
}
```

kddejong:
```
Error
Template validation error: [/Resources/myInstance/Type/EbsOptimized] 'null' values are 
not allowed in templates
```
when doing
```
AWSTemplateFormatVersion: "2010-09-09"
Resources:
  myInstance:
    Type: AWS::EC2::Instance
    Properties:
      ImageId: ami-123456
      EbsOptimized:
```

cmmeyer:
Merged code fix from @kddejong  -- closing issue

PR #4709, opened 2026-09-15
fix: validate CustomerProfiles schedule expressions

0xTaoZ:
## What\n\nAdd a schema patch for `AWS::CustomerProfiles::Integration` so `TriggerProperties.Scheduled.ScheduleExpression` only accepts the four schedule values supported by the Customer Profiles service, case-insensitively.\n\nFixes #4653.\n\n## Why\n\nThe generated schema currently allows any string, so templates with values such as `rate(1 day)` lint cleanly but fail during stack creation.\n\n## Tests\n\n- `.venv/bin/python -m pytest test/unit/module/schema/test_patch.py test/unit/module/schema/test_customerprofiles_patches.py -q`\n- `.venv/bin/ruff check test/unit/module/schema/test_customerprofiles_patches.py`\n- `.venv/bin/ruff format --check test/unit/module/schema/test_customerprofiles_patches.py`\n- `git diff --check`\n\nNote: I could not use the local CLI fixture as the checked-in schema cache in my environment does not contain `AWS::CustomerProfiles::Integration`, and schema refresh is blocked by a local SSL certificate verification failure.


kddejong:
Thanks for tackling #4653 — the diagnosis and the four accepted values are correct. Unfortunately this patch won't take effect as placed, and I'd like to redirect it.

**The `src/cfnlint/data/schemas/patches/` tree is no longer applied.** Since #4539 ("Source schemas from enhanced-schemas repo instead of local patching"), cfn-lint's runtime schemas are the assembled `providers/` + `resources/` content sourced from [resource-provider-enhanced-schemas](https://github.com/aws-cloudformation/resource-provider-enhanced-schemas); nothing reads the `patches/` directory anymore. I confirmed on this branch that the runtime schema for `AWS::CustomerProfiles::Integration` still has `ScheduleExpression: {"pattern": ".*"}`, and the reproduction template from #4653 still lints clean — so the behavior is unchanged.

That's also why the added test passes without catching the bug: it applies the patch to a mock schema and checks the regex in isolation, rather than asserting cfn-lint emits a finding. A real fixture template would currently pass (no finding), which is the tell.

**Where it should go:** a `manual.json` in the enhanced-schemas patch dir `schemas/patches/extensions/aws_customerprofiles_integration/`, modeled as `enumCaseInsensitive` rather than a `(?i)` regex — the service accepts a closed set of four literal strings, not a schedule grammar, so an enum gives a precise error and can't misfire. I've opened that here:

➡️ aws-cloudformation/resource-provider-enhanced-schemas#38

I also validated the accepted set against the live Customer Profiles service (`put-integration`, 28-value sweep): `rate(5minutes)`, `rate(15minutes)`, `rate(1hours)`, `rate(1days)` are accepted in any casing, and everything else is rejected — so the closed enum is complete. Once #38 merges and the schemas sync, cfn-lint will flag this via `E3030`.

Suggest closing this PR in favor of #38. Thanks again for the report and the thorough value list in #4653 — that made confirming the set easy.

PR #40, opened 2018-05-07
Fail on null values in yaml and json templates

kddejong:
*Issue #, if available:* #38 

*Description of changes:*
- Check for null values in yaml and json templates and fail gracefully

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #4712, opened 2026-09-16
fix: allow SAM functions as SNS lambda endpoints

### LINKED ISSUE #4621: W3694 SNS Subscription Endpoint should match Protocol

1.54.0

Windows

### Describe the bug

For Type: AWS::SNS::Subscrition
In Endpoint section we have defined Fn::GetAtt -LambdFunction - Arn with the protocol as lambda

LambdaFunction has the type of AWS::Serverless::Function

And hence we are getting W3694 SNS subscription endpoint should match Protocol does not match destination format of AWS::Lambda::Function.Arn

### Expected behavior

It should accept lambda functions as well as serverless

### Reproduction template

Na

0xTaoZ:
## What

Allow `Fn::GetAtt` references to `AWS::Serverless::Function.Arn` anywhere cfn-lint expects an `AWS::Lambda::Function.Arn`.

## Why

SNS subscriptions with `Protocol: lambda` can target Lambda functions defined with SAM’s `AWS::Serverless::Function`. Today W3694 flags that valid endpoint as a destination-format mismatch.

Fixes #4621.

## Validation

- `.venv/bin/python -m pytest test/unit/rules/resources/sns/test_subscription_endpoint_protocol.py test/unit/rules/functions/test_getatt_format.py -q`
- `.venv/bin/ruff check src/cfnlint/rules/functions/GetAttFormat.py test/unit/rules/resources/sns/test_subscription_endpoint_protocol.py`
- `.venv/bin/ruff format --check src/cfnlint/rules/functions/GetAttFormat.py test/unit/rules/resources/sns/test_subscription_endpoint_protocol.py`
- `git diff --check`


kddejong:
Thanks a lot for surfacing this and putting up a fix, @0xTaoZ — the analysis here was spot on.

We opted to fix the root cause in the schema data instead of the rule. The underlying problem is that `AWS::Serverless::Function`'s `Arn` attribute had no `format`, so the `Fn::GetAtt` format check (E1040/W3694) treated it as a mismatch against `AWS::Lambda::Function.Arn`. Since SAM transforms `AWS::Serverless::Function` into an `AWS::Lambda::Function`, its `Arn` genuinely *is* a Lambda function ARN, so we tagged it with that format in the enhanced schemas:

- aws-cloudformation/resource-provider-enhanced-schemas#39 / #40 (the latter applies it via a tracked `manual.json` so it survives the schema build order)

This is now in the released schemas. After `cfn-lint --update-specs`, the reported template lints clean on `main` with no rule change — and because the fix lives on the attribute's format, it also covers every other destination that expects a Lambda function ARN (EventSourceMappings, Events targets, permissions, etc.), not just SNS subscriptions.

Closing as superseded by the schema fix. Really appreciate the contribution — it directly led us to the fix. 🙏

